PR #20, opened 2020-03-10
Updated to the latest ouroboros-network depenencies

coot:


erikd:
The `sha256` hashes in the `cabal.project` file also need to be updated. The current version of `cardanp-repo-tool` can do this.

PR #25, opened 2020-03-17
CAD-345 annotation constraints

CodiePP:
adaptations to upstream changes (PR https://github.com/input-output-hk/iohk-monitoring-framework/pull/533)

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Tracing/ToObjectOrphans.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Tracing/ToObjectOrphans.hs
@@ -8,52 +8,17 @@ module Cardano.DbSync.Tracing.ToObjectOrphans () where
 
 import           Data.Text (Text)
 import           Data.Aeson ((.=))
-import           Data.Functor.Identity (Identity (..))
 
 import           Cardano.BM.Data.Tracer
-import           Cardano.BM.Tracing
 
 import           Cardano.Tracing.ToObjectOrphans (defaultTextTransformer)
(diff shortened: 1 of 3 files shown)


cardano-db-sync/src/Cardano/DbSync/Tracing/ToObjectOrphans.hs
```diff
@@ -6,17 +6,14 @@
 
 module Cardano.DbSync.Tracing.ToObjectOrphans () where
 
-import           Data.Text (Text, pack)
+import           Data.Text (Text)
```
erikd: Pack is such a generic name, and the rest of this project uses in qualified, so please switch this to use qualified. Eg:
````
import           Data.Text (Text)
import qualified Data.Text as Text
```
and then use `Text.pack` to match the code in the rest of this project.

CodiePP: `pack` is not needed anymore. But, you are right, a qualified import of Text is preferable.

CodiePP:
needs an updated cardano-node, so it's now a "WIP"

erikd:
I recently add a couple of need typeclass instacnes in the `cardano-node`. We should make sure that not instances in this module shadow instances in `node`.

CodiePP:
> I recently added a couple of needed typeclass instacnes in the `cardano-node`. We should make sure that no instances in this module shadows any instances in `node`.

yes, only one instance left here.

PR #27, opened 2020-03-18
move dependencies into a stack snapshot.yaml

KtorZ:
This way, repositories consuming cardano-db-sync as a library like
cardano-rest don't need to copy all dependencies in their own stack.yaml
and can simply rely on the snapshot to pin the exact same versions for
the cardano eco-system.


erikd:
Out of curiosity, what repos use this as a library?

KtorZ:
https://github.com/input-output-hk/cardano-rest

erikd:
Closing this.

PR #36, opened 2020-03-23
db: Correct handling of NominalDiffTime in queryIsFullySynced

erikd:


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -234,7 +235,7 @@ queryIsFullySynced = do
       let psec = diffUTCTime currentTime latestBlockTime
       -- SLot durection is in microseconds, and the difference from current time to
```
mrBliss:
```suggestion
      -- Slot duration is in microseconds, and the difference from current time to
```

mrBliss: It still says "durection" now, is that some special term I don't know? :slightly_smiling_face:

erikd: I really need a comment spell checker.

PR #52, opened 2020-04-07
CAD-710 textual log output

CodiePP:
adapt to upstream changes


cardano-db-sync/src/Cardano/DbSync/Tracing/ToObjectOrphans.hs
```diff
@@ -6,12 +6,12 @@
 
 module Cardano.DbSync.Tracing.ToObjectOrphans () where
 
-import           Data.Text (Text)
+import           Data.Text (Text, pack)
```
erikd, line 9: Please use `Text.pack` instead of the unqualified name (as per the rest of this project).

erikd:
Usually a bad idea to reference git commits that are not on master.

I think that is why that CI build failed.

CodiePP:
obsolete

PR #59, opened 2020-04-09
Docker: Clarify node-ipc volume mount and add docker-compose with secrets examples

rhyslbw:
Align with changes presented in https://github.com/input-output-hk/cardano-node/pull/779/files to enable mounting of the IPC socket into client containers without the node db coming along for the ride. Although this is primarily an upstream fix, I've taken the opportunity to align with the recent change in `cardano-node` to use the more verbose _configuration_ term so we don't end up with a variation.

- changes `data` to `node-ipc` for clarity on what is being mounted.
- Use `configuration` over `config` to conform with style in `cardano-node`
- Adds docker-compose as a getting started resource

### DIFF
--- a/config/secrets/.gitignore
+++ b/config/secrets/.gitignore
@@ -1,0 +1,3 @@
+postgres_db
+postgres_password
+postgres_user
--- a/config/secrets/postgres_db_example
+++ b/config/secrets/postgres_db_example
@@ -1,0 +1,1 @@
+cexplorer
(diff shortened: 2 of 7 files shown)


rhyslbw:
WIP until https://github.com/input-output-hk/cardano-node/pull/779 is resolved

rhyslbw:
@erikd Assigning you for approval to add example secrets config and a `docker-compose`

PR #78, opened 2020-04-22
* Update to the latest version of ouroboros-network

MarcFontaine:
* Update to the latest version of ouroboros-network
* Use ready-made codecs

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync.hs
+++ b/cardano-db-sync/src/Cardano/DbSync.hs
@@ -45,10 +45,9 @@ import           Cardano.Shell.Lib (GeneralException (ConfigurationError))
 
 import           Cardano.Slotting.Slot (WithOrigin (..))
 
-import qualified Codec.CBOR.Term as CBOR
-import qualified Codec.Serialise as Serialise
+import           Cardano.TracingOrphanInstances.Network ()
 
-import           Control.Monad.Class.MonadST (MonadST)
+import qualified Codec.CBOR.Term as CBOR
 import           Control.Monad.Class.MonadSTM.Strict (MonadSTM, StrictTMVar,
                     atomically, newEmptyTMVarM, readTMVar)
 import           Control.Monad.Class.MonadTimer (MonadTimer)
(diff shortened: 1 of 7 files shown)


erikd:
I ran `cardano-repo-tool update-stack-yaml` which updated the `stack.yaml` and then force pushed.

erikd:
I will still want to build and run this to make sure it works as expected.

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -238,50 +236,55 @@ runDbSyncNodeNodeClient iomgr trce plugin nodeConfig (SocketPath socketPath) = d
...
-              (contramap (Text.pack . show) . toLogObject $ appendName "db-sync-local-tx" trce)
-              localTxSubmissionCodec
-              (localTxSubmissionClientPeer (txSubmissionClient txv)))
-        }
+localInitiatorNetworkApplication nodeToClientVersion trce plugin pInfoConfig txv _ =
+    nodeToClientProtocols
+       (NodeToClientProtocols {
+           localChainSyncProtocol = localChainSyncProtocol
+          ,localTxSubmissionProtocol = localTxSubmissionProtocol
+          ,localStateQueryProtocol = Prelude.error "localTxSubmissionProtocol" -- todo!
```
MarcFontaine, line 251: TODO: use the ready-made null protocol

MarcFontaine:
OK `stack.yaml` is updated.
There is still one TODO, then this PR is ready for a review

erikd:
This looks file. Let me build and run it and make sure it all works!

erikd:
I just built and run this and got:
```
[db-sync-node.Handshake:Info:36] [2020-04-24 00:12:09.66 UTC]
    {"event":"Send MsgProposeVersions (fromList [(NodeToClientV_1,TInt 1097911063)])"
    ,"kind":"LocalHandshakeTrace"
    ,"bearer":"ConnectionId {localAddress = LocalAddress {getFilePath = \"\"}
    , remoteAddress = LocalAddress {getFilePath = \"state-node-testnet/node.socket\"}}"
    }
```
I really did expect `NodeToClientV_2` there?

cardano-db-sync/src/Cardano/DbSync.hs
coot: We have a null peer now: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs#L730

coot:
@erikd just to double check, the node is after #809 was merged, right? (i.e. after this commit input-output-hk/cardano-node:0172a96e16f46f1688da6dd8294481bb995fd72d) 376562b0cb6dfde4b9f41e9f3ccea45cbfb41c22

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -206,7 +204,7 @@ runDbSyncNodeNodeClient iomgr trce plugin nodeConfig (SocketPath socketPath) = d
   logInfo trce $ "localInitiatorNetworkApplication: connecting to node via " <> textShow socketPath
   networkState <- newNetworkMutableState
   txv <- newEmptyTMVarM @_ @(GenTx blk)
-  ncSubscriptionWorker_V1
+  ncSubscriptionWorker_V2
```
coot, line 207: @MarcFontaine could we use this pattern: https://github.com/input-output-hk/cardano-node/blob/master/cardano-cli/src/Cardano/CLI/Ops.hs#L436 and `nsSubscriptionWorker`; I'd like to deprecate `ncSubscriptionWorker_V1` and `ncSubscriptionWorker_V2` and remove them.

erikd:
@coot I am building/running `db-sync` from the `mafo/update-ouroboros-network` branch which is why I assume this should be running `V2`.

coot:
You're right, it is not even proposing `NodeToClientV_2`, which I strange; maybe the change I suggested in review will fix it.

erikd:
Nevermind, I just updated my node to current HEAD (commit `376562b0cb6d`) but `db-sync` still says:
```
[db-sync-node.Handshake:Info:38] [2020-04-24 05:16:42.57 UTC]
  [String "Send MsgProposeVersions (fromList [(NodeToClientV_1,TInt 764824073
  (NodeToClientV_2,TInt 764824073)])",String "LocalHandshakeTrace"
  ,String "ConnectionId {localAddress = LocalAddress {getFilePath = \"\"}
   , remoteAddress = LocalAddress {getFilePath = \"state-node-mainnet/node.socket\"}}"]
```
as expected.

cardano-db-sync/src/Cardano/DbSync.hs
erikd: Let me know when you fix that.

cardano-db-sync/src/Cardano/DbSync.hs
MarcFontaine: Yes, it is used now

cardano-db-sync/src/Cardano/DbSync.hs
MarcFontaine: I had looked it at when you suggested it on Slack, but had not been sure
if it is just a syntactic changes or if it breaks anything.
But  I changed it now.

erikd:
:crossed_fingers:

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -236,52 +234,68 @@ runDbSyncNodeNodeClient iomgr trce plugin nodeConfig (SocketPath socketPath) = d
...
-
-localInitiatorNetworkApplication
-  :: forall blk.
-     (blk ~ ByronBlock)
-  -- TODO: the need of a 'Proxy' is an evidence that blk type is not really
-  -- needed here.  The wallet client should use some concrete type of block
-  -- from 'cardano-chain'.  This should remove the dependency of this module
-  -- from 'ouroboros-consensus'.
-  => Trace IO Text
+    dbSyncVersions txv = foldMapVersions
```
coot, line 237: Could you add the type signature, it's not needed, but it makes it easier at imes.

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -236,52 +234,68 @@ runDbSyncNodeNodeClient iomgr trce plugin nodeConfig (SocketPath socketPath) = d
...
+    NodeToClientProtocols {
+          localChainSyncProtocol = localChainSyncProtocol
+        , localTxSubmissionProtocol = localTxSubmissionProtocol
+        , localStateQueryProtocol = dummyLocalQueryProtocol
         }
   where
     localChainSyncTracer :: Tracer IO (TraceSendRecv (ChainSync ByronBlock (Tip ByronBlock)))
     localChainSyncTracer = toLogObject $ appendName "ChainSync" trce
 
+    codecs = clientCodecs (configBlock topLevelConfig) ByronNodeToClientVersion2
```
coot, line 262: We should not use `ByronNodeToClientVersion2` codecs for `ByronNodeToClientVersion1`; even though the second argument is not used currently.  If we do it properly, it will be future proof.  You can add `Codecs` to argument of `dbSyncProtocols` and use the version to which you have access in `dbSyncVersions`.

cardano-db-sync/src/Cardano/DbSync.hs
coot: It does not break anything, and it's good to use the same pattern across whole ecosystem.

coot: `foldMapVersions` is like `foldMap` but for `Semigroup` rather than for `Monoid` (and specialised for `Versions` rather than for any `Semigroup`, just to make the api clearer). It folds the results with `(<>)` (one possible implementation is `foldMapVersions = \f -> foldr1 (<>) . fmap f`.  `ncSubscriptionWorker_V2` combines two versions explicitly with `(<>)` - so there's no difference.

cardano-db-sync/src/Cardano/DbSync.hs
MarcFontaine: Done

cardano-db-sync/src/Cardano/DbSync.hs
MarcFontaine: done

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -236,52 +235,75 @@ runDbSyncNodeNodeClient iomgr trce plugin nodeConfig (SocketPath socketPath) = d
...
+             (ConnectionId LocalAddress
+                 -> OuroborosApplication 'InitiatorApp BSL.ByteString IO () Void)
+    dbSyncVersions txv = foldMapVersions
+        (\v ->
+          versionedNodeToClientProtocols
+            (nodeToClientProtocolVersion (Proxy @blk) v)
+            (NodeToClientVersionData { networkMagic = nodeNetworkMagic (Proxy @blk) topLevelConfig })
+            (dbSyncProtocols v trce plugin topLevelConfig txv)
+        )
+        (supportedNodeToClientVersions (Proxy @blk))
```
coot, line 252: @erikd is `cardano-db-sync` using the same config file as `cardano-node`?
If it is, once https://github.com/input-output-hk/cardano-node/pull/830 is merged we can use the specified versions rather than `supportedNodeToClientVersions`.

erikd: no, the configs are different.

PR #80, opened 2020-04-26
Updated Proposal postgresql-setup.sh

redoracle:
Here few update proposal:
- replaced 'sed' commands with 'cut' command
- minor fix where instead of ${databasename} there was cexplorer


erikd:
Merged (https://github.com/input-output-hk/cardano-prelude/pull/82). Thanks!

PR #86, opened 2020-04-29
[ADP-286] Enable tagging latest container

sevanspowell:
- https://jira.iohk.io/browse/ADP-286
- Adds a GitHub action that, on release, triggers a BuildKite build
  job, passing through information about the GITHUB_REF and
  GITHUB_EVENT_NAME.
- Slightly modified the docker-build-push.nix. It still:
    - Pushes a docker image with the "$gitRevision" tag always.
    - Pushes a docker image with the "master" tag when the master
      branch is updated.
  But now also:
    - If the GITHUB_EVENT_NAME is "release", pushes a docker image with
      the tag "$version" (retrieved from the GITHUB_REF variable) AND
      pushes a docker image with the tag "latest".
- Updated the documentation in docker-build-push.nix.
- Change the way the creation date is set to the current time, to be
  the same as the way "cardano-wallet" does it.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,21 @@
+name: Post Release
+
+on:
+  release:
+    types:
+      - created
+
+jobs:
+    test-job:
+      name: "Tag / Build / Push Container"
+      runs-on: ubuntu-latest
+      steps:
+      - name: Trigger Buildkite Pipeline
+        uses: buildkite/trigger-pipeline-action@v1.2.0
+        env:
+          BUILDKITE_API_ACCESS_TOKEN: ${{ secrets.Buildkite_Token }}
+          PIPELINE: ${{ github.repository }}
+          COMMIT: "HEAD"
+          BRANCH: "master"
+          MESSAGE: ":github: Triggered from a GitHub Action"
+          BUILD_ENV_VARS: "{\"GITHUB_REF\": \"${{ github.ref }}\", \"GITHUB_EVENT_NAME\": \"${{ github.event_name }}\" }"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
+name: Latest Release
```
rhyslbw, line 1:
```suggestion
name: Release
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
...
+      steps:
+      - name: Trigger Buildkite Pipeline
+        uses: buildkite/trigger-pipeline-action@v1.2.0
+        env:
+          BUILDKITE_API_ACCESS_TOKEN: ${{ secrets.Buildkite_Token }}
+          PIPELINE: ${{ github.repository }}
+          COMMIT: "HEAD"
+          BRANCH: "master"
+          MESSAGE: ":github: Triggered from a GitHub Action"
+          BUILD_ENV_VARS: "{\"BUILDKITE_TAG\": \"latest\"}"
```
rhyslbw, line 21: This should be the Git tag, not `latest`

rhyslbw: Also, there should be a specific env for the `RELEASE_VERSION`, and subsequent changes to the build script, to ensure a random tag build doesn't get the `latest` Docker tag (for example). It's still possible to manually create a tag build, or something as simple as the Buildkite filter rules changing would break this. All releases have tags, but not all tags are releases.

.github/workflows/release.yaml
sevanspowell: https://github.com/input-output-hk/cardano-db-sync/pull/86/commits/dd1550ca75a9a8a79c4c6f8e9c3394324e3313fb#diff-499885e3a90e0cf4f87aa4cdc18d4879R1

Using "Post Release" instead, because cardano-node already has a "release" action.

.github/workflows/release.yaml
sevanspowell: Fixed this as discussed, using GITHUB_EVENT_NAME and GITHUB_REF.

PR #89, opened 2020-05-06
[CD-8] Move the schema to Persistant

ksaric:
https://jira.iohk.io/browse/CD-8


cardano-db/test/test-db.hs
```diff
@@ -1,18 +1,37 @@
...
 
+import           System.Environment (lookupEnv, setEnv)
+import           System.Directory (getCurrentDirectory)
+import           System.FilePath ((</>))
+
 main :: IO ()
-main =
+main = do
+  -- If the env is not set, set it to default.
+  maybePgPassFile <- lookupEnv "PGPASSFILE"
```
erikd, line 19: This, and the `whenNothing` should be a separate function. that is called just before `defaultMain`.

cardano-db/test/test-db.hs
```diff
@@ -1,18 +1,37 @@
...
+  _ <- whenNothing maybePgPassFile $ do
+      currentDir        <- getCurrentDirectory
+      let pgPassFile    = currentDir </> "../config/pgpass"
```
erikd, line 22: This code base does not use vertical alignment. on `=` and `<-` and so on.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -136,5 +143,108 @@ share
...
+    poolDeposit         Word64              sqltype=lovelace
+    poolDepositRefund   Word64              sqltype=interval
+    poolDepositDecay    Word64              sqltype=rational
+    maxEpoch            Word64
+    nOptimal            Word64
+    influence           Word64              sqltype=rational
+    monetaryExpandRate  Word64              sqltype=interval
+    treasuryGrowthRate  Word64              sqltype=interval
+    activeSlotCoeff     Word64              sqltype=interval
+    decentralisation    Word64              sqltype=interval
```
erikd, line 215: I thought that was a floating point value in the range `[0 .. 1]`

schema/migration-2-0014-20200430.sql
```diff
@@ -0,0 +1,320 @@
...
+
+    -- Basic rewards data.
+    -- What reward has been earned in the epoch by delegating to the specified pool id.
+    -- This design allows rewards to be discriminated based on how they are earned.
+
+    CREATE TABLE reward (
+        id                      bigint              NOT NULL,
+        reward_address          staking_addr_hash   NOT NULL,
+        instantaneous           boolean             NOT NULL,
+        pool_id                 hash32type          NOT NULL,  -- Arbitrary if an instantaneous reward - needed for primary key
```
erikd, line 61: Were these comments added manually? Probably not a good idea to manually add comments to a generated file.

The comments belong in the Haskell schema definition.

ksaric: This whole file was added manually.

cardano-db/src/Cardano/Db/Schema.hs
ksaric: @kevinhammond The interval type just defines the minimum and the maximum. Do we need this to be a floating-point or a 0 - 100 integer?

schema/migration-2-0014-20200430.sql
erikd: It was added to git manually, but wasn't generated using cardano-db-tool create-migration` ?

If so, deleting it and recreating it with `create-migration` should produce the same file.

erikd: The other thing is that we are likely to want to collapse all the migrations periodically.

Comments in these generated files is unlikely to survive migration collapsing. Comments should go in the Persistent schema definition.

ksaric: Regenerated the file.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -128,5 +128,13 @@ share
     deriving Eq
     deriving Show
 
+  -- The bit of the pool data on the chain.
+  -- This doesn't leave the internal database.
+  PoolOnData
```
erikd, line 140: Why `PoolOnData` ? Can we come up with a better name or provide a better comment to explain what this does?

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -47,6 +47,12 @@ insertTxIn = insertByReturnKey
 insertTxOut :: MonadIO m => TxOut -> ReaderT SqlBackend m TxOutId
 insertTxOut = insertByReturnKey
 
+insertPoolOnData
```
erikd, line 50: I think we need to carefully separate Shelley specific features from Byron especially since the two chains are represetned by different types.

schema/migration-2-0013-20200426.sql
```diff
@@ -0,0 +1,21 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 13 THEN
```
erikd, line 8: This version will need to change. In fact, deleting this and running `create-migration` again should fix it.

cardano-db/test/test-db.hs
erikd: Cherry picked a version of this `PGPASSFILE` fix into another PR.

cardano-db/src/Cardano/Db/Schema.hs
ksaric: Pool online data - PoolOnData.
The comment can be updated, not sure what information is missing.

cardano-db/src/Cardano/Db/Insert.hs
ksaric: Sure, but I would argue that it's still early for that.
The way I see it in phases:
1. add all function in the same module
2. you see duplication in the module and you create a separate module for some
3. you see more duplication and you abstract over both modules

schema/migration-2-0013-20200426.sql
ksaric: Not sure. If I rebase it then sure, otherwise not. Look at the state of https://github.com/input-output-hk/cardano-db-sync/tree/ksaric/CD-8/schema in the current branch now.

ksaric: Fixed.

cardano-db/src/Cardano/Db/Schema.hs
ksaric: Updated.

erikd: In this case i wonder if `PoolOnlineData` may not be a better name

erikd: Kevin's schema calls this `TxMetadata`. I suspect `PoolMetadata` might not be the best name.

erikd:
This is not relevant anymore.

erikd:
This is not relevant any more.

PR #90, opened 2020-05-08
Extract subscription function (Towards simplified API)

MarcFontaine:
This is just an experiment/ Proxy PR to get feedback / comments on the API.
(or this Iteration of API-work)
`ApiTest` is to be moved to the right place.

Basically this reorders the functions, arguments and applications
and moves all the defaults, that are not expected to be changed in the application,  to `ApiTest`.

### DIFF
--- a/cardano-db-sync/src/ApiTest.hs
+++ b/cardano-db-sync/src/ApiTest.hs
@@ -1,0 +1,78 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module ApiTest
+where
+import           Data.Proxy
+import           Data.Void (Void)
+import qualified Data.ByteString.Lazy as BSL
+import           Ouroboros.Consensus.Config (TopLevelConfig, configBlock)
+import           Ouroboros.Consensus.Network.NodeToClient (clientCodecs, ClientCodecs)
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion (
+                    nodeToClientProtocolVersion , supportedNodeToClientVersions)
+import           Ouroboros.Consensus.Node.Run (RunNode, nodeNetworkMagic)
+import           Ouroboros.Network.Mux (AppType (..), OuroborosApplication)
+import           Ouroboros.Network.NodeToClient (ClientSubscriptionParams (..),
+                    ConnectionId, LocalAddress,
+                    NodeToClientProtocols (..),
+                    NetworkClientSubcriptionTracers,
+                    NodeToClientVersionData (..),
+                    ncSubscriptionWorker,
+                    newNetworkMutableState,
+                    versionedNodeToClientProtocols)
+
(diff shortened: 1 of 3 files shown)


cardano-db-sync/src/ApiTest.hs
```diff
@@ -0,0 +1,78 @@
...
+import           Control.Monad.Class.MonadST (MonadST)
+
+import           Prelude
+
+subscribe ::
+  ( RunNode blk , MonadST m1 )
+  => Snocket.LocalSnocket
+  -> TopLevelConfig blk
+  -> NetworkClientSubcriptionTracers
+  -> ClientSubscriptionParams ()
```
coot, line 39: This is not related, but it seems that `a` in `ClientSubscriptionParams a` is not used.

cardano-db-sync/src/ApiTest.hs
```diff
@@ -0,0 +1,78 @@
...
+import           Ouroboros.Network.Protocol.Handshake.Version (DictVersion, Versions, foldMapVersions)
+import qualified Ouroboros.Network.Snocket as Snocket
+
+import qualified Ouroboros.Network.NodeToClient (NodeToClientVersion)
+import           Control.Monad.Class.MonadST (MonadST)
+
+import           Prelude
+
+subscribe ::
+  ( RunNode blk , MonadST m1 )
```
coot, line 35: Why `m1`? `m` is probably good enough.

cardano-db-sync/src/ApiTest.hs
```diff
@@ -0,0 +1,78 @@
...
+  => Proxy blk
+  -> TopLevelConfig blk
+  -> (NodeToClientVersion blk
+      -> ClientCodecs blk m1
+      -> NodeToClientProtocols appType bytes m2 a b)
+  -> Versions
+       Ouroboros.Network.NodeToClient.NodeToClientVersion
+       DictVersion
+       (ConnectionId LocalAddress
+          -> OuroborosApplication appType bytes m2 a b)
```
coot, line 70: Do we need `m1` distinct from `m2`? This seem strange to me. Anyway, when the application will run it must hold that `m1 ~ m2` unless the codec is not used at all.

MarcFontaine: This calls out to `ncSubscriptionWorker`
```ncSubscriptionWorker
    :: forall appType x y.
       ( HasInitiator appType ~ True
       )
    => LocalSnocket
    -> NetworkClientSubcriptionTracers
    -> NetworkMutableState LocalAddress
    -> ClientSubscriptionParams ()
    -> Versions
        NodeToClientVersion
        DictVersion
        (ConnectionId LocalAddress ->
           OuroborosApplication
             appType
             BL.ByteString IO x y)
    -> IO Void
```
This fixes the `m` to `IO`.
Should I make a PR that merges `ApiTest.hs` into `module Ouroboros.Network.NodeToClient` ?

MarcFontaine:
subsumed by https://github.com/input-output-hk/cardano-db-sync/pull/105

PR #92, opened 2020-05-12
Removes SQL Views required by Cardano GraphQL

rhyslbw:
I'm refactoring the application to manage the views directly, so removing here to enforce better isolation.


erikd:
Tested locally. All good!

PR #100, opened 2020-05-22
[CD-9] Insert Shelley data into the (modified) database

ksaric:
https://jira.iohk.io/browse/CD-9

Based on top of https://github.com/input-output-hk/cardano-db-sync/pull/89


cardano-db-sync/cardano-db-sync.cabal
```diff
@@ -41,6 +41,7 @@ library
                         Cardano.DbSync.Plugin
                         Cardano.DbSync.Plugin.Default
                         Cardano.DbSync.Plugin.Default.Insert
+                        Cardano.DbSync.Plugin.Default.InsertShelley
```
erikd: Would much prefer `Cardano.DbSync.Plugin.Default.Byron.Insert` and `Cardano.DbSync.Plugin.Default.Shelley.Insert`.

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
```diff
@@ -0,0 +1,131 @@
...
+import qualified Shelley.Spec.Ledger.BlockChain as SL
+import           Shelley.Spec.Ledger.Tx as SL
+
+import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock (..), Crypto)
+import           Ouroboros.Network.Block (BlockNo (..), Tip, SlotNo (..), getTipBlockNo)
+import           Ouroboros.Network.Point (withOrigin)
+
+
+mkSlotLeader :: Crypto c => SL.Block c -> DB.SlotLeader
+mkSlotLeader blk =
```
erikd: Since that is an internal function that is not exported, can we have that at the bottom of the file?

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
```diff
@@ -0,0 +1,131 @@
...
+        blockHash = Crypto.abstractHashToBytes . Crypto.hash . SL.bhash . SL.bhbody . SL.bheader $ blk
+
+    let blockHeaderSize :: Int
+        blockHeaderSize = SL.bHeaderSize . SL.bheader $ blk
+
+    let getTxInternalUnsafe :: SL.TxSeq c -> Seq (SL.Tx c)
+        getTxInternalUnsafe (SL.TxSeq txSeq) = txSeq
+
+    let txsCount :: Int
+        txsCount = length . getTxInternalUnsafe . SL.bbody $ blk
```
erikd: Can we have all these in the `where` section?

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
```diff
@@ -0,0 +1,131 @@
...
+                    , DB.blockMerkelRoot = Nothing -- Not sure how to translate this.
+                    -- Just $ unCryptoHash (blockMerkelRoot blk)
+                    , DB.blockSlotLeader = slid
+                    , DB.blockSize       = fromIntegral blockHeaderSize
+                    , DB.blockTime       = DB.slotUtcTime meta slotNumber
+                    , DB.blockTxCount    = fromIntegral txsCount
+                    }
+
+    -- TODO(KS): Insert the transaction is MISSING!
+    --mapMVExceptT (insertTx tracer blkId) $ blockPayload blk
```
erikd: Since the DB is not ready for this yet, this code is not ready yet.

cardano-db-sync/src/Cardano/DbSync/Util.hs
```diff
@@ -68,6 +74,11 @@ import qualified Ouroboros.Network.Point as Point
 import           Ouroboros.Network.Block (Point (..), SlotNo (..))
 
 
+
+-- | Currently we use Bitcoin alphabet for representing addresses in base58
+addrToBase58 :: Crypto crypto => Addr crypto -> ByteString
+addrToBase58 = encodeBase58 bitcoinAlphabet . serialize'
```
erikd, line 80: That will not be correct.Mu understanding is that we will be using bech32.

cardano-db/test/test-db.hs
```diff
@@ -1,18 +1,37 @@
...
 main :: IO ()
-main =
+main = do
+  -- If the env is not set, set it to default.
+  maybePgPassFile <- lookupEnv "PGPASSFILE"
+  _ <- whenNothing maybePgPassFile $ do
+      currentDir        <- getCurrentDirectory
+      let pgPassFile    = currentDir </> "../config/pgpass"
+      setEnv "PGPASSFILE" pgPassFile
+      return pgPassFile
```
erikd: Why return this and then throw away the result?

cardano-db/test/test-db.hs
```diff
@@ -1,18 +1,37 @@
...
+import           System.Environment (lookupEnv, setEnv)
+import           System.Directory (getCurrentDirectory)
+import           System.FilePath ((</>))
+
 main :: IO ()
-main =
+main = do
+  -- If the env is not set, set it to default.
+  maybePgPassFile <- lookupEnv "PGPASSFILE"
+  _ <- whenNothing maybePgPassFile $ do
```
erikd: Why not just a simple `case` on that?

cardano-db-sync/cardano-db-sync.cabal
ksaric: The idea was to remove `Default` since I'm not sure what I'm defaulting to and replace that with specific implementations of `Byron` and `Shelley`.

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
ksaric: Sure.

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
ksaric: Sometimes this is more readable to me, but I can move them below, yes.

cardano-db-sync/src/Cardano/DbSync/Plugin/Default/InsertShelley.hs
ksaric: Again, this looks like you are viewing some older commit.

cardano-db-sync/src/Cardano/DbSync/Util.hs
ksaric: For the pool hash? After talking with Matthias, we arrived at the following conclusion.
We use base64 encoding here.
Submissions are identified by the subject's Bech32-encoded Ed25519 public key (all lowercase).
An Ed25519 public key is a 64-byte string. We'll typically show such string in base16.
base64 is fine too, more concise. But bech32 is definitely overkill here.
I used base58 since that is the same format Byron used and can be used without changing the database format.

cardano-db/test/test-db.hs
ksaric: I guess that can work here as well, `whenSomething` is pretty standard as well.

cardano-db/test/test-db.hs
ksaric: Because we have no use for it? What would we do with it?
And if somebody needs to debug this or change this, I would argue that this is the correct return result.

cardano-db/test/test-db.hs
erikd: Nevermind I cherry picked a version of that into `master` since it is completely unrelated to Shelley stuff anyway.

cardano-db-sync/cardano-db-sync.cabal
erikd: The are currently two extecutables in this repo, `cardano-db-sync` and `cardano-db-sync-extended`. The former uses this `Default` version of the functionality and the latter uses `Default` as well as the functionality in `Cardano.DbSync.Plugin.Epoch`.

erikd:
This is not relevant any more.

PR #118, opened 2020-06-03
DO NOT MERGE: Admit 32 byte Shelley address hashes

erikd:
Address hashes are supposed to be 28 bytes on both Byron and Shelley
but currently on Shelley, they are 32 bytes. This commit allows
32 byte hashes to be inserted into the database.


erikd:
The good bits of this has been merged to master via a separate PR.

PR #142, opened 2020-06-19
[CD-52] bump iohk-nix

craigem:
Also:

Removes from cardano-db-sync service:
  * genesisHash
  * RequiresNetworkMagic

Additionally replaces `ghc` with `compiler-nix-name`.

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -60,7 +60,7 @@ let
       src = ../.;
       name = "cardano-db-sync";
     };
-    ghc = buildPackages.haskell-nix.compiler.${compiler};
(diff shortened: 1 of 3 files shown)


nix/haskell.nix
```diff
@@ -60,7 +60,6 @@ let
       src = ../.;
       name = "cardano-db-sync";
     };
-    ghc = buildPackages.haskell-nix.compiler.${compiler};
     modules = [
```
hamishmack, line 63: I think you probably want to add `inherit compiler-nix-name;` here

craigem: Done :-)

PR #156, opened 2020-06-26
[CAD-1266] Fix Stack and expose required modules.

ksaric:
https://jira.iohk.io/browse/CAD-1266

Required if we really want to reuse the plugin functionality outside the `cardano-db-sync` project.

### DIFF
--- a/cardano-db-sync/cardano-db-sync.cabal
+++ b/cardano-db-sync/cardano-db-sync.cabal
@@ -31,30 +31,38 @@ library
                         -Wincomplete-uni-patterns
 
   exposed-modules:      Cardano.DbSync
-                        Cardano.DbSync.Plugin.Epoch
-
-  other-modules:        Cardano.DbSync.Config
-                        Cardano.DbSync.Database
-                        Cardano.DbSync.DbAction
-                        Cardano.DbSync.Era
-                        Cardano.DbSync.Era.Byron.Genesis
-                        Cardano.DbSync.Era.Byron.Util
-                        Cardano.DbSync.Era.Shelley.Genesis
-                        Cardano.DbSync.Era.Shelley.Util
+                        Cardano.DbSync.Types
                         Cardano.DbSync.Error
+                        Cardano.DbSync.Util
+                        Cardano.DbSync.Era
+                        Cardano.DbSync.Era.Byron.Util
+                        Cardano.DbSync.Era.Shelley.Util
+
+                        Cardano.DbSync.Plugin.Epoch
+                        Cardano.DbSync.Config
+                        Cardano.DbSync.Database
+
+                        Cardano.DbSync.DbAction
+
                         Cardano.DbSync.Metrics
                         Cardano.DbSync.Plugin
                         Cardano.DbSync.Plugin.Default
-                        Cardano.DbSync.Plugin.Default.Insert
+
                         Cardano.DbSync.Plugin.Default.Rollback
+
+                        Cardano.DbSync.Tracing.ToObjectOrphans
+
+  other-modules:        Cardano.DbSync.Era.Byron.Genesis
+                        Cardano.DbSync.Era.Shelley.Genesis
+
+                        Cardano.DbSync.Plugin.Default.Insert
+                        
                         Cardano.DbSync.Plugin.Default.Byron.Insert
                         Cardano.DbSync.Plugin.Default.Byron.Rollback
                         Cardano.DbSync.Plugin.Default.Shelley.Insert
                         Cardano.DbSync.Plugin.Default.Shelley.Query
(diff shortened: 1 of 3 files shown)


cardano-db-sync/cardano-db-sync.cabal
```diff
@@ -31,17 +31,20 @@ library
                         -Wincomplete-uni-patterns
 
   exposed-modules:      Cardano.DbSync
+                        Cardano.DbSync.Types
+                        Cardano.DbSync.Error
+                        Cardano.DbSync.Util
+                        Cardano.DbSync.Era.Byron.Util
+                        Cardano.DbSync.Era.Shelley.Util
+
```
erikd, line 39: What specifically do you need from those modules. There are a huge amount of internals exposed from those. If you insist on having those exposed you get to keep all the pieces when I break things. Or we could figure out what you actually need and just expose that.

cardano-db-sync/src/Cardano/DbSync/Types.hs
```diff
@@ -25,6 +25,8 @@ module Cardano.DbSync.Types
   , ShelleyTxOut
   , ShelleyTxSeq
   , SocketPath (..)
+  -- * Re-exports (we don't need 15 deps)
+  , Tip
```
erikd, line 29: Re-exports of types from other packages are *always* a bad.

ksaric:
Running into:
```
$> PGPASSFILE=config/pgpass stack run cardano-db-sync-extended -- --config config/mainnet-config.yaml --genesis-file ../cardano-node/mainnet-genesis.json --socket-path ../cardano-node/state-node-mainnet/node.socket --schema-dir schema/

[db-sync-node:Info:4] [2020-06-26 12:16:03.35 UTC] NetworkMagic: RequiresNoMagic 764824073
cardano-db-sync-extended: DbInsertException "SlotLeader" (SqlError {sqlState = "23514", sqlExecStatus = FatalError, sqlErrorMsg = "value for domain hash32type violates check constraint \"hash32type_check\"", sqlErrorDetail = "", sqlErrorHint = ""})
```
Dropped the constraint for now, requires proper fix.

erikd:
It is currently known to be broken form Shelley (but working for Byron). I have still not fully sorted out all the issues arising from the switch from 32 byte address hashes to 28 byte address hashes.

cardano-db-sync/cardano-db-sync.cabal
ksaric: I need quite a lot?
Currently, I would leave this exposed and minimize this over time.

cardano-db-sync/src/Cardano/DbSync/Types.hs
ksaric: Removed.

cardano-db-sync/cardano-db-sync.cabal
erikd: As long as you are aware that this project is under **HEAVY** development at the moment and anything could change at any time.

I still have not been made aware of what SMASH is or how it interoperates with `db-sync`, nor has my advice been sought on how the interoperate with `db-sync` as smooth as possible.

stack.yaml
```diff
@@ -90,9 +91,9 @@ extra-deps:
     commit: 5e0b8bc8c7862be12da6989440f8644ba7c1e1cf
     subdirs:
       - binary
+      - cardano-crypto-praos
       - binary/test
       - cardano-crypto-class
-      - cardano-crypto-praos
```
erikd, line 95: It helps to avoid semanitcally null changes like this. And before you changed them, the were actually sorted alphabetically and now they are not (OCD alert!!!).

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -18,6 +18,7 @@ module Cardano.DbSync
   , GenesisHash (..)
   , NetworkName (..)
   , SocketPath (..)
+  , DB.MigrationDir (..)
```
erikd, line 21: Why re-export that here instead of just importing it from `Cardano.Db` ?

ksaric: I can do that as well.

ksaric: Fixed.

ksaric: Wait. I now remember why. Because then I depend on `Cardano.Db` as well as `Cardano.DbSync`.

stack.yaml
ksaric: I actually sorted them as they appear in Cabal but then Stack got changed a couple of times and I didn't bother to fix that.

ksaric:
:+1:

PR #169, opened 2020-07-04
update for latest version of cardano-client

MarcFontaine:
Mininal changes to make db-sync compile with latest version of `cardano-client`


erikd:
Thanks!

Issue #183, opened 2020-07-13
cardano-node 15.1 incompatible with latest master of db-sync


papacarp:
I know erik hasn't worked insuring db-sync works with 15.1 yet.  I already tried updating the dependencies and it didn't work so dropping this in here for now.  

db-sync-extended exits with the following error code on startup:
```
cardano-db-sync-extended: DecoderFailure ChainSync ServerAgency TokNext TokCanAwait DeserialiseFailure 90 "Size mismatch when decoding Block.\nExpected 4, but found 2."
```

erikd:
Working on this now. Hopefully today or tomorrow.

erikd:
And of course it has turned out to be more complicated than anticipated.

PR #185, opened 2020-07-14
Support the release/1.15.x version of the node for Shelley

erikd:
Needed to find the exact version of the network dependency that was compatible with the node.

Closes: https://github.com/input-output-hk/cardano-db-sync/issues/183


erikd:
Took another pass at it and this version works with the `release/1.15.x` branch of `node` and is not too much of `networkd` downgrade.

PR #187, opened 2020-07-15
Update to head of cardano node

newhoggy:
Updating to head of `cardano-node` where `Cardano.Config.Shelley.Orphans` is no longer defined.


erikd:
I already have this  as part of a WIP branch. I will cherry pick the gitignore patch an ignore the rest.

Issue #85, opened 2020-04-28
Feature request: Add SocketPath to configuration file, allowing --socket-path CLI arg to override


rdlrt:
Since cardano-node and CLI [started ](https://github.com/input-output-hk/cardano-node/pull/612) accepting `GenesisFile` and `SocketPath` , it would be good to eliminate these from cardano-db-sync as well for consistency.

rhyslbw:
`--socket-path` is still a valid command line argument, it's not also just possible to provide in the config file

erikd:
`genesis-file` is being moved to the config file in https://github.com/input-output-hk/cardano-db-sync/pull/188 .

rdlrt:
Closing as stale

PR #188, opened 2020-07-16
Configuration/CLI preparation for HFC

erikd:


rhyslbw:
Might want to consider https://github.com/input-output-hk/cardano-db-sync/issues/85 in this change too

erikd:
I personally like the socket path being explicit as I often switch around the path, so having to edit a file would be a pain.

Obviously the right solution would be to have it in the config file and allow an optional CLI override, but that is significantly more work (ie testing) than I have time and patience for right now.

PR #219, opened 2020-07-28
db: Add migration notification

erikd:


erikd:
This currently does not work. Need to investigate why.

erikd:
This does seem to work. Running on this branch, in a `psql` terminal:
```
main-candidate=# LISTEN cardano_db_sync_startup ;
LISTEN
```
If i then restart the db-sync program and run the above command again I get:
```
main-candidate=# LISTEN cardano_db_sync_startup ;
LISTEN
Asynchronous notification "cardano_db_sync_startup" with payload "init" received from server process with PID 2856301.
Asynchronous notification "cardano_db_sync_startup" with payload "db-setup" received from server process with PID 2856401.
```

PR #229, opened 2020-07-31
Add a tx_metadata table

erikd:


erikd:
@vargasmontero This should be provide the Tx metadata you are requesting. In the Haskell side it is just a `Text` containing the meta data encoded as JSON, but as far as PostgreSQL is concerned it should be JSON.

There currently is no tx metadata on the mainnet chain so I don't really have a way of testing this.

erikd:
@vargasmontero Thinking some more it may be better to merge this now and fix it if it is incorrect.

erikd:
Merging this.

PR #238, opened 2020-08-06
bugfix - close issue #237, insert certs in correct order.

mebassett:
Before https://github.com/input-output-hk/cardano-db-sync/pull/234 certificates were inserted in the DB in a particular order: PoolCerts, DelegCerts, and then MirCerts.

After that PR introduced a cert index (based on the order in which they appear in the tx) it inserted them based on that order.

This led to cases where a MirCert tried to log rewards against a stake address that did not exist in the database yet, as it had a higher index.

I have fixed this by sorting the certificates based on their type before attempting to insert them.

There are two other issues arising from this: 
1) Some people (including myself!) reported difficulty reproducing #237 if they used a database that had already synced successfully. This happened even if they used `--rollback-to-slot 4550000`.

   The reason for this is that the rollback functions simply call `deleteCascade` (from the Persistent sql library) to delete blocks,
   and the cascade ensures that it deletes txs, et cetera.

   HOWEVER stake addresses have their own id with no foriegn key constraints, so they are not cleared.

   I did not fix this bug, but reviewers can easily reproduce #237  by using `--rollback-to-slot 4550000` and then quickly killing the process before it has a chance to continue.  Then you can run the following sql query:

```sql
  delete from stake_address
  where id not in (select distinct addr_id from
			(select addr_id from delegation
			 union
			 select reward_addr_id from pool_update
			 union
			 select addr_id from reserve
			 union
			 select addr_id from reward
			 union
			 select addr_id from stake
			 union
			 select addr_id from stake_deregistration
			 union
			 select addr_id from stake_registration
			 union
			 select addr_id from treasury union select addr_id from withdrawal) a);
```

   After that query deletes the left-over stake addresses, you can re-run db-sync with the rollback command and it'll reproduce #237.

   I am not sure if this issue should be fixed by adding a foreign key dep on the stake_address table to tx or if one should use a similar delete query.

2) I am confused how a tx could contain "out of order" certs in the first place. 

   If on the one hand a "Stake address" always "exists" whether or not the database is aware of it then perhaps it makes more sense if db-sync used a "just in time" mechanism to query stake addresses.  Id est, if the address does not yet exist, just create it.

   If, on the other hand, a stake address doesn't exist until the respective certificate exists, shouldn't this certificate come before any other certificates that attempt to use that stake address?  
   If this is the case, then the index being returned by `txCertificates` is most likely wrong.

   I must confuse I do not understand the code in
   https://github.com/input-output-hk/cardano-ledger-specs/blob/683182a1baa24582e653da60f237ad0a4368306d/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/TxData.hs#L515
   well enough to be able to say.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
@@ -147,7 +147,28 @@ insertTx tracer env blkId blockIndex tx = do
       Nothing -> pure ()
       Just md -> insertTxMetadata tracer txId md
 
-    mapM_ (insertCertificate tracer env txId) $ Shelley.txCertificates tx
+    -- even though certs come in order and have their index, we still need
+    -- to be careful about the order we insert them into the database.
+    -- DCert_mir, for instance, when written to the db will want to refer
+    -- to a stake address that may only be created by a DCert_regkey that
+    -- happens in the same transaction but with a higher index.
+    -- Therefore we always insert Shelley.DCertDeleg (for DCert_regkey, et al)
+    -- first.
+    -- Question for IOHK: why would this ever happen? it appears that it
+    -- does in https://github.com/input-output-hk/cardano-db-sync/issues/237.
+    let  certOrdering :: (Word16, ShelleyDCert) -> Int
+         certOrdering (_, cert) = 
+                      case cert of
+                        Shelley.DCertPool (Shelley.RegPool _) -> 0
+                        Shelley.DCertPool (Shelley.RetirePool _ _) -> 1
+                        Shelley.DCertDeleg (Shelley.RegKey _) -> 2
+                        Shelley.DCertDeleg (Shelley.DeRegKey _) -> 3
+                        Shelley.DCertDeleg (Shelley.Delegate _) -> 4
+                        Shelley.DCertMir _ -> 5
+                        Shelley.DCertGenesis _ -> 6
+
+    mapM_ (insertCertificate tracer env txId) $ sortOn certOrdering
+                                              $ Shelley.txCertificates tx
     mapM_ (insertWithdrawals tracer txId) $ Shelley.txWithdrawals tx
 
     case Shelley.txParamUpdate tx of


SebastienGllmt:
I just tested an the work fixed for me as described

1. rollback using `db-sync-node-extended/bin/cardano-db-sync-extended   --config config/mainnet-config.yaml   --socket-path ../cardano-node/bin/\\.pipecardano-node   --schema-dir schema/ --rollback-to-slot 4550000` and *SHUT DOWN BEFORE IT RESYNCS*

2. Run the following SQL query
```
psql -U your-username-here -d cexplorer -c "delete from stake_address  where id not in (select distinct addr_id from  (select addr_id from delegation  union  select reward_addr_id from pool_update  union  select addr_id from reserve  union  select addr_id from reward  union  select addr_id from stake  union  select addr_id from stake_deregistration  union  select addr_id from stake_registration  union  select addr_id from treasury union select addr_id from withdrawal) a);"
```

mebassett:
@dcoutts I suspect as much and left a comment about that specifically.   I'll repeat it here:

>   I am confused how a tx could contain "out of order" certs in the first place.
>
>    If on the one hand a "Stake address" always "exists" whether or not the database is aware of it then perhaps it makes more sense if db-sync used a "just in time" mechanism to query stake addresses. Id est, if the address does not yet exist, just create it.
>
>    If, on the other hand, a stake address doesn't exist until the respective certificate exists, shouldn't this certificate come before any other certificates that attempt to use that stake address?
>    If this is the case, then the index being returned by txCertificates is most likely wrong.
>
>    I must confuse I do not understand the code in
>    https://github.com/input-output-hk/cardano-ledger-specs/blob/683182a1baa24582e653da60f237ad0a4368306d/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/TxData.hs#L515
>    well enough to be able to say.

nevertheless, trying to insert the certs into the database in the order in which they appear in the tx is _certainly_ wrong, because it fails.  So its either one of the two cases I described above or something else.  I haven't the skill to figure out which.  Can you advise what I should do to narrow down the possibilities ?

erikd:
The problem was addressed in PR #240 so closing this.

PR #246, opened 2020-08-13
add - index on tx_out.payment_cred

mebassett:
handy if you want to query utxos by payment credential.


rhyslbw:
@mebassett Would you mind adding an index on `pool.hash_id` in this migration too?

PR #258, opened 2020-08-18
schema: Do not drop 'pg_stat_*' views on migration

erikd:
I will not merge this until @johnalotoski has confirmed that it works as required.


erikd:
@johnalotoski pm-ed me to confirm that this works as expected.

PR #261, opened 2020-08-20
schema: Remove redundant indices

erikd:


newhoggy:
Why are these redundant?  Is it because we don't use queries that rely on these indices?

erikd:
According to PgHero, they are redundant due to existing uniqueness constraints.

newhoggy:
Nice!

newhoggy (approved):

newhoggy (approved):

PR #305, opened 2020-09-30
Populate the rewards table

erikd:


erikd:
I am resyncing the database from scratch one last time just to make sure.

PR #308, opened 2020-10-01
db-sync: Completely rejig how db-sync is configured

erikd:


ArturWieczorek:
@erikd Does this PR mean that https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/building-running.md or other docs will have to be updated too ?

erikd:
@ArturWieczorek Once the Nix infrastructure is fixed, that document is probably correct, but it should be validated.

PR #315, opened 2020-10-01
db: Fix uniqueness constraint for PoolRetire table

erikd:


erikd:
Currently syncing the database from scratch so I can validate the fix.

erikd:
Rebased against master to avoid the preformance regression in `cardano-ledger-specs` : https://github.com/input-output-hk/cardano-ledger-specs/pull/1884

PR #370, opened 2020-10-21
Run all tests

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -0,0 +1,96 @@
...
+        echo "LIBSODIUM_PATH=$LIBSODIUM_PATH"
+        echo "$LIBSODIUM_PATH" >> $GITHUB_PATH
+
+    - name: Install libsodium
+      if: matrix.os == 'macos-latest'
+      run: brew install libsodium
+
+    - name: Configure to use libsodium
+      run: |
+        cat >> cabal.project <<EOF
```
erikd, line 50: That is wrong. It should be `cabal.project.local` .

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,96 @@
...
+        restore-keys: cache-${{ runner.os }}-${{ matrix.ghc }}-v1-${{ hashFiles('cabal-cache.cabal') }}-
+
+    - name: Install dependencies
+      run: cabal build all --builddir="$CABAL_BUILDDIR" --only-dependencies
+
+    - name: Build
+      run: cabal build all --builddir="$CABAL_BUILDDIR"
+
+    - name: Git clone
+      run: git clone https://github.com/input-output-hk/cardano-mainnet-mirror
```
erikd, line 93: Why does it need this?

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,96 @@
...
+
+    - name: Run tests
+      run: cabal test --builddir="$CABAL_BUILDDIR" all
```
erikd, line 96: Probably worth adding `-j1` so that one test is run at a time and the outputs of the test suites to not get mangled.

.github/workflows/haskell.yml
newhoggy: The reason I append to `cabal.project` is because `cabal configure` clobbers `cabal.project.local`.

newhoggy: If it move it after the `cabal configure` it should work though.

erikd: Also, `cabal configure` should not be needed with `cabal >= 3.0`. I have not run it in over a year.

newhoggy: I hope to eventually introduce `cabal-cache` to speed up build times which needs the `configure` step, so it's useful to keep it separate.

I've noticed the rudimentary cache we use is broken for MacOS and it isn't very good because you need at least one good build (per change in cabal file version) for something to be cached.

.github/workflows/haskell.yml
newhoggy: Removed!  Thanks!

.github/workflows/haskell.yml
newhoggy: Fixed.  Thanks!

.github/workflows/haskell.yml
erikd: If you need the separate configure step it would still be better to append the libsodium stuff to `cabal.project.local` (after it is created).

newhoggy: Done!

newhoggy:
The database set up has been going for over two hours and looks like this:

image

erikd:
That is not the database setup. That is `db-sync` trying (and retrying) to the socket the `node` creates.

erikd:
Errr, why is it trying to connect to a node?

.github/workflows/haskell.yml
```diff
@@ -117,11 +117,10 @@ jobs:
     - name: Build
       run: cabal build all --builddir="$CABAL_BUILDDIR"
 
-    - name: Set up config cardano-node
+    - name: Configure and run testnet
       run: |
-        git clone https://github.com/input-output-hk/cardano-node.git
-        sed -i 's|^NodeConfigFile:.*$|NodeConfigFile: ../cardano-node/configuration/cardano/mainnet-config.json|g' config/mainnet-config.yaml
-        cat config/mainnet-config.yaml
+        cat config/mainnet-config.yaml | sed 's|^NodeConfigFile:.*$|NodeConfigFile: ../../cardano-node/configuration/cardano/mainnet-config.json|g' > config/mainnet-config.yaml.tmp
```
erikd, line 122: Its sets up `db-sync` but .....

.github/workflows/haskell.yml
```diff
@@ -135,11 +134,14 @@ jobs:
...
-        PGPASSFILE=config/pgpass-mainnet "$cardano_db_sync_exe" \
-            --config config/mainnet-config.yaml \
-            --socket-path node.socket \
-            --state-dir ledger-state/mainnet \
-            --schema-dir schema/
+        # PGPASSFILE=config/pgpass-mainnet "$cardano_db_sync_exe" \
+        #     --config config/mainnet-config.yaml \
+        #     --socket-path node.socket \
+        #     --state-dir ledger-state/mainnet \
+        #     --schema-dir schema/
```
erikd, line 141: Does not run it? Why?

.github/workflows/haskell.yml
```diff
@@ -135,11 +134,9 @@ jobs:
...
 
         PGPASSFILE=config/pgpass-mainnet scripts/postgresql-setup.sh --createdb
-        PGPASSFILE=config/pgpass-mainnet "$cardano_db_sync_exe" \
-            --config config/mainnet-config.yaml \
-            --socket-path node.socket \
-            --state-dir ledger-state/mainnet \
-            --schema-dir schema/
+
+    - name: Clone mainnet
+      run: git clone https://github.com/input-output-hk/cardano-mainnet-mirror
```
erikd, line 139: Why? We don't need it.

.github/workflows/haskell.yml
```diff
@@ -135,11 +134,9 @@ jobs:
...
 
     - name: Run tests
-      run: cabal test --builddir="$CABAL_BUILDDIR" all -j1
+      run: CARDANO_MAINNET_MIRROR="$(pwd)/cardano-mainnet-mirror/epochs" cabal test --builddir="$CABAL_BUILDDIR" all -j1
```
erikd, line 142: Don't need that.

cabal.project
```diff
@@ -47,6 +47,37 @@ package cardano-node
 package ouroboros-consensus-cardano
   tests: False
 
+-- The following is needed because Nix is doing something crazy.
```
erikd, line 50: Why are we adding all these? I worked so hard to remove them all.

.github/workflows/haskell.yml
```diff
@@ -117,11 +117,10 @@ jobs:
     - name: Build
       run: cabal build all --builddir="$CABAL_BUILDDIR"
 
-    - name: Set up config cardano-node
+    - name: Configure and run testnet
```
erikd, line 120: I don't think we need any of this.

.github/workflows/haskell.yml
newhoggy: Deleted this.

erikd (approved):

PR #372, opened 2020-10-21
Update resolver to cardano-1.21.1.yaml

hasufell:
This is needed so the dependencies are properly in sync
and cardano-db-sync or smash can be used from within
the wallet.

----

Related: https://github.com/input-output-hk/cardano-wallet/pull/2249

1. I don't know what those sha256 hashes are and they aren't documented, so I didn't update it :man_shrugging: 
2. I didn't update nix expressions, since `nix/regenerate.sh` removes cabal.project and does nothing else on my machine :sweat_smile:

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -84,7 +84,7 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-node
-  tag: 4f44b01f190030dd56f40edb462a2addb6df4061
+  tag: a819311473563cb2ab3cd91543cf0f63facdf43e
   --sha256: 0pglk1fnj17c36r35dqhfnnksff6vpi3ng6jqq7vc8bxhhh41z51
   subdir:
     cardano-api
@@ -129,7 +129,7 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/ouroboros-network
-  tag: f6466b6473df52a42316061e495f0defa2a71442
+  tag: 9e498e0962044c582df0cbf2f81fa0450a67d5f7
   --sha256: 0wvqrnhhlgx90cccsdgj94qgnvy6yb41x63gihdascx1cnhxzkik
   subdir:
     cardano-client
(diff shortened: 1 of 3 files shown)


cabal.project
```diff
@@ -84,7 +84,7 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-node
-  tag: 4f44b01f190030dd56f40edb462a2addb6df4061
+  tag: a819311473563cb2ab3cd91543cf0f63facdf43e
```
erikd, line 87: That is not going to happen. You are downgrading from a GH from Oct 13th to one from Oct 5th.

One of the next things I need to do is mode to an even more recent recent version of `node` in order to support Allega/Mary.

hasufell: So you are saying your releases don't correspond to any of the snapshots cardano-node and cardano-wallet are using? Can you explain to me how I can include cardano-db-sync in cardano-wallet when all dependencies are different?

hasufell: `4f44b01f190030dd56f40edb462a2addb6df4061` doesn't seem to be any released cardano-node version (correct me if I missed something) and @dcoutts explained on slack that we should only use released versions.

erikd: > So you are saying your releases don't correspond to the snapshots cardano-node and cardano-wallet are using?

Until this week, building with `cardano-wallet` was never even the on the radar.

> Would you be able to explain to me how I can include cardano-db-sync in cardano-wallet

I would strongly advise you not to. Neither `wallet` nor `db-sync` ever considered this before this week.

> 4f44b01f190030dd56f40edb462a2addb6df4061 doesn't seem to be any released cardano-node version (correct me if I missed something) and @dcoutts explained on slack that we should only use released versions.

The `master` branch of `db-sync` will occassionally **need** to depend on `node` git hashes that are not formal releases. You should avoid depending on those. Similarly you should avoid depending on non-release versions of `db-sync`.

So what **release** of `db-sync` are you using?

erikd:
Closing this. It has bitrotted.

PR #376, opened 2020-10-22
Expose all symbols

hasufell:


erikd (approved):
Sure, but when it breaks, you get to keep both pieces.

erikd:
In future, please raise PRs against a branch of this repo rather than a fork of this repo. CI does not fully work against forks of the repo.

PR #377, opened 2020-10-23
Update index state

newhoggy:


erikd:
I have a WIP PR for this that ran into trouble with Allegra/Mary changes. I am about to release `6.0.0` and would prefer not to have Allegra/Mary changes in that release. My branch for this is `erikd/allegra-mary` and currently does not compile.

erikd:
If you can bump `index-state` while avoiding Allegra/Mary then got for it.

newhoggy:
I checked the history, and unfortunately this is not possible because the nothunk fixes in `cardano-node` that I need are rolled up into the same commit `bc97ebd1eb6cdd36efc1061234180f4f62bb54cc` as the Allegra/Mary changes.

newhoggy:
I'll leave this PR parked until `6.0.0` is released.

erikd:
It might make more sense just to deep six this PR in favor of my `erikd/allegra-mary` branch.

erikd:
Index state has been updated in another PR and those dependency updates are almost certainly no longer relevant.

Closing this.

PR #378, opened 2020-10-23
Github action support (no tests)

newhoggy:
Getting to the tests to run in Github Actions is taking longer than I thought, so I'm submitting a PR that just does the compile and will be submitted the test PR separately.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -1,0 +1,112 @@
+name: Haskell CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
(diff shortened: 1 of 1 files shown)


erikd (approved):
Yeah, ok!

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,110 @@
...
+    runs-on: ${{ matrix.os }}
+
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: ["8.6.5"]
```
erikd, line 16: No `8.10.2` ?

newhoggy: `8.10.2` will need to wait until the project updates `cardano-node` dependency to a version that supports this compiler.

PR #385, opened 2020-10-26
Enable unit and property tests in CI

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -110,3 +110,6 @@ jobs:
 
     - name: Build
       run: cabal build all --builddir="$CABAL_BUILDDIR"
+
+    - name: Test
+      run: cabal test cardano-db:test --builddir="$CABAL_BUILDDIR"
```
erikd, line 115: There are two tests in `cardano-db`. I would therefore suggest `cabal test all -j1 ...`.

erikd (changes requested):

.github/workflows/haskell.yml
newhoggy: I'm skipping that for the moment because I haven't figured out how to properly set up database and testnet for it to run.

erikd: The tests need a database to connect to (because it tries to run the migrations) but does not need to connect to a node.

PR #389, opened 2020-10-28
adds txMetadata as cbor into a bytea column on the tx_metadata table.

mebassett:
Yoroi requires this for Catalyst.

as the unique key on this table is (key, txid), it continues to split
up the tx_metadata by key.  it recreates the TxMetadata value for
each key.


erikd (changes requested):
Can we please rename the `cbor` column to `bytes`? And then squash the commits down to a single commit.

erikd:
Thanks @mebassett !

PR #393, opened 2020-10-30
set correct gitrev on CI built binaries

disassembler:
this needs some haskell changes as well as it's not using the same approach to getting the gitrev as other projects.


disassembler:
Haskell code updated to use `gitRev` from `cardano-config`

erikd (approved):

PR #401, opened 2020-11-11
db-sync: Log an error if excessive rollback is required on startup

erikd:


cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -33,6 +33,13 @@ rollbackToSlot trce slotNo =
               [ "Rolling back to slot ", textShow (unSlotNo slotNo), ", hash "
               , maybe (if unSlotNo slotNo == 0 then "genesis" else "unknown") renderByteArray mHash
               ]
+        count <- lift $ DB.queryBlocksAfterSlot (unSlotNo slotNo)
+        when (count > 50000) .
+          liftIO . logError trce $
+            mconcat
+              [ "Rollback block count is high and therefore rollback is likley to be incredibly slow. "
```
xdzurman, line 40: typo "likely"

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -33,6 +33,13 @@ rollbackToSlot trce slotNo =
...
+              , "Dropping the database and resyncing from scratch will be faster."
```
xdzurman, line 41: I would also add a line saying 'Make sure you have correctly set the state directory/volume.'

erikd: The is superfluous. It must be set on the command line, and the directory is created it if does not exist.

Issue #256, opened 2020-08-18
Delete/Cascade is slow


erikd:
The delete/cascade operation is slow. Its main purpose is testing and debugging, but deleting 1 million rows can take 30 minutes or more.

Persistent issue: https://github.com/yesodweb/persistent/issues/1113

erikd:
Progress is being made: https://github.com/yesodweb/persistent/pull/1121#issuecomment-684107454 .

erikd:
The Persistent PR has been merged, but last time I checked, Esqueleto did not yet work with the new Persistent.

PapaSwer:
How can I implement it in the cardano-db-sync ?

erikd:
This is a library (Persistent/Esqueleto) issue, rather than a db-sync issue.

PapaSwer:
Is there anyway to import the fixed library to work with cardano-db-sync?

erikd:
I need to investigate that, but it is not at the top of my list if priorities.

Issue #397, opened 2020-11-05
Slot rollback take longer than syncing the db


PapaSwer:
After starting cardano-db-sync:
```
[db-sync-node:Info:31] [2020-11-05 05:02:14.10 UTC] Starting chainSyncClient
[db-sync-node:Info:31] [2020-11-05 05:02:14.15 UTC] Cardano. Db tip is at slot 7066517, block 4618318
[db-sync-node:Info:36] [2020-11-05 05:02:14.15 UTC] Running DB thread
[db-sync-node:Info:36] [2020-11-05 05:02:17.60 UTC] Rolling back to slot 6640000, hash dc3a7980681d38a4bbcbc1bec5f4e6f74d755848ee07889857ef83e78ca6765b
[db-sync-node:Info:36] [2020-11-05 05:02:18.03 UTC] Deleting slots numbered: [7066517..6640025]
[db-sync-node:Info:39] [2020-11-05 05:02:34.74 UTC] getHistoryInterpreter: acquired
-------------------FREEZE-------------------------------
```
Even after 24h no change is made.

erikd:
If you  look at `top` or similar you will see that both postgres and `db-sync` are working really hard. 

Whenever the number of slots to roll back is large (like you have there) it will be slow due to https://github.com/input-output-hk/cardano-db-sync/issues/256 .

The solution (while #256 remains unfixed) is to drop and recreate the database. That will take significantly less than 24m hours.

rcmorano:
Problem for me with this is that it won't ever catch up tip because it falls into one of these rollbacks no matter how many times I've tried a sync from the scratch :(

erikd:
Unless db-sync is being restarted this should not happen.

rcmorano:
> The only way this rollback could happen while syncing is if db-sync or the node it is connected to is being stopped and restarted (and even then is unlikely).
> 
> If there is a problem, I need to be able to recreate it. Unfortunately, I have synced from scratch at least 50 times and I have not seen anything like this even once.

Maybe also because of some postgres crash/hang up? I've been trying to sync on very low resources instances and also on this "weird" setup where I share the socket through TCP with `socat` (as a I told you about on some other issue), so yeah in my case postgres/cardano-node probably hung up for too long on some query, restarted... A myriad of bad things could have happened, but somehow 5.0.2 version managed to recover from those issues but on 5.0.3+ I haven't been able to reach tip :(

[...]

PR #414, opened 2020-11-30
Add Migration for efficient rollback

kderme:
Fixes https://github.com/input-output-hk/cardano-db-sync/issues/256 and maybe https://github.com/input-output-hk/cardano-db-sync/issues/397. There are two persistent issues that needs to be fixed before merging this
- One has a pr https://github.com/yesodweb/persistent/pull/1167
- The other I haven't fixed yet. The generated migration is slightly wrong since it drops a constrant:
`    EXECUTE 'ALTER TABLE "block" DROP CONSTRAINT "block_previous_id_fkey"' ;`
without replacing it with a correct one. It looks like a persistent bug, easy to fix.


cabal.project
```diff
@@ -146,3 +146,21 @@ source-repository-package
...
+  tag: 305b11e58e6a7cff9d639d409f0ace88cf154af0
+  --sha256: 0cihcscwgs0d8wmvbm9934lj6ghdyswil26dc0nhrfzdjj7iig4g
+  subdir:
+    .
+
+source-repository-package
+  type: git
+  location: https://github.com/kderme/persistent
+  tag: 2d02c002761ba0ed05ce95aeef1dfc7bd3caa073
+  --sha256: 0jl6ph2cgrl3n2f6j7ff9gqm0scz2dwxq5vs1q1ijxlli37zfzsx
```
erikd, line 162: This is great, but I wonder if it might be more sensible to wait until these are formally released.

kderme:
The second bug is discussed [here](https://github.com/yesodweb/persistent/issues/1169) and a fix is proposed. The fix seems to work at least for our case, so this pr is ready to test now. My manual tests show that slot rollback works fine.

erikd:
Probably worth investigating these CI failures.

kderme:
> Probably worth investigating these CI failures.

This seems to be related to the --only-dependencies flag of ci and the fact that I use a custom branch of persistent at the cabal.project. Eventually though we won't have a custom branch (unless we want this pr merged asap).

erikd:
Rebased and squashed down this PR. Hope it passes CI.

PR #416, opened 2020-12-03
[DO NOT MERGE] 7.0.x release branch

erikd:


erikd:
The conflicts in ` docker-compose.yml` and `stack.yaml` noted above are not relevant the `7.0.x` release.

cardano-db-sync/src/Cardano/DbSync/Era/Allegra/Util.hs
```diff
@@ -215,8 +215,8 @@ txInvalidHereAfter :: Shelley.Tx StandardAllegra -> Maybe Word64
 txInvalidHereAfter =
```
rhyslbw, line 215:
```suggestion
txInvalidHereafter =
```

rhyslbw: I detected the casing error while mapping the column in `cardano-graphql`

At this stage I'll be remapping as
```
tx.invalidHereAfter AS "invalidHereafter"
```
... , however it would be best to fix now for consistency.

rhyslbw (changes requested):

cardano-db-sync/src/Cardano/DbSync/Era/Allegra/Util.hs
erikd: PR is https://github.com/input-output-hk/cardano-db-sync/pull/417 .

PR #427, opened 2020-12-08
[DO NOT MERGE] Version 7.1.0

erikd:


docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
         max-file: "10"
 
   cardano-node:
-    image: inputoutput/cardano-node:1.23.0
+    image: inputoutput/cardano-node:1.24.1
```
rhyslbw, line 27: `1.24.2`

rhyslbw (changes requested):

docker-compose.yml
erikd: Fixed.

ksaric:
@erikd Can you please leave the commit `3cf868ec8c06265e4b670ac737af640d716e5ef7` separate and not rewrite Git history? For now, I'm depending on that commit and once you have things ready we can sync up and I can target your merge commit.

erikd:
If no further issues are found and this does indeed become the `7.1.0` release, this commit will remain. If an issue is found and it needs to be corrected, this commit will be replaced by the fix commit.

ksaric:
> If no further issues are found and this does indeed become the `7.1.0` release, this commit will remain. If an issue is found and it needs to be corrected, this commit will be replaced by the fix commit.

Yes, but could you just temporarily add a new commit and squash them when it's finalized?

PR #428, opened 2020-12-08
Fix building with Stack and use existing resolvers.

ksaric:
Now able to build with Stack and is reusing existing resolvers from `cardano-haskell` which the wallet team provides.


erikd:
Thanks!

PR #431, opened 2020-12-09
db-sync: Fix DB types for txInvalidBefore/txInvalidHereafter

erikd:


cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -104,9 +104,10 @@ share
     deposit             Int64                                   -- Needs to allow negaitve values.
     size                Word64              sqltype=uinteger
 
-    invalidBefore       Word64 Maybe        sqltype=uinteger
-    invalidHereafter    Word64 Maybe        sqltype=uinteger
+    invalidBefore       Word64 Maybe        sqltype=word64type
+    invalidHereafter    Word64 Maybe        sqltype=word64type
```
kderme, line 108:
```suggestion
    invalidHereafter    DBWord64 Maybe      sqltype=word64type
```
I think this is better. A failure like this was reported 
```
runDBThread: PersistMarshalError "Couldn't parse field `invalidHereafter` from table `tx`. Failed to parse Haskell type 
`Word64`; expected integer from database, but received: PersistRational (4500280 % 1). Potential solution: Check 
that your database schema matches your Persistent model definitions."
```
which seems to trigger the very strange bug related to Rationals. I see that your `DBWord64` has proper fix for this, but `Word64` from Persistent doesn't. 

I also see a related commend 
`-- If the value is greater than MAX_INT64, it comes back as a PersistRational (wat??).`
It seems a Rational can appear for even smaller numbers (probably there is some bigger bug here at Persistent). But `DBWord64` should handle this.

kderme (changes requested):

kderme:
I tested with `postgresql-setup.sh` and my fix doesn't create any migrations. This is expected since my change affects the type level and not the db level.

erikd:
Thanks @kderme . I had that same fix about 10 minutes before I got it from you.

PR #438, opened 2020-12-14
Use cabal-install-3.4.0.0-rc4 on POSIX systems

newhoggy:


erikd:
What do we get from using different versions of cabal on different platforms?

newhoggy:
Build time dropped from `40m` to `5m` on Ubuntu.  MacOS build times are still long because caching is disabled there.  I will make a separate PR to re-enable it.

erikd (approved):

newhoggy:
bors merge

PR #448, opened 2020-12-15
Trevor benson patch 1 voter registration

TrevorBenson:
Fixed the line formatting of the query and removed the uneeded join on tx_metadata.


erikd:
Commits squashed down into a single commit in https://github.com/input-output-hk/cardano-db-sync/pull/451 .

PR #456, opened 2020-12-20
Readme.md: Add System Requirements section

erikd:


erikd:
This has passed all the hydra/buildkite tests and since this is only documentation I will merge it as is.

PR #468, opened 2021-01-07
Add Ledger Validations for the address balance

kderme:


cardano-db-sync-extended/app/cardano-db-sync-extended.hs
```diff
@@ -29,6 +31,7 @@ main = do
   case cmd of
     CmdVersion -> runVersionCommand
     CmdRun params -> runDbSyncNode extendedDbSyncNodePlugin params
+    CmdValidate params -> runValidator params
```
erikd, line 34: Why is this being added here? Is it not possible to do this validation in the `cardano-db-tool` program?

cardano-db-sync/src/Cardano/DbSync/LedgerState.hs
```diff
@@ -305,6 +335,40 @@ ledgerRewardUpdate env lsc =
       LedgerStateAllegra als -> Generic.allegraRewards env als
       LedgerStateMary mls -> Generic.maryRewards env mls
 
+-- Given an address, return it's current UTxO balance.
+ledgerAddrBalance :: Text -> LedgerState (CardanoBlock StandardCrypto) -> Either Text Word64
```
erikd, line 339: If the ledger state was validated in `cardano-db-tool` then then the following could be in the `cardano-db/app/Cardano/Db/App` part of the tree.

cardano-db-sync/src/Cardano/DbSync/Validator.hs
```diff
@@ -0,0 +1,67 @@
+{-# LANGUAGE RecordWildCards #-}
```
erikd, line 1: `RecordWildCards` is potentially bad for readability and is not use anywhere in `db-sync` . It should be removed.

This code in this file also seems to belong in the `cardano-db-tool`  app and be in the `cardano-db/app/Cardano/Db/App` part of the tree.

cardano-db-sync-extended/app/cardano-db-sync-extended.hs
kderme: The new validations added here need to import `Cardano.DbSync.LedgerState` which is found at `cardano-db-sync` package. This would create circular package dependencies, since `cardano-db-sync` already import `cardano-db`

I don't understand the last question: This is what this new validator does: it reads a state from a file and validate that against the db.

erikd: I can totally see the point of moving `cardano-db-tool` from the `cardano-db` to the `cardano-db-sync` package (as a separate PR).

cardano-db-sync/src/Cardano/DbSync/LedgerState.hs
kderme: moved to `cardano-db-tool/src/Cardano/Db/Tool/Validate/Balance.hs`

kderme:
Tested manualy:
Sync up to any point.

get some utxo info from the current tip:
    `PGPASSFILE=config/pgpass-mainnet cabal new-run cardano-db-tool -- utxo-set --slot-no SLOT`

run validation for an address which has some balance:
    `PGPASSFILE=config/pgpass-mainnet cabal new-run cardano-db-tool -- validate  --skip-simple --config config/mainnet-config.yaml  --state-dir ledger-state/mainnet --address ADDRESS`

PR #473, opened 2021-01-08
db-sync: Fix tx_count for genesis blocks

erikd:


erikd:
On mainnet:
```
cexplorer=# select tx_count from block where epoch_no is NULL ; 
 tx_count 
----------
    14505
```

Issue #478, opened 2021-01-10
Sync stalls at epoch boundary


erikd:
When moving to epoch 241 noticed a > 10 minute stall as the epoch rewards were retrieved and inserted into the database.

xdzurman:
I can confirm. The db sync instances were all they were deleting 1 block, the instances that were restarted did not want to delete any block but still took a lot of time. In my case, the total outage was 17 min.

erikd:
> were all saying they were deleting 1 block

I saw this too, but it was **after** the stall and therefore not the cause of the stall. The cause of the stall is that at the epoch boundary there is a large amount of data to be pulled out of the ledger-state and inserted into the database.

xdzurman:
Strangely enough, my instances logged this and they proceeded to be frozen for another several minutes though.

kderme:
I can confirm that epoch boundaries are very slow. Not sure if something changed on Allegra or it was the same before. For me there is an ~20mins slowdown when the epoch changes:
```
[db-sync-node:Info:46] [2021-01-16 05:32:42.40 UTC] insertShelleyBlock(Allegra): epoch 236, slot 16663749, block 5090000, hash 8c3ffbfa77644c4558afed4aaec63095dfc908ff99d7567654d3c752721f4080
[db-sync-node:Info:46] [2021-01-16 05:34:22.94 UTC] insertShelleyBlock(Allegra): epoch 236, slot 16767050, block 5095000, hash 93361fb1ab1a1678fdf80adf63bbed49e9c29aed2e0fbd1f86d8eb3b9cfc3fea
[db-sync-node:Info:46] [2021-01-16 05:35:32.02 UTC] insertShelleyBlock(Allegra): epoch 236, slot 16868599, block 5100000, hash 3d219d321aabba7b7e89a6f34445f52ce1ca26a21e5401ce22a9d60ed1607748
[db-sync-node:Info:46] [2021-01-16 05:36:39.45 UTC] insertShelleyBlock(Allegra): epoch 236, slot 16971050, block 5105000, hash 37ad21a46ac078b569300fb42e6e968f4321acb357ddb4199df3d3f3435bcca4
[db-sync-node:Info:46] [2021-01-16 05:59:00.70 UTC] insertShelleyBlock(Allegra): epoch 237, slot 17072425, block 5110000, hash d07c4ee219d86fd339a34e5fd6fef4109e75b8d467175e6b4877e3413da6ac39
[...]

PR #489, opened 2021-01-18
Insert epoch rewards and stakes in chunks

kderme:
Recent epochs, especially after Allegra cause a very big delay on epoch boundaries as reported https://github.com/input-output-hk/cardano-db-sync/issues/478. For me it's around 20mins for the last epochs.

The main cause of the delay is that we add each reward on a separate query, even if there are >100000 rewards. A better approach is to group these rewards and insert them in chunks. The same applies to stakes, but they created a smaller delay.

With chunks of 10000 the delay drops to 5mins. I tried to add all of them in a single chunk but it was taking 10mins. Maybe it's worth investigating more options.

This is expected to save hours in total sync time.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Rollback.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Rollback.hs
@@ -7,7 +7,7 @@ module Cardano.DbSync.Rollback
 
 import           Cardano.Prelude
 
-import           Cardano.BM.Trace (Trace, logError, logInfo)
+import           Cardano.BM.Trace (Trace, logInfo)
 
 import qualified Cardano.Db as DB
 import           Cardano.DbSync.Error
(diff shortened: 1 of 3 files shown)


cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -31,20 +31,13 @@ rollbackToSlot trce slotNo =
               [ "Rolling back to slot ", textShow (unSlotNo slotNo), ", hash "
               , maybe (if unSlotNo slotNo == 0 then "genesis" else "unknown") renderByteArray mHash
               ]
-        count <- lift $ DB.queryBlocksAfterSlot (unSlotNo slotNo)
-        when (count > 50000) .
-          liftIO . logError trce $
-            mconcat
-              [ "Rollback block count is high and therefore rollback is likely to be incredibly slow. "
-              , "Dropping the database and resyncing from scratch will be faster."
-              ]
```
kderme, line 40: Not slow any more 
```
[db-sync-node:Info:48] [2021-01-18 06:13:16.00 UTC] Deleting slots numbered: [19383774..17882263]
[db-sync-node:Info:48] [2021-01-18 06:14:50.20 UTC] Slots deleted
```

CodiePP (commented):
looking forward to see this landing upstream. I might already cherry-pick this one and have a go :-)
:+1: :+1:

PR #520, opened 2021-02-04
Refactor Environment

kderme:


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -390,12 +400,12 @@ querySelectCount predicate = do
             pure countRows
   pure $ maybe 0 unValue (listToMaybe xs)
 
-querySlotHash :: MonadIO m => SlotNo -> ReaderT SqlBackend m (Maybe (SlotNo, ByteString))
+querySlotHash :: MonadIO m => SlotNo -> ReaderT SqlBackend m [(SlotNo, ByteString)]
```
kderme, line 403: Since we can have multiple blocks in the same slot (EBBs), we return all of them and allow the caller to decide.

kderme:
this should be ready to review

erikd:
Rebased, moved the PR message to the commit and did some minor tweaks.

erikd (approved):

erikd:
Will merge this as soon as I have verified that there is no performance regression.

PR #527, opened 2021-02-11
[CAD-2671] Modify cardano-sync to enable reusing from SMASH.

ksaric:
https://jira.iohk.io/browse/CAD-2671

### DIFF
--- a/cardano-db/cardano-db.cabal
+++ b/cardano-db/cardano-db.cabal
@@ -22,6 +22,8 @@ library
   default-language:     Haskell2010
   hs-source-dirs:       src
   ghc-options:          -Wall
+                        -Werror
+                        -Wunused-imports
                         -Wcompat
                         -fwarn-redundant-constraints
                         -fwarn-incomplete-patterns
@@ -87,6 +89,8 @@ test-suite test
   hs-source-dirs:       test
 
   ghc-options:          -Wall
+                        -Werror
+                        -Wunused-imports
                         -Wcompat
                         -fwarn-redundant-constraints
                         -fwarn-incomplete-patterns
@@ -120,6 +124,8 @@ test-suite test-db
                         Test.IO.Cardano.Db.Util
 
   ghc-options:          -Wall
+                        -Werror
+                        -Wunused-imports
                         -Wcompat
                         -fwarn-redundant-constraints
                         -fwarn-incomplete-patterns
--- a/cardano-sync/src/Cardano/Sync.hs
+++ b/cardano-sync/src/Cardano/Sync.hs
@@ -32,7 +32,6 @@ import           Cardano.Prelude hiding (Meta, Nat, option, (%))
 import           Control.Tracer (Tracer)
 
 import           Cardano.BM.Data.Tracer (ToLogObject (..))
-import qualified Cardano.BM.Setup as Logging
 import           Cardano.BM.Trace (Trace, appendName, logError, logInfo)
 import qualified Cardano.BM.Trace as Logging
 
@@ -42,7 +41,8 @@ import qualified Cardano.Crypto as Crypto
 
 import           Cardano.Sync.Api
 import           Cardano.Sync.Config
-import           Cardano.Sync.Database
+import           Cardano.Sync.Database (DbAction (..), DbActionQueue, lengthDbActionQueue,
+                   mkDbApply, mkDbRollback, newDbActionQueue, runDbStartup, writeDbActionQueue)
 import           Cardano.Sync.Error
 import           Cardano.Sync.Metrics
 import           Cardano.Sync.Plugin (SyncNodePlugin (..))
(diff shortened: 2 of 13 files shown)


cardano-sync/src/Cardano/Sync.hs
```diff
@@ -146,8 +152,15 @@ runSyncNode dataLayer plugin enp insertValidateGenesisDist =
       case genCfg of
           GenesisCardano _ bCfg _sCfg -> do
             syncEnv <- ExceptT $ mkDbSyncEnvFromConfig dataLayer (enpLedgerStateDir enp) genCfg
-            liftIO $ runDbSyncNodeNodeClient (dncPrometheusPort enc) syncEnv
-                iomgr trce plugin (cardanoCodecConfig bCfg) (enpSocketPath enp)
+            liftIO $ runDbSyncNodeNodeClient
```
erikd, line 155: Why change the layout style from what was there. A good rule of thumb when submitting a PR to a project is to follow the existing coding style, eg "When in Rome do as the Romans".

cardano-db/src/Cardano/Db/Schema/Types.hs
```diff
@@ -8,6 +8,8 @@ module Cardano.Db.Schema.Types
   , PaymentAddrHash (..)
   ) where
 
+import           Cardano.Prelude
+
```
erikd, line 12: What benefit does this actually give us?

cardano-db/src/Cardano/Db/Migration/Version.hs
```diff
@@ -23,9 +23,9 @@ data MigrationVersion = MigrationVersion
   } deriving (Eq, Ord, Show)
 
 
-parseMigrationVersionFromFile :: String -> Maybe MigrationVersion
+parseMigrationVersionFromFile :: Text -> Maybe MigrationVersion
```
erikd, line 26: Why are these types changing from `String` to `Text`? It does not seem to be related to "Modify cardano-sync to enable reusing from SMASH:.

cardano-db/src/Cardano/Db/Migration/Version.hs
```diff
@@ -40,18 +40,18 @@ nextMigrationVersion (MigrationVersion _stage ver _date) = do
   (y, m, d) <- Time.toGregorian . Time.utctDay <$> Time.getCurrentTime
   pure $ MigrationVersion 2 (ver + 1) (fromIntegral y * 10000 + m * 100 + d)
 
-renderMigrationVersion :: MigrationVersion -> String
+renderMigrationVersion :: MigrationVersion -> Text
 renderMigrationVersion mv =
-  List.intercalate "-"
+  toS $ intercalate "-"
```
erikd, line 45: Why are we moving from qualified ` List.intercalate` to unqualified `intercalate` ? I know `Cardano.Prelude` exposes `intercalate` but I do not see a good reason to switch from qualified to unqualified usage.

cardano-db/cardano-db.cabal
```diff
@@ -20,6 +20,9 @@ extra-source-files:     CHANGELOG.md
 
 library
   default-language:     Haskell2010
+  default-extensions:
+    NoImplicitPrelude
+    OverloadedStrings
```
erikd, line 25: This seems un-related to the aim of the PR. In general I think defining default extensions in the cabal file is a bad idea. Just about the only one where I think this is not a bad idea is `NoImplicitPrelude` .

cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -65,6 +65,7 @@ module Cardano.Db.Query
   , unValueSumAda
   ) where
 
+import           Cardano.Prelude hiding (Meta, from, isJust, isNothing, maybeToEither, on)
```
erikd, line 68: This change does not offer a net benefit.

cardano-sync/src/Cardano/Sync.hs
ksaric: Fine, I can revert that.

cardano-db/src/Cardano/Db/Schema/Types.hs
ksaric: It's a standard prelude we are supposed to use? I guess I don't have to talk about the benefits of a custom Prelude?

cardano-db/src/Cardano/Db/Migration/Version.hs
ksaric: Well. 
First, I noticed that it would be better if I add `cardano-prelude` in a single module.
Then, there was a collision in that module between `cardano-prelude` and base Prelude.
Then I added the extension for the whole library which I consider worthy, `NoImplicitPrelude`.
Then other modules started complaining.
Then I fixed the other modules.

After a while, I realized it might have been better to do this in another PR, but you don't know how extensive the changes are until you do them.

cardano-db/src/Cardano/Db/Migration/Version.hs
ksaric: https://github.com/input-output-hk/cardano-db-sync/pull/527#discussion_r575045248

cardano-db/cardano-db.cabal
ksaric: https://github.com/input-output-hk/cardano-db-sync/pull/527#discussion_r575045248

cardano-db/src/Cardano/Db/Query.hs
ksaric: https://github.com/input-output-hk/cardano-db-sync/pull/527#discussion_r575045248

cardano-db/cardano-db.cabal
ksaric: I think that using `Text` is what you want to do, in most cases, so this is something that I used quite a lot. Additionally, the current `cardano-prelude` does not export `String`. So there is that.

cardano-db/src/Cardano/Db/Schema/Types.hs
erikd: > It's a standard prelude we are supposed to use? 

No, there is no mandate.

erikd (changes requested):
* Rebase against master.
* The change that includes `Cardano.Prelude` everywhere needs to be split out into a separate commit.
* Remove language pragmas from cabal files and put them in the files that need then (only).
* Fix all un-used import warnings (due ot use of `Cardano.Prelude`).

ksaric:
I rebased against `master` and reverted most of the changes related to the `cardano-prelude` and removed the excess pragmas. Should be ok for now, we can always propagate it through the codebase in later PRs, if so required.

erikd (approved):

PR #530, opened 2021-02-18
fixed: Fail to run migration if postgresql port number is not default (5432)

wei3hua2:
## Expected Behavior

PGPASSFILE file should be able to use port number other than default.

## Current Behavior

Change PGPASSFILE connection string, port number to anything other than default (5432). The application still reference to default port number.

## To reproduce
_config/pgpass-mainnet_
`/var/run/postgresql:5432:cexplorer:*:*`

change to :
`host:<port number other than 5432>:db:username:password`

then execute:
```
PGPASSFILE=config/pgpass-mainnet db-sync-node/bin/cardano-db-sync \
    --config config/mainnet-config.yaml \
    --socket-path ../cardano-node/state-node-mainnet/node.socket \
    --state-dir ledger-state/mainnet \
    --schema-dir schema/

```


erikd:
Closing in favour of #530 (which will run through CI).

PR #533, opened 2021-02-24
db: Improve the way database inserts are done

erikd:


cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -142,21 +172,70 @@ data DbInsertException
...
+
+    fieldNames, placeholders :: [Text]
+    (fieldNames, placeholders) =
+      unzip (Util.mkInsertPlaceholders (entityDef (Proxy @record)) escape)
+
+    exceptHandler :: SqlError -> ReaderT SqlBackend m a
+    exceptHandler e =
+      liftIO $ throwIO (DbInsertException vtype e)
+
+escape :: DBName -> Text
```
kderme, line 235: A pity this is not exported

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -142,21 +172,70 @@ data DbInsertException
...
+
+    values :: [PersistValue]
+    values = pvalues ++ [head pvalues]
+
+    pvalues :: [PersistValue]
+    pvalues = map toPersistValue (toPersistFields record)
+
+    fieldNames, placeholders :: [Text]
+    (fieldNames, placeholders) =
+      unzip (Util.mkInsertPlaceholders (entityDef (Proxy @record)) escape)
```
kderme, line 229: Does this work? I see persistent uses qualified names (see `escapeF` instead of `escape`.

kderme (dismissed):
LGTM. Only minor improvements

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -142,21 +172,70 @@ data DbInsertException
...
+        , MonadIO m
+        , PersistEntity record
+        , PersistRecordBackend record SqlBackend
+        )
+    => String -> record -> ReaderT SqlBackend m (Key record)
+insertCheckUnique vtype record = do
+    res <- handle exceptHandler $ rawSql query values
+    case res of
+      [ident] -> pure ident
+      _other -> error $ mconcat [ "insertByReturnKey2: Inserting ", vtype, " failed with ", show res ]
```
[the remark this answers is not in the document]
erikd: Yeah, print string from an earlier version of this fix.  Unfortunately strings are not compiler checked :trollface:

cardano-db/src/Cardano/Db/Insert.hs
erikd: `escapeF` has type `FieldNameDB -> Text` and is not in the current release of `persistent` that we are using.

my version of `escape` has the type `DBName -> Text`, the types fit and the function makes sense.

cardano-db/src/Cardano/Db/Insert.hs
erikd: Yes, and it seems it is changing between `persistent-2.11.0.1` and the next release.

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -142,21 +172,70 @@ data DbInsertException
...
+  where
+    query :: Text
+    query =
+      Text.concat
+        [ "INSERT INTO "
+        , unDBName (entityDB . entityDef $ Just record)
+        , " (", Util.commaSeparated fieldNames
+        , ") VALUES (", Util.commaSeparated placeholders
+        , ") ON CONFLICT ON CONSTRAINT "
+        , unDBName (uniqueDBName . head . entityUniques . entityDef $ Just record)
```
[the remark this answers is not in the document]
erikd: I think I actually need `onlyOneUniqueDef` which [unfortunately is not exported](https://github.com/yesodweb/persistent/issues/1194). I will cargo cult that in for now.

erikd:
For some reason Hydra is not feeding its result back to GH even though the build was good.

Merging this anyway.

PR #537, opened 2021-03-02
Add Prometheus metrics for slot_no

ksaric:
https://jira.iohk.io/browse/CAD-2547

Also removed the direct dependency, Prometheus is optional and we now have the option to simply ignore Prometheus (for example, tests).

@DavidEichmann If you are the owner of `cardano-db-tool`, it currently doesn't start Prometheus.

Also, changed the names of the metrics, so I guess I have to inform DevOps after this is merged.


DavidEichmann:
@ksaric, I'm not the owner of `cardano-db-tool`.

cardano-db-sync/src/Cardano/DbSync/Metrics.hs
```diff
@@ -34,9 +57,8 @@ withMetricsServer port action = do
...
   Metrics
-    <$> registerGauge "db_block_height" mempty
-    <*> registerGauge "remote_tip_height" mempty
-    <*> registerGauge "action_queue_length_pre" mempty
-    <*> registerGauge "action_queue_length_post" mempty
-    <*> registerGauge "action_queue_length_post_write" mempty
+    <$> registerGauge "remote_block_tip_height" mempty
+    <*> registerGauge "database_queue_height" mempty
+    <*> registerGauge "local_block_number" mempty
+    <*> registerGauge "local_slot_number" mempty
```
erikd, line 63: Have you socialised these name changes for existing metrics? Are @iohk-devops aware of them?

cardano-db-sync-extended/app/cardano-db-sync-extended.hs
```diff
@@ -29,7 +30,14 @@ main = do
   cmd <- Opt.execParser opts
   case cmd of
     CmdVersion -> runVersionCommand
-    CmdRun params -> runDbSyncNode extendedDbSyncNodePlugin params
+    CmdRun params -> do
+        let configFile = enpConfigFile params
+        enc <- readSyncNodeConfig configFile
+        let prometheusPort = dncPrometheusPort enc
```
erikd, line 36: We do not need all these `let` bindings?
```
prometheusPort <- dncPrometheusPort <$> readSyncNodeConfig (enpConfigFile params)
```
Removing the let bindings shows that `configFile` and `enc` are un-needed and the only thing we actually need is `prometheusPort`.

cardano-sync/src/Cardano/Sync/Types.hs
```diff
@@ -66,3 +67,14 @@ newtype BlockId = BlockId Int
 -- @Word64@ is valid as well.
 newtype MetaId = MetaId Int
     deriving (Eq, Show)
+
+-- The metrics we use.
+-- PLEASE do not put this into ENV, it's defeating the purpose
+-- of these constructs.
+data MetricsLayer = MetricsLayer
```
erikd, line 74: Why is this a "Layer"?

cardano-sync/src/Cardano/Sync/Util.hs
```diff
@@ -131,3 +132,14 @@ renderSlotList xs
   | length xs < 10 = textShow (map unSlotNo xs)
   | otherwise =
       mconcat [ "[", textShow (unSlotNo $ List.head xs), "..", textShow (unSlotNo $ List.last xs), "]" ]
+
+-- | This is required for testing or when disabling the metrics layer.
+emptyMetricsLayer :: MetricsLayer
+emptyMetricsLayer =
+    MetricsLayer
+        { mlSetNodeBlockHeight = \_ -> pure ()
```
erikd, line 140: `const` ?

cardano-sync/src/Cardano/Sync.hs
```diff
@@ -387,12 +387,20 @@ chainSyncClient dataLayer trce env queryVar metrics latestPoints currentTip acti
...
+                setNodeBlockHeight (withOrigin 0 unBlockNo (getTipBlockNo tip))
+
                 details <- getSlotDetails trce env queryVar (cardanoBlockSlotNo blk)
                 newSize <- atomically $ do
                                 writeDbActionQueue actionQueue $ mkDbApply blk details
                                 lengthDbActionQueue actionQueue
-                Gauge.set (fromIntegral newSize) $ mQueuePostWrite metrics
+
+                -- The action queue size
+                let setQueuePostWrite = mlSetQueuePostWrite metricsLayer
```
erikd, line 401: Please pull out `setQueuePostWrite` , `setNodeBlockHeight` as top level functions.

cardano-sync/src/Cardano/Sync/Types.hs
```diff
@@ -66,3 +67,14 @@ newtype BlockId = BlockId Int
 -- @Word64@ is valid as well.
 newtype MetaId = MetaId Int
     deriving (Eq, Show)
+
+-- The metrics we use.
+-- PLEASE do not put this into ENV, it's defeating the purpose
+-- of these constructs.
```
erikd, line 73: That is not a particularly useful comment.  It tells me do not do something, but now why I should not do it.

How about "Kept as a separate struct because ......."

ksaric:
> @ksaric, I'm not the owner of `cardano-db-tool`.

I know you were creating some tool, not sure where is it actually, so :man_shrugging:

cardano-db-sync/src/Cardano/DbSync/Metrics.hs
ksaric: There is a comment for notifying them. These names are much clearer.

cardano-db-sync-extended/app/cardano-db-sync-extended.hs
ksaric: I prefer a style where 3 lines of code are written down if that means that it will be clearer what the intention is. I can do what you asked as well.

cardano-sync/src/Cardano/Sync/Types.hs
ksaric: Like a `DataLayer`, we used names `Layer` to describe such interfaces. I think the wallet does a similar thing, it's not a first. I think the name fits quite well. It's a layer that can be cascaded (like the caching layer on top of a db layer), and it's an interface, a small surface that "touches" the world.

cardano-sync/src/Cardano/Sync/Util.hs
ksaric: This is much clearer to me than `const`.

cardano-sync/src/Cardano/Sync.hs
ksaric: Not sure what you mean by `top level functions`?

cardano-sync/src/Cardano/Sync/Types.hs
ksaric: Will update.

cardano-db-sync/src/Cardano/DbSync/Metrics.hs
jbgi: I think `remote_block_tip_height` could be rather `remote_tip_block_height`.

But more importantly I'm not sure what would be the new name of `db_block_height`, is it `database_queue_height` or `local_block_number`?

In either case I think the original name was clearer.

FYI, in our alerting we compare `db_block_height` with `cardano_node_metrics_blockNum_int`.

Also what about using similar convention as `cardano-node` ? something like
`cardano_db_sync_DbBlockNum_int` (the `metrics` part is useless, but the application prefix, especially, is useful).

ksaric: > I think remote_block_tip_height could be rather remote_tip_block_height.

Fine, will fix.

> But more importantly I'm not sure what would be the new name of db_block_height, is it database_queue_height or local_block_number?

`db_block_height` is the height of blocks in the database. So the `local_block_number` is the information from the database and `database_queue_height` is the number of `DbAction` in the queue waiting for DB submission (this can be used to track how well the DB is performing and whether there is some congestion).

`db_block_height` it is, then.

The last part changes things. To just to see whether we understand each other, new names would be:
- `remote_block_tip_height` - `cardano_db_sync_node_block_height`
- `database_queue_height` - `cardano_db_sync_db_queue_height`
- `local_block_number` - `cardano_db_sync_db_block_height`
- `local_slot_number` - `cardano_db_sync_db_slot_height`

Does this look ok, @jbgi ?

jbgi: yep, lgtm! thanks.

cardano-sync/src/Cardano/Sync.hs
ksaric: Fixed.

ksaric:
bors try+

cardano-db-sync/src/Cardano/DbSync/Metrics.hs
```diff
@@ -4,25 +4,52 @@
...
+
 import           System.Metrics.Prometheus.Concurrent.RegistryT (RegistryT (..), registerGauge,
                    runRegistryT, unRegistryT)
 import           System.Metrics.Prometheus.Http.Scrape (serveMetricsT)
 import           System.Metrics.Prometheus.Metric.Gauge (Gauge)
+import qualified System.Metrics.Prometheus.Metric.Gauge as Gauge
 
-
+-- |The metrics we use for Prometheus.
 data Metrics = Metrics
```
erikd, line 22: Why do we have both a `Metrics` struct and a `MetricsLayer` ?

And why is it a layer? To me "layer" implies that if we have say 3 layers, `A`, `B` and `C`, that `A` calls into `B` and then `B` calls into `C`. That is not the case here. Its not a layer, its a Metrics component.

cardano-sync/src/Cardano/Sync/Types.hs
```diff
@@ -66,3 +67,16 @@ newtype BlockId = BlockId Int
 -- @Word64@ is valid as well.
 newtype MetaId = MetaId Int
     deriving (Eq, Show)
```
erikd, line 69: What are `MetaId` and `BlockId` . Are they used anywhere else in this repo? I am pretty sure they are not because the names clash with identifiers that are generated by Persisitent in `cardano-db/src/Cardano/Db/Schema.hs`.

erikd: Same applies to `Block` in this same file.

erikd: If I remove `BlockId`, `Meta` and `Block`, everything still compiles. So why are they here?

cardano-sync/src/Cardano/Sync/Types.hs
erikd: The name `MetricsSetters` is a much better name. It removes the confusing "layer" from the name and "setters" describes *exactly* what the components of this struct do.

Yes, naming things in hard. It took me some time to come up with this improvement.

erikd:
> @DavidEichmann If you are the owner of cardano-db-tool, it currently doesn't start Prometheus.

There is no point in having Prometheus metrics in `cardano-db-tool` .

cardano-sync/src/Cardano/Sync/Types.hs
ksaric: They were supposed to be here and used by some other functions in the `CardanoDataLayer`. For now, they are not used anywhere so if you want, you can drop them.

cardano-db-sync/src/Cardano/DbSync/Metrics.hs
ksaric: The term `Layer` has been used throughout the codebase in all these instances. Please, let's use the name `Layer` that is a standard term for these things. A `Setter` is not. 
Examples:
- https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Pool/DB.hs#L15
- https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Network.hs#L15
- https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Transaction.hs#L21

erikd (approved):

PR #539, opened 2021-03-06
Add new AdaPots table

kderme:
the pr creates one entry at the beginning of each epoch, starting from Shelley, for these values from the ledger https://github.com/input-output-hk/cardano-ledger-specs/blob/5bccddb0ef0a1ae270b72fe0782a8ad680ed5852/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs#L448-L456

### DIFF
--- a/cardano-db/src/Cardano/Db/Schema.hs
+++ b/cardano-db/src/Cardano/Db/Schema.hs
@@ -160,6 +160,24 @@ share
     deriving Eq
     deriving Show
 
+  -- A table with all the different types of total balances.
+  -- This is only populated for the Shelley and later eras, and only on epoch boundaries.
+  -- The treasury and rewards fields will be correct for the whole epoch, but all other
+  -- fields change block by block.
+  AdaPots
+    slotNo              Word64              sqltype=uinteger
+    epochNo             Word64              sqltype=uinteger
+    treasury            DbLovelace          sqltype=lovelace
+    reserves            DbLovelace          sqltype=lovelace
+    rewards             DbLovelace          sqltype=lovelace
+    circulation         DbLovelace          sqltype=lovelace
+    deposits            DbLovelace          sqltype=lovelace
+    fees                DbLovelace          sqltype=lovelace
+    blockId             BlockId             OnDeleteCascade
+    UniqueAdaPots       blockId
+    deriving Eq
+    deriving Show
+
   -- -----------------------------------------------------------------------------------------------
   -- A Pool can have more than one owner, so we have a PoolOwner table that references this one.
 
--- a/cardano-sync/src/Cardano/Sync/LedgerState.hs
+++ b/cardano-sync/src/Cardano/Sync/LedgerState.hs
@@ -23,6 +23,9 @@ module Cardano.Sync.LedgerState
 import           Cardano.Binary (DecoderError)
 import qualified Cardano.Binary as Serialize
 
+import           Cardano.Ledger.Shelley.Constraints (UsesTxOut, UsesValue)
+import qualified Cardano.Ledger.Val as Val
+
 import           Cardano.Sync.Config.Types
 import qualified Cardano.Sync.Era.Cardano.Util as Cardano
 import qualified Cardano.Sync.Era.Shelley.Generic.EpochUpdate as Generic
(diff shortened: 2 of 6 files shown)


cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -455,3 +474,29 @@ tickThenReapplyCheckHash cfg block lsb =
...
                   ]
+
+totalAdaPots
+    :: (UsesValue era, UsesTxOut era)
+    => LedgerState (ShelleyBlock era)
+    -> Generic.AdaPots
+totalAdaPots = totalAdaPotsES . nesEs . Shelley.shelleyLedgerState
+
+  -- | cardano-ledger-spec define a similar function, but it is not exported.
+totalAdaPotsES
```
kderme, line 485: same function already exists in cardano-ledger-spec but it's not exported

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -160,6 +160,22 @@ share
     deriving Eq
     deriving Show
 
+  -- A table with all the different types of total balances
+  -- This is only populated when we are close to the tip of the chain and only
+  -- after epoch boundaries.
+  AdaPots
+    blockId             BlockId             OnDeleteCascade
+    slotNo              Word64              sqltype=uinteger
```
erikd, line 168: This should probably have a `DeleteOnCascade` on it. I also wonder if we should not use the epoch number there instead of the slot number.

erikd (commented):
`DeleteOnCascade` and `slot` vs `epoch`.

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -437,6 +446,16 @@ ledgerRewardUpdate env lsc =
       LedgerStateAllegra als -> Generic.allegraRewards (leNetwork env) als
       LedgerStateMary mls -> Generic.maryRewards (leNetwork env) mls
 
+-- We only compute 'AdaPots' for later eras. This is a time consuming
+-- function and we only want to run it on epoch boundaries, close to the chain tip.
```
erikd, line 450: How is the "close to the chain tip" specified?

cardano-db/src/Cardano/Db/Schema.hs
kderme: `blockId` which is the foreign key, does have a `OnDeleteCascade` already.
Also from the blockId we can always find the block and get all the information (epoch or slot). I just replicate the slot here to make it easier to find the latest entry of AdaPots without joins.

cardano-sync/src/Cardano/Sync/LedgerState.hs
kderme: `getAdaPots` returns `Just` only for Allegra and Mary.

erikd: Can this information also be provided for Shelley (ie pre-Allegra) ? If so, we probably should.

cardano-db/src/Cardano/Db/Schema.hs
erikd: Brain fart! Somehow I missed the `OnDeleteCascade` on the `blockId` .

The value in this `AdaPot` should be constant for a whole epoch right? If that is the case, this field should probably be `epochNo` instead of `slotNo`. Is there any relevant different in the query between `epochNo` and `slotNo`?

cardano-sync/src/Cardano/Sync/LedgerState.hs
kderme: Yes it can, but it seems a somewhat expensive function, since it traverses the utxo in ledger state and epoch boundaries are already overloaded. It can be added easily though. For Byron, I haven't checked.

erikd: We can drop it for Byron, ebcasue that will be mostly un-interesting. Worth doing for all Shelley related eras.

cardano-db/src/Cardano/Db/Schema.hs
kderme: > The values in this AdaPot should be constant for a whole epoch right?
It seems the reserve and treasury are constant, but the utxo changes for every block.

erikd: Is it worth dropping the UTxO  component for this table? Would that still be useful?

erikd (approved):

PR #541, opened 2021-03-10
schema: Rename AdaPots column from circulation to utxo

erikd:


cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -486,7 +486,7 @@ totalAdaPots lState =
       { Generic.apTreasury = Shelley._treasury accountState
       , Generic.apReserves = Shelley._reserves accountState
       , Generic.apRewards = rewards
-      , Generic.apCirculation = circulation
+      , Generic.apUtxo = circulation
```
rhyslbw, line 489:
```suggestion
      , Generic.apUtxo = utxo
```

rhyslbw (changes requested):

cardano-sync/src/Cardano/Sync/LedgerState.hs
erikd: done

rhyslbw (approved):

PR #547, opened 2021-03-15
Fix stack with local packages.

ksaric:
Fix the Stack build.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,6 @@
-resolver: https://raw.githubusercontent.com/input-output-hk/cardano-haskell/master/snapshots/cardano-1.25.1.yaml
-compiler: ghc-8.6.5
+resolver: lts-17.4 # We will soon be able to switch to cardano-haskell
 
-#allow-newer: true
+allow-newer: true
 
 packages:
   - cardano-sync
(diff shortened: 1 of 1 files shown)


stack.yaml
```diff
@@ -30,6 +30,161 @@ ghc-options:
...
+    - ntp-client
+    - ouroboros-consensus-mock
+  
+  - git: https://github.com/snoyberg/http-client.git
+    commit: 1a75bdfca014723dd5d40760fad854b3f0f37156
+    subdirs:
+    - http-client
+
+  # Additional deps
   - persistent-2.11.0.1
```
erikd, line 188: Please move these up into the same block as the other dependencies.

cabal.project
```diff
@@ -61,8 +61,8 @@ package ouroboros-consensus-cardano
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-base
-  tag: 4251c0bb6e4f443f00231d28f5f70d42876da055
-  --sha256: 02a61ymvx054pcdcgvg5qj9kpybiajg993nr22iqiya196jmgciv
+  tag: b364d925e0a72689ecba40dd1f4899f76170b894
+  --sha256: 0igb4gnzlwxy1h40vy5s1aysmaa04wypxn7sn67qy6din7ysmad3
```
erikd, line 65: Sorry, why is this being changed? I always keep this `cabal.project` in sync with the one in `cardano-node`.

If you are fixing `stack.yaml` there should not be any changes in `cabal.project`.

erikd (changes requested):

cabal.project
ksaric: Because the newer version of that library breaks the build in stack, so I synced them manually, there should be no significant difference.

ksaric:
```
cardano-binary                   > [4 of 8] Compiling Cardano.Binary.ToCBOR                                
cardano-binary                   >                                                                   
cardano-binary                   > /tmp/stack-9255e5bf5797aedb/cardano-binary-1.5.0/src/Cardano/Binary/ToCBOR.hs:47:31: error:
cardano-binary                   >     Module ‘Data.Functor.Foldable’ does not export ‘Fix(..)’      
cardano-binary                   >    |                                                              
cardano-binary                   > 47 | import Data.Functor.Foldable (Fix(..), cata, unfix)          
cardano-binary                   >    |                               ^^^^^^^                        
cardano-binary                   >                                                                   
cardano-binary                   > /tmp/stack-9255e5bf5797aedb/cardano-binary-1.5.0/src/Cardano/Binary/ToCBOR.hs:47:46: error:
cardano-binary                   >     Module ‘Data.Functor.Foldable’ does not export ‘unfix’        
cardano-binary                   >    |                                                              
cardano-binary                   > 47 | import Data.Functor.Foldable (Fix(..), cata, unfix)          
cardano-binary                   >    |                                              ^^^^^           
cardano-binary                   >                                                                   
```

ksaric: This version of the library is actually newer (so the bump makes sense to me):
- `b364d925e0a72689ecba40dd1f4899f76170b894` - 11.01.2021
- `4251c0bb6e4f443f00231d28f5f70d42876da055` - 16.12.2020

erikd: I already have a branch where I am updating dependencies and I also lock db-sync deps to the same ones as used by node.

By all means, use whatever revision you like in stack but changes to `cabal.project` will only cause me problems.

erikd (dismissed):

Issue #515, opened 2021-01-29
Release binaries


PegasusPool:
Is it possible to include the cardano-db-sync binaries as part of the release (as it's done for cardano-wallet)?

erikd:
I am pretty sure there is already Nix as well as docker recipes for this. Is that not enough?

 How would you propose this is package any binary? `db-sync` is in some ways more difficult to package than the wallet because it has external dependencies (postgresql) as well as having to find somewhere to put the SQL migration files.

Fell-x27:
> Is that not enough?
No. Not every user uses docker and\or nix. Wold be nice if there was choice to download native binaries only.
> because it has external dependencies (postgresql) 
It's not a problem of the project. It's problem of user who didn't read docs. Also nobody asks to publish binaries **instead** docker\nix. Who needs docker - downloads docker. But many people want just clear binaries.

dmitrystas:
Yes, it would be great if we could just download binaries, like for example https://hydra.iohk.io/build/5446647 for a node

hodlonaut:
I echo the above sentiments, archive with binaries would be much appreciated!

erikd:
The `node` binaries are built as statically linked using Nix. Unfortunately this is not possible for `db-sync` because it needs to link to the PostgreSQL C library which is not easily buildable.

Fell-x27:
>Unfortunately this is not possible for db-sync because it needs to link to the PostgreSQL C library which is not easily buildable.

cardano-wallet is built with a whole SQLite onboard, and there is no problem to distribute it as standalone-binary. 
Moreover, there is a [native haskell psql adapter](https://hackage.haskell.org/package/postgresql-simple). It's just a first one i've found. There should be more I'm sure.

erikd:
What the wallet does with SQLIte is irrelevant. `db-sync` uses PostgreSQL, not SQLIte.

Also, we already use `postgres-simple` which dependes in `postgres-libpq` which binds to and requires the C library.

Fell-x27:
So what's problem? Just release binaries. Developer, who needs binaries only, will install psql and all libraries by himself, it's a simple and common case.

We dont ask you to release all files needed to launch db-sync by one command. Just binaries of db-sync. Without psql, without external libs. We can satisfy dependences, don't worry. We will see runtime errors like "cant load library_name, file doesnt exist" and understand wnat it means. It's not a big deal.

[...]

PR #551, opened 2021-03-19
[CAD-2060] Adding a SMASH executable alongside db-sync.

ksaric:
https://jira.iohk.io/browse/CAD-2060

The added executable that allows SMASH to be run alongside `db-sync`.

Synced up to `Mary` on `mainnet`.
Two things still missing:
- re-fetching of the unavailable metadata, but that can be added easily after we fix the release of `smash`
- fetching of the unique tickers (it was added recently), but we need to create a new release of `smash` and deploy that

Disabled the Github Cabal building only dependencies, because of an issue I run into - https://github.com/input-output-hk/cardano-db-sync/pull/551/files#diff-c09cc8dc973415b1221308559b96942a1c4b872e690bf360fc22d9b42822d519R140

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -137,8 +137,8 @@ jobs:
           cache-cabal-store-v1-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
           cache-cabal-store-v1-${{ runner.os }}-${{ matrix.ghc }}-
 
-    - name: Build dependencies
-      run: cabal build all --builddir="$CABAL_BUILDDIR" --only-dependencies
+          #    - name: Build dependencies
+          #      run: cabal build all --builddir="$CABAL_BUILDDIR" --only-dependencies
 
     - name: Build
       run: cabal build all --builddir="$CABAL_BUILDDIR"
--- a/cabal.project
+++ b/cabal.project
@@ -5,6 +5,7 @@ packages:
   cardano-db/test
   cardano-db-sync
   cardano-db-sync-extended
+  cardano-db-sync-smash
   cardano-db-tool
   cardano-sync
 
@@ -23,6 +24,9 @@ package cardano-db-sync
   ghc-options: -Wall -Werror -Wredundant-constraints -Wincomplete-uni-patterns -Wincomplete-record-updates -Wpartial-fields -Wunused-imports -Wunused-packages
 
 package cardano-db-sync-extended
+  ghc-options: -Wall -Werror -Wredundant-constraints -Wincomplete-uni-patterns -Wincomplete-record-updates -Wpartial-fields -Wunused-imports -Wunused-packages
+
+package cardano-db-sync-smash
   ghc-options: -Wall -Werror -Wredundant-constraints -Wincomplete-uni-patterns -Wincomplete-record-updates -Wpartial-fields -Wunused-imports -Wunused-packages
 
 package cardano-sync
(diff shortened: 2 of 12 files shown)


nix/haskell.nix
```diff
@@ -126,6 +126,7 @@ let
         # systemd can't be statically linked
         packages.cardano-config.flags.systemd = !pkgs.stdenv.hostPlatform.isMusl;
         packages.cardano-node.flags.systemd = !pkgs.stdenv.hostPlatform.isMusl;
+        packages.smash.flags.systemd = !pkgs.stdenv.hostPlatform.isMusl;
```
ksaric, line 129: This is probably absurd and not needed. Can be fixed in the review cycle.

.github/workflows/haskell.yml
```diff
@@ -137,8 +137,8 @@ jobs:
           cache-cabal-store-v1-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
           cache-cabal-store-v1-${{ runner.os }}-${{ matrix.ghc }}-
 
-    - name: Build dependencies
-      run: cabal build all --builddir="$CABAL_BUILDDIR" --only-dependencies
+          #    - name: Build dependencies
```
ksaric, line 140: This was commented out because it raises compile errors. We can simply build everything.

erikd: Compile errors in what? Dependencies?

cabal.project
```diff
@@ -170,6 +174,16 @@ source-repository-package
     plutus-tx
     prettyprinter-configurable
 
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/smash
```
erikd, line 179: Isn't this a circular dependency? `Smash` depends on `db-sync` which depends on `smash`.

nix/haskell.nix
erikd: I do not think it is absurd and it probably is need.

See https://github.com/input-output-hk/cardano-db-sync/issues/515

cabal.project
erikd: We **really** want to avoid circular dependencies.

erikd (changes requested):

cabal.project
erikd: What is the end goal here? I suspect this is not the right first step to achieve that goal.

erikd: The Jira ticket talks about getting stake pool metadata from a SMASH server. That is not what this PR does.

ksaric: This is not a circular dependency. It uses some parts of `db-sync` for the plugin mechanism and uses most/all of `smash` to enable that functionality inside together with `db-smash`. This is what they wanted, access to stake pool metadata and the ability to switch between filtering in seconds.

.github/workflows/haskell.yml
ksaric: Yes. See https://input-output-rnd.slack.com/archives/CAP8NM7N0/p1616416227016800

cabal.project
erikd: It may not be a circular code dependency, but it absolutely is a circular build dedendency.

In this PR, you update the `cabal.project` here in `db-sync` to include things from the `smash` repo. Here in `db-sync` the build depends on things like `cardano-ledger-specs`, but smash also depends on that package. That means that when I try to update the `ledger-specs` dependency for `db-sync`, that new `ledger-specs` version may break the version of `smash` specified by the git hash in the `cabal.project` file. 

This will result in a broken build in the `db-sync` repo. This is why we avoid circular package dependencies.

erikd:
The biggest problem here is that `smash` currently uses a separate database to the one used by `db-sync`. That means that it is not possible to write a query that joins tables maintained by `smash` and `db-sync` .

There is a [stack overflow question](https://stackoverflow.com/questions/4678862/joining-results-from-two-separate-databases) about this and there does not seem to be an easy solution, other than putting everything in the same database.

ksaric:
> The biggest problem here is that `smash` currently uses a separate database to the one used by `db-sync`. That means that it is not possible to write a query that joins tables maintained by `smash` and `db-sync` .
> 
> There is a [stack overflow question](https://stackoverflow.com/questions/4678862/joining-results-from-two-separate-databases) about this and there does not seem to be an easy solution, other than putting everything in the same database.

We are talking about different schemas here. So I don't see an issue with that one.
That query is quite simple - https://stackoverflow.com/questions/1463849/querying-across-schemas-in-postgres

erikd:
Closing this.

PR #554, opened 2021-03-23
Add example SQL queries

Craphtex:
* Get historical UTxO set
* Get tagged genesis addresses

### DIFF
--- a/doc/interesting-queries.md
+++ b/doc/interesting-queries.md
@@ -298,6 +298,69 @@ from the treasury. These can be coalesced into single query via an SQL `union` o
 ...
 ```
 
+### Get historical UTxO set for a given timestamp
+The UTxO set is dependent on time, this will return it for a given timestamp
+```
+# with const as ( select to_timestamp( '2020-10-10 17:00:00', 'YYYY-MM-DD HH24:MI:SS' ) as effective_time_ )
+  select tx_out.address as address, tx_out.value as lovelace, generating_block.time as timestamp
+  from const
+  cross join tx_out
+  inner join tx as generating_tx on generating_tx.id = tx_out.tx_id
+  inner join block as generating_block on generating_block.id = generating_tx.block_id
+  left join tx_in as consuming_input on consuming_input.tx_out_id = generating_tx.id
+    and consuming_input.tx_out_index = tx_out.index
+  left join tx as consuming_tx on consuming_tx.id = consuming_input.tx_in_id
+  left join block as consuming_block on consuming_block.id = consuming_tx.block_id
+  where ( -- Ommit outputs from genesis after Allegra hard fork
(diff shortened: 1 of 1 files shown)


doc/interesting-queries.md
```diff
@@ -298,6 +298,82 @@ from the treasury. These can be coalesced into single query via an SQL `union` o
...
 
+### Get historical UTxO set for a given timestamp
+The UTxO set is dependent on time, this will return it for a given timestamp
+```
+WITH const AS (
+  SELECT TO_TIMESTAMP('2020-10-10 17:00:00', 'YYYY-MM-DD HH24:MI:SS') AS effective_time_
+)
+
+# SELECT tx_out.address        AS address
+    ,    (tx_out.value / 1e6)  AS ada
```
erikd, line 309: If you use `1000000` instead of `1e6` I think the result will be truncated so we avoid things like `1.00000000000000000000 ada` as below.

doc/interesting-queries.md
```diff
@@ -298,6 +298,82 @@ from the treasury. These can be coalesced into single query via an SQL `union` o
...
+    ,    generating_block.time AS timestamp
+
+  FROM const
+  CROSS JOIN tx_out
+  JOIN       tx     AS generating_tx    ON generating_tx.id             = tx_out.tx_id
```
erikd, line 314: Best not to vertically align anything other than indentation. As soon as one of these names change, the alignment wil be off. You should try to match the way I have done it elsewhere in this file.

doc/interesting-queries.md
```diff
@@ -298,6 +298,82 @@ from the treasury. These can be coalesced into single query via an SQL `union` o
...
+ Ae2tdPwUPEYyvoxws84ogrBf7CqRjiGaqP6yKeb2gsLZaKPaDWKdxiQn1Pm |   1838066.000000000000 | 2017-09-23 21:44:51
+...
+```
+
+### Get tagged Genesis addresses
+The genesis block contains multiple addresses, this query will tag them with their origin.
+```
+# SELECT genesis_output.address            AS address
+    ,    FLOOR(genesis_output.value / 1e6) AS ada
+    ,    redemption_block.time             AS redeemed_at
```
erikd, line 346: Similar comments here.

erikd (changes requested):

doc/interesting-queries.md
Craphtex: I tried, but the result is the same, I could leave it in Lovelace instead.

erikd: Yes please.

erikd:
Closing in favour of https://github.com/input-output-hk/cardano-db-sync/pull/555 .

PR #562, opened 2021-03-29
Get time metadata from ledger state

kderme:
https://jira.iohk.io/browse/CAD-2796

### DIFF
--- a/cardano-db-sync-extended/src/Cardano/DbSync/Plugin/Extended.hs
+++ b/cardano-db-sync-extended/src/Cardano/DbSync/Plugin/Extended.hs
@@ -3,7 +3,7 @@ module Cardano.DbSync.Plugin.Extended
   ) where
 
 import           Cardano.DbSync.Plugin.Default (defDbSyncNodePlugin)
-import           Cardano.DbSync.Plugin.Epoch (epochPluginInsertBlock, epochPluginOnStartup,
+import           Cardano.DbSync.Plugin.Epoch (epochPluginInsertBlockDetails, epochPluginOnStartup,
                    epochPluginRollbackBlock)
 
 import           Cardano.Sync (SyncNodePlugin (..))
@@ -19,7 +19,9 @@ extendedDbSyncNodePlugin backend =
               ++ [epochPluginOnStartup backend]
         , plugInsertBlock =
             plugInsertBlock defPlugin
-                ++ [epochPluginInsertBlock backend]
+        , plugInsertBlockDetails =
+            plugInsertBlockDetails defPlugin
+                ++ [epochPluginInsertBlockDetails backend]
         , plugRollbackBlock =
             plugRollbackBlock defPlugin
               ++ [epochPluginRollbackBlock]
--- a/cardano-sync/src/Cardano/Sync.hs
+++ b/cardano-sync/src/Cardano/Sync.hs
@@ -49,8 +49,6 @@ import           Cardano.Sync.Database (DbAction (..), DbActionQueue, lengthDbAc
 import           Cardano.Sync.Error
 import           Cardano.Sync.Metrics
 import           Cardano.Sync.Plugin (SyncNodePlugin (..))
-import           Cardano.Sync.StateQuery (StateQueryTMVar, getSlotDetails, localStateQueryHandler,
-                   newStateQueryTMVar)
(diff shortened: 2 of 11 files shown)


cardano-db-sync-extended/src/Cardano/DbSync/Plugin/Extended.hs
```diff
@@ -19,7 +19,9 @@ extendedDbSyncNodePlugin backend =
               ++ [epochPluginOnStartup backend]
         , plugInsertBlock =
             plugInsertBlock defPlugin
-                ++ [epochPluginInsertBlock backend]
```
erikd, line 22: Does not seem correct for that to be removed.

cardano-sync/src/Cardano/Sync.hs
```diff
@@ -276,9 +268,9 @@ dbSyncProtocols trce env metricsSetters plugin queryVar runDBThreadFunction vers
     localStateQuery :: RunMiniProtocol 'InitiatorMode BSL.ByteString IO () Void
     localStateQuery =
       InitiatorProtocolOnly $ MuxPeer
-        (contramap (Text.pack . show) . toLogObject $ appendName "local-state-query" trce)
+        Logging.nullTracer
         (cStateQueryCodec codecs)
-        (localStateQueryClientPeer (localStateQueryHandler queryVar))
+        localStateQueryPeerNull
```
erikd, line 273: Why? I suspect there is useful logging coming out of those two loggers.

cardano-sync/src/Cardano/Sync/Plugin.hs
```diff
@@ -36,8 +38,8 @@ data SyncNodePlugin = SyncNodePlugin
     -- This will not be called for the original genesis block, but will be called for
     -- all subsequent blocks.
     -- Blocks (including epoch boundary blocks) are called in sequence from the oldest to the newest.
-  , plugInsertBlock :: [Trace IO Text -> SyncEnv -> [BlockDetails] -> IO (Either SyncNodeError ())]
-
+  , plugInsertBlock :: Trace IO Text -> SyncEnv -> [CardanoBlock] -> ExceptT SyncNodeError IO [BlockDetails]
```
erikd, line 41: Why? This seems to reduce the generality of the "plugin" system.

erikd (commented):
A couple of questions on some of the seemingly unrelated changes.

cardano-db-sync-extended/src/Cardano/DbSync/Plugin/Extended.hs
kderme: It is moved to `plugInsertBlockDetails`

cardano-sync/src/Cardano/Sync.hs
kderme: we no longer use the local state query

cardano-sync/src/Cardano/Sync/Plugin.hs
kderme: The idea is that `plugInsertBlock` also creates the `[BlockDetails]`. `plugInsertBlockDetails` makes sure that we don't reduce the generality of the "plugin" system.

ksaric:
Judging from what I see, the issue I have is not related to this change, and this plugin change will be backward compatible with `smash` changes.

cardano-sync/src/Cardano/Sync/Plugin.hs
erikd: What is the difference between `plugInsertBlock` and `plugInsertBlockDetails` ?

Seems the only difference is `CardanoBlock` vs `BlockDetails`. Can these be used interchangably? Can we do without the older version?

kderme: `plugInsertBlock` applies the block to the ledger state and creates the `BlockDetails`. So `plugInsertBlockDetails` has the details available to do any extra work with them.

erikd: Ah ok, now I get it. I will add some comments to make that more clear.

erikd: Maybe `plugInsertBlock` should even be renamed. Maybe something like `pluginPrepareBlocks` .

PR #565, opened 2021-03-31
Update Win32 dependency

newhoggy:


erikd:
Will pull this commit into my branch, Thanks!

PR #567, opened 2021-04-01
sync: Tweak API of mkLedgerEnv function

erikd:


erikd:
I think this is an improvement, but there are currently too many things in progress to make this worth merging.

PR #578, opened 2021-04-12
Generate events when the ledger state is updated 

erikd:


erikd:
I will need to investigate that Travis failure. I may even disable Travis all together.

erikd:
This branch is actually broken. The `reward` and the `stake_distribution` tables are not populated.

Closing this.

PR #579, opened 2021-04-15
 Make all fields strict

kderme:


erikd (dismissed):

erikd (dismissed):

cardano-sync/src/Cardano/Sync/Era/Shelley/Generic/EpochUpdate.hs
```diff
@@ -7,11 +7,13 @@ module Cardano.Sync.Era.Shelley.Generic.EpochUpdate
...
   ) where
 
+import           Prelude hiding (Maybe)
+
 import           Cardano.Sync.Era.Shelley.Generic.ProtoParams
 import           Cardano.Sync.Era.Shelley.Generic.Rewards
 import           Cardano.Sync.Era.Shelley.Generic.StakeDist
 
-import           Data.Maybe (fromMaybe)
+import           Data.Strict.Maybe
```
erikd, line 16: Imports from outside this repo should always be explicit or qualified. It makes it much more obvious where things come from.

erikd (approved):

PR #580, opened 2021-04-16
[CAD-2867] Migrate SMASH DB types to db-sync.

ksaric:
https://jira.iohk.io/browse/CAD-2867


erikd (changes requested):
This should be two PRs. Once to add the changes to the schema and one to add `smash-servant-types`.

smash-servant-types/src/Cardano/SMASH/DBSync/Db/Types.hs
```diff
@@ -0,0 +1,145 @@
...
+import qualified Data.ByteString.Char8 as BS
+
+import           Quiet (Quiet (..))
+
+-- | The stake pool identifier. It is the hash of the stake pool operator's
+-- vkey.
+--
+-- It may be rendered as hex or as bech32 using the @pool@ prefix.
+--
+newtype PoolId = PoolId { getPoolId :: Text }
```
erikd, line 27: This should not be `Text`, it should be `ByteString` because we store the raw hashes in the database.

erikd: Basically `cardano-db` should not be dependent on `smash-servant-types`.

Smash types in the database should be inline with other types in the database and not special Smash types.

ksaric: I already mentioned this. The wallet team depends on the servant types and they use them for their API, so they don't have to depend on the full `smash` repo. So while it might be possible to remove _some_ of the types, they would still need the `smash-servant-types` package.

erikd: The data type presented to the wallet via `smash-servant-types` does not need to be the same as the type in the database.

ksaric: > The data type presented to the wallet via `smash-servant-types` does not need to be the same as the type in the database.

Can you clarify what you mean by that?

erikd: The `cardano-db` package should not depend on `smash-servant-types`. The smash extensions to the database schema should be in terms of existing types in `cardano-db`. When smash presents data to the wallet, it can wrap them in whatever newtypes or whatever makes sense for the wallet.

ksaric: > The `cardano-db` package should not depend on `smash-servant-types`. The smash extensions to the database schema should be in terms of existing types in `cardano-db`. When smash presents data to the wallet, it can wrap them in whatever newtypes or whatever makes sense for the wallet.

To make this clear, you mean - remove `smash-servant-types` and put all the types in the `cardano-db` repo?

erikd: No, I mean the smash db extensions should use `ByteString` for things like hashes, `Text` for human readable strings etc. When presenting these values to the wallet the can be converted to the types the wallet expects.

ksaric: So then I leave `smash-servant-types` as is and switch from `Text` to `ByteString`?

ksaric:
So the `smash-servant-types` package contains some types from the database since they are used both in the DB and the Servant API. Examples can be found here - https://github.com/input-output-hk/smash/blob/master/smash-servant-types/src/Cardano/SMASH/DBSync/Db/Types.hs

You want me to focus on the DB only for now and resolve the Servant API package issues later? 
In other words, for now, include the types in the `cardano-db` and move them from `smash-servant-types` to `cardano-db`?

erikd:
What types do you intend to move into `cardano-db` ?

ksaric:
> What types do you intend to move into `cardano-db` ?

The types in `Cardano.SMASH.DBSync.Db.Types`, which are used in the schema.

cardano-db/src/Cardano/Db/Types.hs
```diff
@@ -133,3 +140,35 @@ showRewardType rt =
 word64ToAda :: Word64 -> Ada
 word64ToAda w =
   Ada (fromIntegral w / 1000000)
+
+-- | The stake pool identifier. It is the hash of the stake pool operator's
+-- vkey.
+--
+-- It may be rendered as hex or as bech32 using the @pool@ prefix.
+newtype DbPoolId = DbPoolId { getPoolId :: Text }
```
erikd, line 148: * In the exisiting `db-sync` code **all** database fields the have a suffix of `Id` are postgres indices. That means we should avoid using the `Id` suffix for anything that is not a postgres index.
* Is that field a hash? If so, why is it `Text` and not `ByteString`? Please tell me you are not storing raw `ByteString`s in the database hex encoded.

cardano-db/src/Cardano/Db/Types.hs
```diff
@@ -133,3 +140,35 @@ showRewardType rt =
...
+-- | The hash of a stake pool's metadata.
+--
+-- It may be rendered as hex.
+newtype DbPoolMetadataHash = DbPoolMetadataHash { getPoolMetadataHash :: Text }
+  deriving stock (Eq, Show, Ord, Generic)
+  deriving newtype PersistField
+
+-- | The stake pool metadata. This type represents it in
+-- its raw original form. The hash of this content is the 'PoolMetadataHash'.
+newtype DbPoolMetadataRaw = DbPoolMetadataRaw { getPoolMetadata :: Text }
```
erikd, line 161: That should *defintiely* be `ByteString`.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -439,14 +439,14 @@ share
...
+    poolId              Types.DbPoolId        sqltype=text
     UniqueDelistedPool poolId
 
   -- A table containing a managed list of reserved ticker names.
   -- For now they are grouped under the specific hash of the pool.
   ReservedTicker
-    name                Types.TickerName        sqltype=text
-    poolHash            Types.PoolMetadataHash  sqltype=text
+    name                Types.DbTickerName        sqltype=text
+    poolHash            Types.DbPoolMetadataHash  sqltype=text
```
erikd, line 449: That should be:
```
    name                Text
    poolHash            ByteString
 ```

File: cardano-db/src/Cardano/Db/Schema.hs [hunk 5]
```diff
@@ -439,14 +439,14 @@ share
 
   -- A table containing a list of delisted pools.
   DelistedPool
-    poolId              Types.PoolId        sqltype=text
+    poolId              Types.DbPoolId        sqltype=text
```
erikd, line 442: I am still not sure what type that field should be. Maybe it should even be an index into the `pool_update` table.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -393,31 +393,31 @@ share
...
 
   PoolMetadata
-    poolId              Types.PoolId              sqltype=text
-    tickerName          Types.TickerName          sqltype=text
-    hash                Types.PoolMetadataHash    sqltype=text
-    metadata            Types.PoolMetadataRaw     sqltype=text
+    poolId              Types.DbPoolId              sqltype=text
+    tickerName          Types.DbTickerName          sqltype=text
+    hash                Types.DbPoolMetadataHash    sqltype=text
+    metadata            Types.DbPoolMetadataRaw     sqltype=text
```
erikd, line 407: `tickerName` should be `Text`. `hash` and probably `metadata` should be `ByteString`.

`poolId` should only be named that if it is an index into the `pool_update` table (which it probably is).

erikd (changes requested):

erikd:
Can we close this? https://github.com/input-output-hk/cardano-db-sync/pull/593 has been merged.

ksaric:
Closing this in favor of https://github.com/input-output-hk/cardano-db-sync/pull/593

PR #583, opened 2021-04-17
Update docker-compose.yml

ggcaponetto:


erikd:
Version `1.26.2` has been tagged but not released yet. When it is released this should be updated.

erikd:
Closing in favor of #588 which bumps the node version to `1.26.2`.

PR #584, opened 2021-04-18
fix rollbacks

kderme:


erikd:
Rebased, squashed, and improved commit message from @kderme .

erikd:
Still testing this. Will keep on rebasing as things are merged to master.

erikd:
Have had one crash running this branch (following the chain). I was not able to see the actual error because it kept on retrying and failing and the real error was not in the scrollback.

erikd:
Ok, this thing crashed again:
```
[2021-04-29 02:07:11.39 UTC] runDBThread: DbInsertException "Tx: 32"
  ( SqlError {sqlState = "23503"
  , sqlExecStatus = FatalError
  , sqlErrorMsg = "insert or update on table \"tx\" violates foreign key constraint \"tx_block_id_fkey\""
  , sqlErrorDetail = "Key (block_id)=(5899482) is not present in table \"block\"."
  , sqlErrorHint = ""
  })
```
Something is really wrong in this PR. I highly suspect this is due to trying to implement rollbacks at the source end of the `TBQueue` instead of the destination end.

erikd (changes requested):
Need to fix the crash.

kderme:
Indeed there is a race condition. `drainQueue` is not an good enough check to access the db from the main thread.

kderme:
I'm thinking about changing or adding a new lock, but I havent found a good solution yet.

erikd (approved):

PR #601, opened 2021-05-04
[CAD-2895] Move the simple DBSyncPlugin as a separate plugin in db-sync

ksaric:
https://jira.iohk.io/browse/CAD-2895

I have returned the schema back, while trying to do as few changes as possible, doing it with all those changes to the schema is next to impossible.


cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -392,47 +393,60 @@ share
   -- The table containing pools' on-chain reference to its off-chain metadata.
 
   PoolMetadataRef
-    poolId              PoolHashId
-    url                 Text
-    hash                ByteString          sqltype=hash32type
+    poolId              Types.PoolIdentifier      sqltype=text
+    url                 Types.PoolUrl             sqltype=text
+    hash                Types.PoolMetaHash        sqltype=text
```
erikd, line 398: No, the original types were correct.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -392,47 +393,60 @@ share
...
-    metadata            Text
+    poolId              Types.PoolIdentifier      sqltype=text
+    tickerName          Types.TickerName          sqltype=text
+    hash                Types.PoolMetaHash        sqltype=text
+    metadata            Types.PoolMetadataRaw     sqltype=text
     pmrId               PoolMetadataRefId Maybe
     UniquePoolMetadata  poolId hash
 
+  -- The pools themselves (identified by the owner vkey hash)
+
```
erikd, line 412: No, the existing `db-sync` already has tables that contain what you are putting in `Pool` and `RetiredPool`.

The tables are `PoolUpdate` and `PoolRetire`.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -19,6 +19,7 @@ module Cardano.Db.Schema where
 
 import           Cardano.Db.Schema.Orphans ()
 
+import           Cardano.Db.Schema.Types as Types
```
erikd, line 22: My understanding is that the tables required by SMASH have **already** been added and should probably not need to be changed.

erikd (changes requested):
Change requests already submitted.

ksaric:
> I have returned the schema back, while trying to do as few changes as possible, doing it with all those changes to the schema is next to impossible.

Erik, I have already noted this. I tried to do this, but this requires a large number of changes and once I realized it's a rabbit hole that will take a huge amount of time, I switched back to the original SMASH schema.
I'm not saying this couldn't be better, but for now, _please_, let us merge this, and then you can modify this to your heart's content and shape it in any way you want.

erikd:
Sorry, I will not be merging this until its right, because if I do there is not guarantee that it will ever get fixed.  Blocking merge until it is fixed is the only to guarantee it is fixed.

erikd:
This has been superceded by #619 .

PR #611, opened 2021-05-12
Fix active_epoch_no of PoolUpdates

kderme:


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -296,14 +297,23 @@ insertPoolRegister tracer network (EpochNo epoch) txId idx params = do
         ]
 
   poolHashId <- insertPoolHash (Shelley._poolId params)
+
+  let wasPoolRegistered = Set.member (Shelley._poolId params) $ getPoolParams $ lssOldState lStateSnap
```
erikd, line 301: I think this is big enough to be pulled out into a separate function, even if its in a `where` clause.

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -86,6 +88,7 @@ import           Ouroboros.Network.Block (HeaderHash, Point (..))
 import qualified Ouroboros.Network.Point as Point
 
 import qualified Shelley.Spec.Ledger.BaseTypes as Shelley
+import           Shelley.Spec.Ledger.Keys
```
erikd, line 91: All imports from outside this repo should be qualified or explicit.

This makes it easier to figure out where a symbol comes from

PR #632, opened 2021-06-04
Improve behaviour of LedgerNewEpoch event

erikd:


erikd:
This is still being tested/validate. It is unlikely to be merged before Monday.

PR #639, opened 2021-06-09
Make sure we never keep more than 1 ledger state in memory

kderme:
Ledger states have grown to take a lot of memory and we must be very careful with handling them. On rollbacks, db-sync parses a new ledger state from a file. We have to make sure that we don't keep a pointer to the old ledger state while parsing the new ledger state or this can cause big memory spikes.

Rollbacks also happen on startups. We have fixed startups in a slightly different way. Since the first message is always a `MsgRollBackward` we don't parse any ledger state before this message.

spike

After the fix the spike disappears:
no-spike

### DIFF
--- a/cardano-sync/src/Cardano/Sync/LedgerState.hs
+++ b/cardano-sync/src/Cardano/Sync/LedgerState.hs
@@ -58,7 +58,6 @@ import           Cardano.Slotting.Slot (EpochNo (..), SlotNo (..), WithOrigin (.
 import qualified Control.Exception as Exception
 import           Control.Monad.Class.MonadSTM.Strict (StrictTVar, TBQueue, atomically, newTBQueueIO,
                    newTVarIO, readTVar, writeTVar)
-import           Control.Monad.Extra (firstJustM, fromMaybeM)
 
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Char8 as BS
@@ -128,7 +127,7 @@ data LedgerEnv = LedgerEnv
   { leProtocolInfo :: !(Consensus.ProtocolInfo IO CardanoBlock)
   , leDir :: !LedgerStateDir
   , leNetwork :: !Ledger.Network
-  , leStateVar :: !(StrictTVar IO CardanoLedgerState)
+  , leStateVar :: !(StrictTVar IO (Maybe CardanoLedgerState))
   , leEventState :: !(StrictTVar IO LedgerEventState)
   -- The following do not really have anything to do with maintaining ledger
   -- state. They are here due to the ongoing headaches around the split between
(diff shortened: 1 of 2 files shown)


cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -458,10 +440,14 @@ cleanupLedgerStateFiles env slotNo = do
 
 loadLedgerAtPoint :: LedgerEnv -> CardanoPoint -> Bool -> IO (Either [LedgerStateFile] CardanoLedgerState)
 loadLedgerAtPoint env point delFiles = do
+    -- Ledger states are growing to become very big in memory.
+    -- Before parsing the new ledger state we need to make sure the old ledger state
+    -- is or can be garbage collected.
+    writeLedgerState env Nothing
     mst <- findStateFromPoint env point delFiles
```
kderme, line 447: maybe we want to force a gc at this point, since the old ledger state is a very big memory chunk.

kderme: `performMajorGC`?

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -222,6 +216,15 @@ initCardanoLedgerState pInfo = CardanoLedgerState
       { clsState = Consensus.pInfoInitLedger pInfo
       }
 
+-- TODO make this type safe. We make the assumption here that the first message of
+-- the chainsync protocol is 'RollbackTo'.
+readStateUnsafe :: LedgerEnv -> STM CardanoLedgerState
+readStateUnsafe env = do
+    mState <- readTVar $ leStateVar env
+    case mState of
+      Nothing -> panic "ledger state is not found"
```
erikd, line 225: I realize this is still a work in progress, but `panic` message should have the module name in the message so if it ever gets hit we know where it came from.

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -222,6 +216,15 @@ initCardanoLedgerState pInfo = CardanoLedgerState
...
+      Just st -> return st
```
erikd, line 226: The rest of the code uses `pure` instead of `return` in preparation for https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/monad-of-no-return .

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -128,7 +127,7 @@ data LedgerEnv = LedgerEnv
   { leProtocolInfo :: !(Consensus.ProtocolInfo IO CardanoBlock)
   , leDir :: !LedgerStateDir
   , leNetwork :: !Ledger.Network
-  , leStateVar :: !(StrictTVar IO CardanoLedgerState)
+  , leStateVar :: !(StrictTVar IO (Maybe CardanoLedgerState))
```
erikd, line 130: I wonder if we could have an empty or minimal ledger state there instead of having the `Maybe` .

kderme: We read `leStateVar` in 1 places: `applyBlock`. If we have a minimal state at this point it would simply lead to more cryptic errors.

We could try to make this type safe. This would probably need to follow the approach of ouroboros-network with typed protocols and parametrise `LedgerEnv` over the `Nat n`. Probably a pretty big refactoring.

`leStateVar` can be empty in 2 cases: after initiation and if loadLedgerAtPoint returns Left. In both cases we send to the node a `FindIntersect` message. The node replies with the point we should roll back and so `loadLedgerAtPoint` is called again.

cardano-sync/src/Cardano/Sync/LedgerState.hs
```diff
@@ -222,6 +216,15 @@ initCardanoLedgerState pInfo = CardanoLedgerState
       { clsState = Consensus.pInfoInitLedger pInfo
       }
 
+-- TODO make this type safe. We make the assumption here that the first message of
+-- the chainsync protocol is 'RollbackTo'.
+readStateUnsafe :: LedgerEnv -> STM CardanoLedgerState
+readStateUnsafe env = do
+    mState <- readTVar $ leStateVar env
+    case mState of
+      Nothing -> panic "LedgerState.readStateUnsafe: Ledger state is not found"
```
erikd, line 226: :+1:

erikd (approved):

PR #650, opened 2021-06-16
Fix deleting during rollback

kderme:
Currently during rollbacks we delete blocks by their slot. This causes issues with EBBS because they don't have a slot. In practice most EBBS are still deleted because their previous block is deleted and deletes are cascaded. However the previous block is not deleted, if it's the point we actually roll back to. To catch this case we need to manually delete the block right after the block we roll back to. A special instance of this case is where we roll back to genesis.

I think this hasn't been an issue so far, because insert block was lenient and idempotent, so inserting an existing block didn't raise an error.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Rollback.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Rollback.hs
@@ -1,16 +1,20 @@
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications #-}
 module Cardano.DbSync.Rollback
-  ( rollbackToSlot
+  ( rollbackToPoint
   , unsafeRollback
   ) where
 
 import           Cardano.Prelude
+import qualified Data.ByteString.Short as BSS
 
 import           Cardano.BM.Trace (Trace, logInfo)
 
 import qualified Cardano.Db as DB
+
+import           Cardano.DbSync.Era.Util
 
 import           Cardano.Sync.Error
 import           Cardano.Sync.Types
@@ -19,14 +23,15 @@ import           Cardano.Sync.Util
 import           Database.Persist.Sql (SqlBackend)
 
 import           Ouroboros.Consensus.Block.Abstract (ConvertRawHash (..))
+import           Ouroboros.Consensus.HardFork.Combinator.AcrossEras (getOneEraHash)
 
 import           Ouroboros.Network.Block
 import           Ouroboros.Network.Point
 
 -- Rollbacks are done in an Era generic way based on the 'Point' we are
 -- rolling back to.
-rollbackToSlot :: SqlBackend -> Trace IO Text -> CardanoPoint -> IO (Either SyncNodeError ())
-rollbackToSlot backend trce point =
+rollbackToPoint :: SqlBackend -> Trace IO Text -> CardanoPoint -> IO (Either SyncNodeError ())
+rollbackToPoint backend trce point =
     DB.runDbIohkNoLogging backend $ runExceptT action
   where
     action :: MonadIO m => ExceptT SyncNodeError (ReaderT SqlBackend m) ()
@@ -42,9 +47,21 @@ rollbackToSlot backend trce point =
                 ]
           mapM_ (lift . DB.deleteCascadeSlotNo) xs
           liftIO $ logInfo trce "Slots deleted"
+        -- Deleting by slot does not guarantee that all blocks will be deleted,
+        -- because EBBS don't have a slot. In practice most EBBS are still
+        -- deleted because their previous block is deleted and deletes are cascaded.
+        -- However the previous block won't be deleted if it's the point we
+        -- actually roll back to. To catch this case we need to manually delete
+        -- the block right after the block we roll back to.
+        prevId <- liftLookupFail "rollbackToPoint" $ rollbackToId point
+        void $ lift $ DB.deleteCascadeAfter prevId
 
     slotsToDelete Origin = DB.querySlotNos
     slotsToDelete (At sl) = DB.querySlotNosGreaterThan (unSlotNo sl)
+
+    rollbackToId pnt = case getPoint pnt of
+      Origin -> DB.queryGenesis
+      At blk -> DB.queryBlockId (BSS.fromShort $ getOneEraHash $ blockPointHash blk)
 
     msg :: Text
     msg = case getPoint point of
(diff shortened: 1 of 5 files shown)


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -118,6 +119,15 @@ queryAddressBalanceAtSlot addr slotNo = do
                   pure $ sum_ (txout ^. TxOutValue)
         pure $ unValueSumAda (listToMaybe res)
 
+queryGenesis :: MonadIO m => ReaderT SqlBackend m (Either LookupFail BlockId)
+queryGenesis = do
+  res <- select . from $ \ blk -> do
+            where_ (isNothing (blk ^. BlockEpochNo))
```
kderme, line 125: Genesis block is the only block without an EpochNo and it must be unique.

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -7,18 +8,22 @@ module Cardano.DbSync.Rollback
...
+import           Cardano.DbSync.Era.Util
+
 import           Cardano.Sync.Error
 import           Cardano.Sync.Types
 import           Cardano.Sync.Util
 
 import           Database.Persist.Sql (SqlBackend)
 
 import           Ouroboros.Consensus.Block.Abstract (ConvertRawHash (..))
+import           Ouroboros.Consensus.HardFork.Combinator.AcrossEras
```
erikd, line 26: Whatever we are using from that should be qualified or explicit.

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -42,10 +47,22 @@ rollbackToSlot backend trce point =
...
+        -- because EBBS don't have a slot. In practice most EBBS are still
+        -- deleted because their previous block is deleted and deletes are cascaded.
+        -- However the previous block won't be deleted if it's the point we
+        -- actually roll back to. To catch this case we need to manually delete
+        -- the block right after the block we roll back to.
+        prevId <- liftLookupFail "rollbackToSlot" $ rollbackToId point
+        void $ lift $ DB.deleteCascadeAfter prevId
 
     slotsToDelete Origin = DB.querySlotNos
     slotsToDelete (At sl) = DB.querySlotNosGreaterThan (unSlotNo sl)
```
erikd, line 60: Currently, we return a list if slot numbers, and then `mapM_ delete` across the list.

However, have have delete cascade now, right? Would it not work to just delete the minimum slot number and have PostgreSQL do the heavy lifting?

I just tried this near the tip of my database and it seems to work:
```
cexplorer=# select block_no from block where slot_no >= 32320510 ; 
 block_no 
----------
  5860449
  5860450
  5860451
  5860452
(4 rows)

cexplorer=# delete from block where block_no = 5860449 ; 
DELETE 1
cexplorer=# select block_no from block where slot_no >= 32320510 ; 
 block_no 
----------
(0 rows)
```
Restarting `db-sync` after the above operation worked as expected and the deleted blocks were re-insterted.

erikd:
We should also rename:
```
rollbackToSlot :: SqlBackend -> Trace IO Text -> CardanoPoint -> IO (Either SyncNodeError ())
```
to `rollbackToPoint` to more accurate match its function.

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
kderme: Yes it should work, `deleteCascadeAfter` should be enough to delete everything. However I was concerned about performance implications of this. Starting from the bottom of the blocks, means we will rely on postgres to cascade these deletes to possibly millions other rows in a sane way. I can see this causing stack overflows if not done properly.

These are just speculations though, it's probably worth stress-testing this.

erikd: I am pretty sure the query to get the slot numbers to delete, and they are returned in ascending order. Deleting the first should delete them all, leaving the following deletes with nothing to do.

kderme: It looks to me it's in descending order https://github.com/input-output-hk/cardano-db-sync/blob/master/cardano-db/src/Cardano/Db/Query.hs#L443 or you mean something else?

kderme: Actually I benchmarked this and it looks like deleting the first block and letting postgres cascade everything is  faster than what we do in master currently.
```
blocks    | this pr   | master 
47239     | 1m52s     | 2m33s
5419239   | 1h20m     | 2h10m
```
results are taken from full logs below:

rollbacks in this pr (1m52s for 47239 blocks and 1h:
```
[db-sync-node:Info:52] [2021-06-18 13:23:58.25 UTC] Cardano.Db tip is at slot 23323439, block 5419239
[db-sync-node:Info:57] [2021-06-18 13:23:58.25 UTC] Running DB thread
[db-sync-node:Info:57] [2021-06-18 13:23:58.26 UTC] Rolling back to 22359702, hash 7914cf09c65ef87298fe06aeb4e8e01a4362bbb3b73b227dde7c587e757d1cd2
[db-sync-node:Info:57] [2021-06-18 13:23:58.62 UTC] Deleting roughly 47239 blocks up to slot 23323439
[db-sync-node:Info:57] [2021-06-18 13:25:50.47 UTC] Blocks deleted


[db-sync-node:Info:52] [2021-06-18 13:29:18.38 UTC] Cardano.Db tip is at slot 23323439, block 5419239
[db-sync-node:Info:57] [2021-06-18 13:29:18.38 UTC] Running DB thread
[db-sync-node:Info:57] [2021-06-18 13:29:18.41 UTC] Rolling back to genesis
[db-sync-node:Info:57] [2021-06-18 13:29:47.93 UTC] Deleting roughly 5419239 blocks up to slot 23323439
[db-sync-node:Info:57] [2021-06-18 14:48:08.13 UTC] Blocks deleted
```

same rollbacks in master:
```
[db-sync-node:Info:52] [2021-06-18 15:09:03.65 UTC] Cardano.Db tip is at slot 23323439, block 5419239
[db-sync-node:Info:57] [2021-06-18 15:09:03.65 UTC] Running DB thread
[db-sync-node:Info:57] [2021-06-18 15:09:03.66 UTC] Rolling back to 22359702, hash 7914cf09c65ef87298fe06aeb4e8e01a4362bbb3b73b227dde7c587e757d1cd2
[db-sync-node:Info:57] [2021-06-18 15:09:03.94 UTC] Deleting 47239 slots: [23323439..22359715]
[db-sync-node:Info:57] [2021-06-18 15:11:36.41 UTC] Slots deleted

[db-sync-node:Info:52] [2021-06-18 15:22:31.47 UTC] Cardano.Db tip is at slot 23323439, block 5419239
[db-sync-node:Info:57] [2021-06-18 15:22:31.47 UTC] Running DB thread
[db-sync-node:Info:57] [2021-06-18 15:22:31.51 UTC] Rolling back to genesis
[db-sync-node:Info:57] [2021-06-18 15:22:59.53 UTC] Deleting 5419239 slots: [23323439..0]
[db-sync-node:Info:57] [2021-06-18 17:32:03.22 UTC] Slots deleted
```

Issue #657, opened 2021-06-20
Restarting db-sync can result in loss of volatilve data.


erikd:
Data that is in memory (like rewards and stake distribution) and not yet inserted in the database will be lost if `db-sync` is stopped and restarted.

~My mainnet instance is now in epoch `273` but the rewards table for epoch `271` have not been inserted yet.~

~This is probably due to a2fb225ac0fbc3c8ae4c5d1cf63eeabcfc74cbf0 .~

erikd:
When the letdger state provides the set of rewards for an epoch, they are stuck in a queue so that insertion of the reward data can be interleaved with regular blockchain data. If `db-sync` is stopped and restarted, the data in the queue is lost and never gets inserted in the database.

erikd:
There are a number of reasons `db-sync` might be stopped requiring a restart:

* User initiated `control-C`.
* OS initiated stop (eg killed by the OOM killer).
* Machine reboot (for whatever reason).

The easiest and most obvious solution would be the one provided by an improved ledger state API that would not require `db-sync` to maintain its own ledger state.

erikd:
A possible (but somewhat rather fiddly) solution would be to get the original data from the ledger and then immediately store it on disk. When all this volatile data has been inserted into the database, the file can be deleted. On restart, `db-sync` could check if there is a data file and if there is load it and start or continuing inserting.

There are a bunch of problems with this solution. It add yet more complexity, it has a bunch of corner cases that must all be handled correctly and it is likely to be fragile.

kderme:
What about the solution I propose here https://github.com/input-output-hk/cardano-db-sync/issues/665?
In that ticket I meant ledger file snapshots, not the full db snapshots that were recently introduced.
If we never take ledger snapshots when there are volatile data, after a restart, dbsync will rollback to a point that is old enough and the rewards/stake will become available and pushed to the queue again.

erikd:
That will not help because this is a problem we can run into without snapshots. I have seen this when I just stop and restart `db-sync` at the right time.

kderme:
> As an example, if a LedgerRewards even is received and we wait for at least one more block, and then stop/restart db-sync, then all of the rewards that were queued internally will be lost,

Won't dbsync rollback to a previous state and get this reward once again?

erikd:
[...]

PR #679, opened 2021-07-02
Take only 1 ledger snapshot on disk per epoch

kderme:
## Current issue
Fixes https://github.com/input-output-hk/cardano-db-sync/issues/657
This pr fixes an issue where rewards, stakes and orphaned_rewards that stay in memory for some time, could get lost in case of a restart. To fix this issue we make sure to never take a ledger snapshot when there are volatile data in memory. 

## Proposed solution
The simplest way to achieve this is to flush the TBQueue and commit the data to db before taking a ledger snapshot. By doing so, in case of a restart, db-sync rollbacks to a point where ledger and db are consistent with each other and there are no missing data. However we don't want to do this flush when there are still too many data in the TBQueue, or this will cause big downtime. The solution we follow in this pr is to take the ledger snapshot right before the epoch boundary, where most epochs and rewards from the previous epoch are probably already inserted. There may be other alternative solutions here.

## Performance optimization and profiling
Another result of this pr is that we take very little ledger snapshots: 1 per epoch instead of 1 per 2000 blocks that we used to. Rollback times are pretty good, so we don't need to take snapshots that often. Also taking snapshots is an expensive operation which according to profiling, adds to total sync time 
For example

syncing between epoch 217-223 (74 ledger snapshots) snapshoting took 13,2% of sync time and 22,4% allocations (this could mean there are memory spikes during snapshoting, but I haven't verified this)

syncing between epochs 257-259 (27 ledger snapshots) takes 9.8% of sync time and 23,9 of total allocations.

These data show that as epochs advance, snapshotting takes less and less time percentage wise, but it still takes a lot of absolute time. For earslier epochs, eg Byron, snapshotting takes up to 25% of time.

## Following the chain
While following the chain, db-sync used to take snapshots for every block. These suffered from the previous issues (rewards/stakes could get lost) and was time consuming (even though there is a average 20s delay between new blocks). With this pr, while following the chain, we still still follow the same policy regarding disk snapshots (1 per epoch), which fixes the issue with restarts. However the node can also do small rollbacks, coming from the consensus and chain selection. In this case, we don't want to rollback too far (previous epoch boundary). To fix this, we keep multiple ledger snapshots in memory. This is the same approach that the LedgerDB of the node follows. I used the same data structure as the LedgerDB, to avoid logic duplication as much as possible.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -194,7 +194,7 @@ source-repository-package
     word-array
 
 
--- From the node. No idea WTF.
+-- Something in plutus-core requries this.
 source-repository-package
   type: git
   location: https://github.com/Quid2/flat.git
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/Epoch.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/Epoch.hs
@@ -56,15 +56,15 @@ insertEpochInterleaved
      -> ExceptT SyncNodeError (ReaderT SqlBackend m) ()
 insertEpochInterleaved tracer bop =
     case bop of
-      BulkOrphanedRewardChunk epochNo icache orwds ->
+      BulkOrphanedRewardChunk epochNo _ icache orwds ->
         insertOrphanedRewards epochNo icache orwds
-      BulkRewardChunk epochNo icache rwds ->
+      BulkRewardChunk epochNo _ icache rwds ->
         insertRewards epochNo icache rwds
-      BulkRewardReport epochNo rewardCount orphanCount ->
+      BulkRewardReport epochNo _ rewardCount orphanCount ->
         liftIO $ reportRewards epochNo rewardCount orphanCount
-      BulkStakeDistChunk epochNo icache sDistChunk ->
+      BulkStakeDistChunk epochNo _ icache sDistChunk ->
         insertEpochStake tracer icache epochNo sDistChunk
-      BuldStakeDistReport epochNo count ->
+      BulkStakeDistReport epochNo _ count ->
         liftIO $ reportStakeDist epochNo count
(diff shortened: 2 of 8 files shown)


kderme:
There is still a small issue. After a rollback to a point A and after the next block is handled, db-sync will take again a snapshot on point A, overriding the existing checkpoint. It does unnecessary work.

erikd:
> This pr fixes an issue where rewards, stakes and orphaned_rewards that stay in memory for some time, could get lost in case of a restart. To fix this issue we make sure to never take a ledger snapshot when there are volatile data in memory.

This comment seems to be using "restart" interchangable with "take ledger snapshot". Is this fix only for taking a snapshot or does it also work for the restart case?

> With this pr, while following the chain, we still still follow the same policy regarding disk snapshots (1 per epoch), which fixes the issue with restarts. 

How does that fix the issue with restarts?

kderme:
Let me explain this a bit more: Assume we have 4 point A,B,C,D
At A we flush TBQueue and take a ledger snapshot
At B rewards and stakes become available
At C we take a ledger snapshot without first flushing the TBQueue
At D the last rewards and stakes are inserted in the db.

If there is a restart between A and C, it will rollback to point A. There is no loss of data, because rewards and stakes will become available again later at point C. However if there is a restart between C and D, it will rollback to C and some rewards and stakes will get lost. I believe for this case the solution is simple: don't take a snapshot at point C.

> How does that fix the issue with restarts? Is that because it only stores a ledger state snapshot at epoch boundaries?

By not taking a snapshot when there are still volatile data.


> This comment seems to be using "restart" interchangable with "take ledger snapshot". Is this fix only for taking a snapshot or does it also work for the restart case?

This pr aims to fix the issue with restart primarily

PR #682, opened 2021-07-07
Avoid address serialisation roundtrip

kderme:
The TxOut patterns for addresses that we use, do a silent deserialisation. Later we serialise addresses again to insert them in the db. We can avoid this round trip by keeping the initial serialised address

This can be optimised even further. `Generic.renderAddress` does another serialization before transforming to the bech32/base58 format. We could use the serialised address to avoid this. However the cardano-api doesn't make this too easy.


kderme:
I've synced with and without the pr and found no differences on the tx_out table.

erikd:
Pulled into #864 .

PR #698, opened 2021-07-18
Use the latest cardano-db-sync docker image.

tmcgilchrist:


docker-compose.yml
```diff
@@ -38,7 +38,7 @@ services:
         max-file: "10"
 
   cardano-db-sync:
-    image: inputoutput/cardano-db-sync:10.0.0
+    image: inputoutput/cardano-db-sync:10.0.1
```
erikd, line 41: This file is updated on release. If you look at the `release/10.0.x` branch you will see that this has been updated.

`db-sync` should only really be run from release branches where this value is always correct.

erikd:
Sorry. not valid.

PR #700, opened 2021-07-19
Add documentation of the database schema

erikd:


rvl:
Nice idea @erikd.

The SQL schema forms part of the API of cardano-db-sync. I was wondering if it would be worth writing in this doc how stable this API will be, or which bits are likely to change and which bits aren't?

erikd:
@rvl asked:
> I was wondering if it would be worth writing in this doc how stable this API will be, or which bits are likely to change and which bits aren't?

If the schema changes, in a non-compatible way, the major version number changes to indicate that it is not compatible.

PR #704, opened 2021-07-22
db: Improve doc on Reward table and clarify when they are spendable

erikd:


yushih:
@erikd is a reward row already withdrawable when it is inserted by cardano-db-sync (i.e. 2 epochs have already passed since it is earned), or do I have to filter by `epoch_no <= current epoch -2` if I want to sum up *withdrawable* rewards? Thanks!

erikd:
> is a reward row already withdrawable when it is inserted by cardano-db-sync

No! The docs say:
 
> The rewards earned in epoch `N` are added in this  table when they are calculated during epoch `N + 1` but are not spendable until after the start of epoch `N + 2`.

> or do I have to filter by epoch_no <= current epoch -2 if I want to sum up withdrawable rewards?
 
Correct.

> I'm asking this because it seems to me that when a MIR reward is added to treasury or reserve

The reward table currently does not list payments from the treasury or the reserves. I am currently in the process of fixing that.

vsubhuman:
> The reward table currently does not list payments from the treasury or the reserves. I am currently in the process of fixing that and will likely be adding a spendable_epoch column.

@erikd, having spendable-epoch there would be perfect! Clarify plz, am I getting it right that for the newly inserted rows the spendable-epoch would be pointing into the future then?

erikd:
Yes, it an reward entry is calculated for epoch `N`, it will be inserted into the database in epoch `N + 1` and spendable after the start of epoch `N + 2` .

I intend to renamed the current `epoch_no` field to `earned_epoch` and the new field will be `spendable_epoch` .

PR #694, opened 2021-07-14
Store information about Alonzo scripts in the database

kderme:
This fixes https://jira.iohk.io/browse/CAD-3050 and https://jira.iohk.io/browse/CAD-3161
It creates migrations for data related to Alonzo scripts

For each transansaction this adds:
- Number of ExUnits
- How much these ExUnits contribute to the minimum amount of fees of the transaction
- sum of sizes of serialised plutus scripts in the tx

PR #706, opened 2021-07-22
 Store information about Alonzo scripts (without the script_size column) in the database 

kderme:
This is a subset of https://github.com/input-output-hk/cardano-db-sync/pull/694 and fixes https://jira.iohk.io/browse/CAD-3050

The split happens because of https://github.com/input-output-hk/cardano-db-sync/pull/694#issuecomment-882870959

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -95,8 +95,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-crypto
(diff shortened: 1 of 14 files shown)


erikd:
This is the version without the `script_size` field, which I think we need. The fact that `script_size` is `0` for all transactions which actually have scripts. Obviously that needs to be fixed.

kderme:
Replaced by https://github.com/input-output-hk/cardano-db-sync/pull/694

PR #710, opened 2021-07-26
Add Redeemer table and other script related data

kderme:
Introduces a new table for Redeemers

### DIFF
--- a/cardano-db/src/Cardano/Db/Schema.hs
+++ b/cardano-db/src/Cardano/Db/Schema.hs
@@ -19,7 +19,8 @@ module Cardano.Db.Schema where
 
 import           Cardano.Db.Schema.Orphans ()
 
-import           Cardano.Db.Types (DbInt65, DbLovelace, DbWord64, SyncState)
+import           Cardano.Db.Types (DbInt65, DbLovelace, DbWord64, ScriptPurpose, ScriptType,
+                   SyncState)
 
 import           Data.ByteString.Char8 (ByteString)
 import           Data.Int (Int64)
@@ -113,14 +114,13 @@ share
 
     -- New for Alonzo
     validContract       Bool                                    -- False if the contract is invalid, True otherwise.
-    exUnitNumber        Word64              sqltype=uinteger
-    exUnitFee           DbLovelace          sqltype=lovelace
     scriptSize          Word64              sqltype=uinteger
     UniqueTx            hash
 
   StakeAddress          -- Can be an address of a script hash
     hashRaw             ByteString          sqltype=addr29type
     view                Text
+    scriptHash          ByteString Maybe    sqltype=hash28type
     registeredTxId      TxId                OnDeleteCascade     -- Only used for rollback.
     UniqueStakeAddress  hashRaw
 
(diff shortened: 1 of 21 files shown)


kderme:
Redeemers are now linked with the inputs they reference. So one can do queries like (results from alonzo-white testnet):
```
# select * from tx_in  join redeemer on tx_in.redeemer_id = redeemer.id;
  id   | tx_in_id | tx_out_id | tx_out_index | redeemer_id | id  | tx_id |  units_mem  | units_steps | purpose | index 
-------+----------+-----------+--------------+-------------+-----+-------+-------------+-------------+---------+-------
 10213 |    10184 |     10180 |            0 |          10 |  10 | 10184 |    70000000 |    70000000 | spend   |     0
 11801 |    11680 |     11671 |            1 |          31 |  31 | 11680 |   700000000 |   700000000 | spend   |     1
 11821 |    11700 |     11697 |            1 |          32 |  32 | 11700 |   700000000 |   700000000 | spend   |     0
 11934 |    11812 |     11808 |            1 |          36 |  36 | 11812 |   100000000 |   100000000 | spend   |     0
 12107 |    11983 |     11979 |            1 |          39 |  39 | 11983 |       39500 |   204050000 | spend   |     1
 12112 |    11987 |     11985 |            1 |          40 |  40 | 11987 |       15800 |    81620000 | spend   |     1
 12118 |    11992 |     11990 |            1 |          41 |  41 | 11992 |       15800 |    81620000 | spend   |     1
....
 29727 |    29397 |     29294 |            0 |         200 | 200 | 29397 |   100000000 |   100000000 | spend   |     0
 30350 |    30020 |     29790 |            0 |         201 | 201 | 30020 | 10000000000 | 10000000000 | spend   |     0
(75 rows)
```
This is the same number of valid Spend scripts:
```
# select count(*) from redeemer join tx on tx.id = redeemer.tx_id  where purpose = 'spend' and tx.valid_contract = true;
 count 
-------
    75
(1 row)

```

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -113,8 +113,7 @@ share
 
     -- New for Alonzo
     validContract       Bool                                    -- False if the contract is invalid, True otherwise.
-    exUnitFee           DbLovelace          sqltype=lovelace
-    scriptSize          Word64              sqltype=uinteger
+    scriptSize          DbWord64            sqltype=word64type
```
erikd, line 116: I actually think `Word64` would be fine there.

`DbWor64` maps to a PostgreSQL `numeric` type and should only be required for things where all possible values are likely. As for `scriptSize` they simply are not allowed to be that big (worst case the size of a transaction would be too big before there is any chance of that overflowing.

`unitneger` is probably OK, but if you really wanted something mostly guaranteed to to be right, try `Word64` with `sqltype=word63type` which maps for PostgreSQL `bigint` (64 bit integer) with a constraint that it be `>= 0`.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -374,6 +373,13 @@ share
     scriptHash          ByteString Maybe    sqltype=hash28type
     UniqueRedeemer      txId purpose index
 
+  Script
+    txId                TxId                OnDeleteCascade
+    hash                ByteString          sqltype=hash28type
+    type                ScriptType          sqltype=scripttype
+    serialisedSize      DbWord64 Maybe      sqltype=word64type
```
erikd, line 380: Again, for sizes, `Wrod64` and `sqltype=word63type` make more sense.

PR #733, opened 2021-08-04
postgres port declaration not needed

redoracle:
in docker compose declaring the port:
    ports:
      - 5432:5432
 will make the postgres instance reachable publicly  which might lead to security attacks against it.

Removing it will keep the port available for db-sync and other containers running on the same host without compromising the security of it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,8 +14,6 @@ services:
       - postgres_db
     volumes:
       - postgres:/var/lib/postgresql/data
-    ports:
-      - 5432:5432
     restart: on-failure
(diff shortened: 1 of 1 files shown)


erikd:
@rhyslbw ?

rhyslbw (approved):
The compose file primarily serves as an example, in which we're demonstrating access to the DB from the host for development or testing purposes. I think it's acceptable to remove this as the default though.

erikd:
Pulled this into https://github.com/input-output-hk/cardano-db-sync/pull/737 .

PR #751, opened 2021-08-12
db: Fix CollateralTxIn uniqueness constraint

erikd:


erikd:
Connected to the `alonzo-purple` testnet and confirmed that it did crash without this fix and does not crash with it.

erikd:
The fixed version has been tagged `alonzo-purple-1.0.1`, which is simply the `alonzo-purple-1.0` tag with this fix.

PR #757, opened 2021-08-19
postgresql-setup.sh: only dump public schema

jbgi:
since all db-sync object are in public this avoid dumping
objects from other applications using separate schemas.

### DIFF
--- a/nix/nixos/cardano-db-sync-service.nix
+++ b/nix/nixos/cardano-db-sync-service.nix
@@ -192,7 +192,7 @@ in {
         done
         if [ -z ''${SKIP_SNAPSHOT:-} ]; then
           set +e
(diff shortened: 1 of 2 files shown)


nix/nixos/cardano-db-sync-service.nix
```diff
@@ -191,8 +191,15 @@ in {
           fi
         done
         if [ -z ''${SKIP_SNAPSHOT:-} ]; then
+          set +e
           SNAPSHOT_SCRIPT=$(cardano-db-tool prepare-snapshot --state-dir ./ | tail -n 1)
-          ${../../scripts/postgresql-setup.sh} ''${SNAPSHOT_SCRIPT#*scripts/postgresql-setup.sh}
+          res=$?
```
erikd, line 196: I would actually prefer to drop the interactivity of the `prepare-snapshot`.

erikd:
Rebased and pushed.

PR #762, opened 2021-08-20
Also check release branches for dependency commits

newhoggy:


erikd:
Why can't I approve this PR?

erikd (approved):

Issue #759, opened 2021-08-19
Fix `insertCheckUnique` on conflict


kderme:
We should check if we're using `insertCheckUnique` in the correct way or if we should change its behaviour.
Currently `insertCheckUnique` checks if a unique contraint is violated and if it does, it updates a random field. The reason it does this, is because an update is necessary for PostgreSQL to return the Id. However, we should actually update a field which is part of the Unique Key.

erikd:
`insertUnchecked` makes the assumption that there will be no conflict.

> Currently insertUnchecked checks if a unique constraint is violated and if it does, it updates a random field.

What? That is not intended. It should just abort.

kderme:
umm sorry I meant `insertCheckUnique`

erikd:
> Currently `insertCheckUnique` checks if a unique contraint is violated and if it does, it updates a random field.

If this is the case, then it should be possible to add a test to the test suite to show this, right?

erikd:
Looking at the code, it updates the first field (`head`). This may be incorrect if the first field is not a field that is part of the uniqueness constraint.

kderme:
Yes I think this hasn't caused issues by luck: basically because in most tables the first field happens to be part of the Unique Key.

I think I have found a simple solution https://github.com/input-output-hk/cardano-db-sync/pull/763. I'l work on a test

erikd:
Yeah, would really like a test for this, even if it requires a custom database schema to trigger it.

PR #763, opened 2021-08-20
Fix insertCheckUnique

kderme:
Fixes https://github.com/input-output-hk/cardano-db-sync/issues/759

### DIFF
--- a/cardano-db/test/Test/IO/Cardano/Db/Insert.hs
+++ b/cardano-db/test/Test/IO/Cardano/Db/Insert.hs
@@ -21,6 +21,7 @@ tests =
   testGroup "Insert"
     [ testCase "Insert zeroth block" insertZeroTest
     , testCase "Insert first block" insertFirstTest
+    , testCase "Insert twice" insertTwice
     ]
 
 insertZeroTest :: IO ()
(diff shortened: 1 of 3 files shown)


kderme:
Added a test which fails without the fix
```
  Insert
    Insert zeroth block:                                    OK
    Insert first block:                                     OK
    Insert twice:                                           FAIL
      Exception: 0 /= 1
      CallStack (from HasCallStack):
        error, called at test/Test/IO/Cardano/Db/Util.hs:39:25 in main:Test.IO.Cardano.Db.Util
```

cardano-db/test/Test/IO/Cardano/Db/Insert.hs
```diff
@@ -48,6 +49,19 @@ insertFirstTest =
...
+    slid <- insertSlotLeader testSlotLeader
+    bid <- insertBlockChecked (blockZero slid)
+    let adaPots = adaPotsZero bid
+    _ <- insertAdaPots adaPots
+    Just pots0 <- queryAdaPots bid
+    -- Insert with same Unique key, different first field
+    _ <- insertAdaPots (adaPots {adaPotsSlotNo = 1 + adaPotsSlotNo adaPots})
+    Just pots0' <- queryAdaPots bid
+    assertBool (show (adaPotsSlotNo pots0) ++ " /= " ++ show (adaPotsSlotNo pots0'))
+      (adaPotsSlotNo pots0 == adaPotsSlotNo pots0')
```
kderme, line 64: AdaPots triggers the bug, because its first field (Slot) is not part of the Unique Key.

PR #781, opened 2021-09-01
fix(docs): rewardtype

1000101:
This is also present in[ release notes](https://github.com/input-output-hk/cardano-db-sync/releases/tag/11.0.0).


erikd:
The file `doc/schema.md` is actually an auto generated file (from `cardano-db/src/Cardano/Db/Schema.hs`) so I will fix that there.

Closing this.

PR #788, opened 2021-09-04
Update example query - Get the reward history, epoch number

lunarpool:
Column reward.epoch_no does not exist, should be replaced by reward.earned_epoch


erikd:
Closed in favor of https://github.com/input-output-hk/cardano-db-sync/pull/792.

Thanks.

PR #800, opened 2021-09-09
Deduplicate JSON handling to Utils.safeDecodeToJson

sorki:
I've started to copy it one more time for storing
`Timelock` scripts in JSON format so instead decided to make it standalone.

PR #803, opened 2021-09-09
Store scripts

sorki:
Store both `timelock` and `plutus` script contents.

We store `timelock`s as `jsonb` and `plutus` scripts as CBOR encoded `bytea`s.

Contains #800

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
@@ -137,6 +137,8 @@ data TxRedeemer = TxRedeemer
 data TxScript = TxScript
   { txScriptHash :: !ByteString
   , txScriptPlutusSize :: Maybe Word64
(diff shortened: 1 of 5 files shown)


kderme:
Thank you for working on this. We don't store whole serialized scripts, because we don't have yet a good use case for it. Do you have one?

sorki:
I think so, I'm working on PoC of Alonzo explorer that supports scripts / datums. 

Also to be able to build [PAB chain-index compatible API](https://github.com/input-output-hk/plutus/blob/master/plutus-chain-index/src/Plutus/ChainIndex/Api.hs#L24-L29) on top of db-sync.

Being able to view `timelock` scripts is interesting as well.

mmahut:
Having a possibility of viewing the `timelock` scripts would be really useful to verify the integrity of NFTs.

sorki:
I've tested this on testnet/mainnet and it looks good. There's a bit of duplication in timelock script handling but otherwise I think this is ready for review.

erikd:
Closed in favor of #863 so it can be run though full CI.

PR #807, opened 2021-09-11
Fix foreign key

kderme:


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
```diff
@@ -460,7 +460,7 @@ fromAlonzoTx pp (blkIndex, tx) =
 
     encodeData :: Alonzo.Data StandardAlonzo -> ByteString
     encodeData dt = LBS.toStrict $ Aeson.encode $
-      Api.scriptDataToJson Api.ScriptDataJsonDetailedSchema $ Api.fromAlonzoData dt
+      Api.scriptDataToJson Api.ScriptDataJsonNoSchema $ Api.fromAlonzoData dt
```
sorki, line 463: What's the reason for this one? Wouldn't we loose expressiveness regarding constructors?

kderme: `DetailedSchema` doesn't seem very user friendly

sorki: I think this should keep using `DetailedSchema` to have the full representational capacity as outlined in https://github.com/input-output-hk/cardano-node/blob/master/cardano-api/src/Cardano/Api/ScriptData.hs#L238-L299


Quoting the most relevant part

```
It also means any script data can be converted into the JSON and
back without loss. That is we can round-trip the script data via the JSON and
also round-trip schema-compliant JSON via script data.
```

which according to my understanding means that `NoSchema` is lossy as in some cases you can't covert `NoSchema` JSON to original `ScriptData`.

If an explorer or some tool wants to display user-friendly `NoSchema` it can convert `DetailedSchema` to `ScriptData` and then build `NoSchema` one.

erikd:
Have taken the "Fix foreign key" commit in #864. 

Closing this.

PR #808, opened 2021-09-12
db-sync: Handle roll back of PoolMetadataRef before result inserted

erikd:


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Offline.hs
```diff
@@ -80,18 +82,23 @@ insertOfflineResults trce resultQueue = do
...
   where
-    insert :: (MonadBaseControl IO m, MonadIO m) => FetchResult -> ReaderT SqlBackend m ()
+    insert :: (MonadBaseControl IO m, MonadCatch m, MonadIO m) => FetchResult -> ReaderT SqlBackend m ()
     insert fr =
-      case fr of
-        ResultMetadata md -> void $ DB.insertPoolOfflineData md
-        ResultError fe -> void $ DB.insertPoolOfflineFetchError fe
+      -- If a rollback occurs after the PoolMetadataRef is fetched, the insertion of the
+      -- result can fail. In this insert fail case, we simply log and drop the exception.
+      handle (liftIO . logFail) $
```
kderme, line 89: Maybe we could handle the error in specific cases, to avoid ignoring other kind of errors. For example we could check that `sqlErrorMsg` contains `violates foreign key constraint`.

erikd: I hate comparing strings that could change without us knowing, but yeah, I suppose that's probably better.

Updated.

PR #829, opened 2021-09-16
fix constraint error

kderme:


kderme:
In favour of https://github.com/input-output-hk/cardano-db-sync/pull/835

PR #835, opened 2021-09-18
Check foreign key exists before insertion

kderme:


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -588,6 +590,22 @@ queryAdaPots blkId = do
             pure adaPots
   pure $ fmap entityVal (listToMaybe res)
 
+checkPoolHashId :: MonadIO m => PoolHashId -> ReaderT SqlBackend m Bool
+checkPoolHashId phid = do
+  res <- select . from $ \ poolHash -> do
+            where_ (poolHash  ^. PoolHashId ==. val phid)
+            limit 1
+            pure poolHash
```
erikd, line 598: Maybe change that to `pure (poolHash  ^. PoolHashId)` ?

Having `pure poolHash` means Postgres will return the whole record to `db-sync` whereas `pure (poolHash  ^. PoolHashId)` only returns the `id` field.

Also the name `check` is a not very well defined. `exists` is much clearer, even if `existsPoolHashId` reads a but funny.

PR #840, opened 2021-09-20
Take more frequent snapshots

kderme:


erikd (approved):

erikd:
Testing this on epoch `292` and `293` shows that while it might make a difference for earlier epoch when the set of stake addresses was small, it makes only a minor improvement for later epochs.

Raised #860 to explore other options.

erikd (approved):

PR #841, opened 2021-09-21
Update dependencies

erikd:


tmcgilchrist:
Good to be rid of the `coerce` usages, nice work @erikd!

PR #853, opened 2021-09-27
Place TH gitRev into cardano-db-sync project.

tmcgilchrist:


erikd (approved):

erikd:
> You could possibly use on of those
> 
>     * https://hackage.haskell.org/package/githash
> 
>     * https://hackage.haskell.org/package/gitrev

The problem is with the cross-compiler and musl builds. Those two packages will not make that any easier.

PR #866, opened 2021-10-01
Add multisig as a new scripttype enum

kderme:


sorki (dismissed):
LGTM! Thanks for adding this.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
```diff
@@ -139,11 +140,14 @@ data TxRedeemer = TxRedeemer
 
 data TxScript = TxScript
   { txScriptHash :: !ByteString
+  , txScriptType :: ScriptType
   , txScriptPlutusSize :: Maybe Word64
   , txScriptJson :: Maybe ByteString
   , txScriptCBOR :: Maybe ByteString
   }
 
+data ScriptType = MultiSig | Timelock | Plutus
```
erikd, line 149: Why not just use the `ScriptType` defined in `cardano-db/src/Cardano/Db/Types.hs` ?

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -888,9 +888,10 @@ insertScript tracer txId script = do
   where
     scriptType :: DB.ScriptType
     scriptType =
-      case Generic.txScriptPlutusSize script of
-        Nothing -> DB.Timelock
-        Just _ -> DB.Plutus
+      case Generic.txScriptType script of
+        Generic.MultiSig -> DB.MultiSig
+        Generic.Timelock -> DB.Timelock
+        Generic.Plutus -> DB.Plutus
```
erikd, line 894: Then we would not need this.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
kderme: We don't import anything from Cardano.Db in the Generic module, so I thought I'd keep it that way.

erikd: `cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Util.hs` has
```
import qualified Cardano.Db as Db
```

erikd (approved):

PR #880, opened 2021-10-11
Port smash web server

kderme:
Port smash web server to db-sync

This changes the whole logic of providing data. In smash tables could be deleted, updated etc, however in db-sync we follow an append only approach.

In particular this fixes https://github.com/input-output-hk/smash/issues/186

The way we check if a pool is retired is to gather all the related certificates and check the latest.
If it's a register the pool is registered. If it's retired, the pool is retired if we've reached the retirement epoch.

The reserved ticker endpoint is not implemented. This is because it was never used and it needs rework.


erikd:
Nice work @kderme !

erikd (approved):

PR #894, opened 2021-10-18
Rework ticker endpoints

kderme:
The ticker endpoints were never used and were actually unusable. Previous implementation associated tickers to pool metadata hash. It makes more sense to associate tickers to pools, in the sense that a pool owns a ticker.

The idea is that if a pool uses a ticker that is reserved by another pool, its "metadata" endpoint will start returning 404.

### DIFF
--- a/cardano-db/src/Cardano/Db/Schema.hs
+++ b/cardano-db/src/Cardano/Db/Schema.hs
@@ -517,7 +517,7 @@ share
   -- For now they are grouped under the specific hash of the pool.
   ReservedPoolTicker
     name                Text
-    poolId              PoolHashId
+    poolHash            ByteString          sqltype=hash28type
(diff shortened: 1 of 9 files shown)


cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -517,7 +517,7 @@ share
   -- For now they are grouped under the specific hash of the pool.
   ReservedPoolTicker
     name                Text
-    poolId              PoolHashId
+    poolHash            ByteString          sqltype=hash28type
```
erikd, line 520: Why is this? The pool hash can still be obtained via a `join` and using the index adds a small amount of consistency checking.

erikd (approved):

PR #904, opened 2021-10-21
Simplify db-sync

kderme:
This pr
- merges cardano-sync and cardano-db-sync packages.
- Removes SyncDataLayer which was necessary for the previous split
- Removes the Plugin system and merges cardano-db-sync and cardano-db-sync-extended

This is now possible since smash is ported in this repository and uses the same schema as db-sync.

PR #937, opened 2021-11-02
Batch insert Tx data

kderme:
Single commit on top of https://github.com/input-output-hk/cardano-db-sync/pull/904

Optimization which gathers TxIn and MaTxOut in batches. In epoch 297 there are average 36 TxIns per block and 157 MaTxOut, so grouping them together instead of doing multiple inserts makes sence. Doing the same for other tables doesn't contribute much, since they are rare (very little per block). We should be careful to avoid dependencies between these elements within the same block. TxIn and MaTxOut live at the bottom of the tree of dependencies (no other table reference them) so it's safe to use them. The same does not apply to TxOut.


erikd:
This looks good, and will probably make things faster, but I would really to to see some benchmarking to show that it does.

I will probably tackle this tomorrow.

erikd:
Changes in this PR have been subsumed by #971 (already merged).

PR #946, opened 2021-11-11
Use state at the epoch boundary for Ada Pots

kderme:
It was after the first block of the epoch before


nc6:
I don't think this will address your issue (if your issue is wanting the value extracted here to match the sum of rewards over the epoch). At this point, the fees pot will contain the fees both for this and for the previous epoch, since they are only deducted at the epoch boundary.

erikd (changes requested):
I have tested this and it does not fix the problem.

kderme:
This requires a ledger event to work properly

PR #949, opened 2021-11-12
sync: Keep last 5 ledger state files instead of just 2

erikd:


erikd:
Syncing from genesis is such a huge pain in the neck. This should make it necessary less often.

PR #955, opened 2021-11-15
Update dependencies

erikd:


erikd:
Can't fix the git deps until https://github.com/input-output-hk/cardano-node/pull/3367 is merged.

PR #976, opened 2021-12-04
Minor fix to drop_db function to include cascade

chadle-git:
Fix this error:  ./postgresql-setup.sh --recreatedb
ERROR:  cannot drop desired object(s) because other objects depend on them
DETAIL:  function f_mints(character varying) depends on type int65type
HINT:  Use DROP ... CASCADE to drop the dependent objects too.


erikd:
Closed in favor of #1012 .

Issue #785, opened 2021-09-02
Missing testnet config file


javiertc:
Hi, in order to run dbsync on the testnet we have to run this command, but the file `config/testnet-config.yaml` doesn't exist 

```
PGPASSFILE=config/pgpass-testnet scripts/postgresql-setup.sh --createdb
PGPASSFILE=config/pgpass-testnet ./cardano-db-sync-extended-exe \
    --config config/testnet-config.yaml \
    --socket-path ../cardano-node/state-node-testnet/node.socket \
    --state-dir ledger-state/testnet \
    --schema-dir schema/
```

Can I use a copy of mainnet-config.yaml with `NetworkName: testnet` and changing the path for `NodeConfigFile` ?

MrMitch17:
Similar, when running..

`NETWORK=testnet docker-compose up`

This is the error that returns

`cardano-db-sync_1  | [db-sync-node:Info:6] [2021-09-03 20:56:46.39 UTC] Using byron genesis file from: "/nix/store/kax0css4lx3ywihvsgrqjym0jpi20f99-byron-genesis.json"
cardano-db-sync_1  | [db-sync-node:Info:6] [2021-09-03 20:56:46.39 UTC] Using shelley genesis file from: "/nix/store/2xhy92909anynqsvx1b1x153cxwnfmzx-shelley-genesis.json"
cardano-db-sync_1  | [db-sync-node:Info:6] [2021-09-03 20:56:46.39 UTC] Using alonzo genesis file from: "/nix/store/5qpskcxlvakm8d0qf4gg0mhs3ckr264l-alonzo-genesis.json"
cardano-db-sync_1  | [db-sync-node:Info:6] [2021-09-03 20:56:46.41 UTC] NetworkMagic: 1097911063
cardano-db-sync_1  | [db-sync-node:Info:6] [2021-09-03 20:56:46.41 UTC] Inserting Byron Genesis distribution
cardano-db-sync_1  | Error: insertValidateGenesisDist: Genesis data mismatch.
`

MrMitch17:
I think the error I'm experiencing here is related to how the container brings up postgres and its databses.  Its only creating the DB on first start, so if you need to switch between mainnet and testnet, its using the same database and credentials.  So it fails because when you switch, it stays on the first network that you started with.

PR #993, opened 2021-12-17
Add config file for `testnet`

ArturWieczorek:
Config files for `testnet` are present in `cardano-node` so we can add  `testnet-config.yaml`  to `cardano-db-sync` now.


erikd (commented):
I thought there were two testnets, one called `shelley_qa` and one called `testnet` ?

Which one is this for? If its actually for `testnet` then the PGPASS file name should reflect that,

ArturWieczorek:
@erikd This config file is for `testnet`, **not** for `shelley-qa`. We have currently following `PGPASS` files: 

```
pgpass-mainnet
pgpass-testnet
pgpass-shelley-qa-testnet
```

however we never had config for `testnet` 

Our config files are for:

```
mainnet-config.yaml
shelley-qa-config.json
```

That's why people were raising some tickets for that:

https://github.com/input-output-hk/cardano-db-sync/issues/785


I think that `QA` in the name is already giving that suggestion that it has something to do with being test network.
 
The reason for renaming `pgpass-shelley-qa-testnet` to `pgpass-shelley-qa`  is that it is easier to handle in scripts
where I am using convention  `PGPASSFILE=config/pgpass-${environment}`  and `${environment}-config.yaml` for starting `cardano-db-sync` 

```
PGPASSFILE=config/pgpass-${environment} db-sync-node/bin/cardano-db-sync \
--config config/${environment}-config.yaml \
--socket-path ../cardano-node/${environment}/node.socket \
--state-dir ledger-state/${environment} \
--schema-dir schema/
```

and I don't need to deal with that `testnet` suffix that is only present in `pgpass-shelley-qa-testnet` but it is not a big deal and if you would still like to me change it back to original name then let me know.

PR #1005, opened 2022-01-06
Store transaction extraKeyWitnessess

sorki:
Store transaction extra key witnesses (aka `requiredSigners`) in new `extra_key_witness` table.


erikd:
Closed in favor of #1015 (to run it through CI).

Thanks @sorki !

PR #1006, opened 2022-01-06
Don't take snapshot on epoch change

### LINKED ISSUE #991: Block synchronization stalled due to rollback on epoch boundary
**OS**
Your OS: Nixos 21.03

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 12.0.0
PostgreSQL version: 14

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: Nix

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): Nix

**Problem Report**
During an epoch change from 308 to 309, couple of our instances got stalled because instantly after receiving an epoch change ledger event there was a rollback to a block in previous epoch. This triggers a snapshot condition since epoch changed (again?) and forces a bulk ops queue flush which takes quite long since it just received a new set of bulk inserts for rewards.

Maybe the `on-epoch` [snapshotting path](https://github.com/input-output-hk/cardano-db-sync/blob/release/12.0.x/cardano-db-sync/src/Cardano/DbSync/Plugin/Default.hs#L111-L120) could be dropped (or made optional) since we have the more frequent snapshots since https://github.com/input-output-hk/cardano-db-sync/pull/840 until the queues can be dropped according to https://github.com/input-output-hk/cardano-db-sync/issues/860#issuecomment-931746889.

Periodical snapshots are also guarded by [`isEmptyEpochBulkOps`](https://github.com/input-output-hk/cardano-db-sync/blob/release/12.0.x/cardano-db-sync/src/Cardano/DbSync/Plugin/Default.hs#L123) so the flush/snapshot cannot occur while there are bulk ops in queue.

sorki:
Cross-epoch snapshot path can trigger flush of bulk queue if a rollback
to previous epoch is triggered right after, causing long delays
until normal operation is resumed.

Fixes #991.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Default.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Default.hs
@@ -10,7 +10,7 @@ module Cardano.DbSync.Default
 
 import           Cardano.Prelude
 
-import           Cardano.BM.Trace (Trace, logDebug, logInfo)
+import           Cardano.BM.Trace (Trace, logInfo)
 
 import qualified Cardano.Db as DB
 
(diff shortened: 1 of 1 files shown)


kderme (approved):
lgtm thanks!

erikd:
Closing in favor of #1014 (so it can be run through CI).

erikd:
Love PRs that remove code and maintain or improve functionality.

PR #1014, opened 2022-01-11
Don't take snapshot on epoch change

erikd:
Cross-epoch snapshot path can trigger flush of bulk queue if a rollback
to previous epoch is triggered right after, causing long delays
until normal operation is resumed.

Closes: https://github.com/input-output-hk/cardano-db-sync/issues/991


erikd:
@kderme already approved this in #1006, which is good enough for me to approve this here.

Issue #1016, opened 2022-01-11
[FR]  Add stake_address_id for Pointer address type


dmitrystas:
At the moment table `tx_out` has a `stake_address_id` field as refer to the stake_address table, in case this address has delegation part (stake key). However (as it was found out in this issue https://github.com/input-output-hk/cardano-node/issues/3491) this is only true for Payment address, but not for Pointer address (https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0019/#pointers).  It would be great if DB Sync would also store the key for such addresses.

erikd:
@dmitrystas I think I have found where `db-sync` is doing something wrong, but if I fix this, how would I test it (on mainnet would be fine).

erikd:
For my own reference, from `cardano-ledger/libs/cardano-ledger-core/src/Cardano/Ledger/Credential.hs`:
```
data StakeReference crypto
  = StakeRefBase !(StakeCredential crypto)
  | StakeRefPtr !Ptr
  | StakeRefNull
  deriving (Show, Eq, Generic, NFData, Ord)

type Ix = Word64

-- | Pointer to a slot, transaction index and index in certificate list.
data Ptr
  = Ptr !SlotNo !Ix !Ix
  deriving (Show, Eq, Ord, Generic, NFData, NoThunks)
  deriving (ToCBOR, FromCBOR) via CBORGroup Ptr
```

dmitrystas:
> @dmitrystas I think I have found where `db-sync` is doing something wrong, but if I fix this, how would I test it (on mainnet would be fine).

For this address `addr1g8alec26en95yzvzwpxfukglsw4nx9wrx99p3m8k2drwppvzj2enszcqyu9ps2 ( 41fbfce15acccb420982704c9e591f83ab3315c3314a18ecf65346e0858292b3380b00)` the values should be

1. `41fbfce15acccb420982704c9e591f83ab3315c3314a18ecf65346e085**8292b3380b00**`
2. `8292b3380b00 -> slot_no 4495800, block_index 11,  0`

```
SELECT stake_registration.addr_id
FROM block
LEFT JOIN tx ON tx.block_id = block.id AND tx.block_index = 11
LEFT JOIN stake_registration ON stake_registration.tx_id = tx.id AND cert_index = 0
WHERE block.slot_no = 4495800;
```

erikd:
[...]

PR #1024, opened 2022-01-18
Fix handling of StakeRefPtr

erikd:


erikd:
Currently testing this solution, but so far it seems like it is working,

erikd:
This seems to be working correctly: https://github.com/input-output-hk/cardano-db-sync/issues/1016#issuecomment-1015906075

PR #1034, opened 2022-01-26
Update interesting-queries.md

frog357:
When I tried to use the example I received the following error:
HINT:  Perhaps you meant to reference the column "redeemer.unit_mem".
ERROR:  column redeemer.units_mem does not exist
LINE 1: ...ee as fees, SUM(redeemer.fee) as script_fees, SUM(redeemer.u...

My changes just updated the query to work as intended.


erikd:
Merged. thanks!

PR #1049, opened 2022-02-04
Backport: offline: correct metadata url for retries. #1048 

jbgi:


erikd:
Would prefer to cherry pick the commit from master rather than merge a second PR.

PR #1057, opened 2022-02-10
doc: Explain why for a given block, block.size /= sum (tx.size)

erikd:


cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -564,7 +564,7 @@ schemaDocs =
       BlockBlockNo # "The block number."
       BlockPreviousId # "The Block table index of the previous block."
       BlockSlotLeaderId # "The SlotLeader table index of the creator of this block."
-      BlockSize # "The block size (in bytes)."
+      BlockSize # "The block size (in bytes). Note, this size value is not expected to be the same as the sum the the tx sizes due to oddities in the CBOR encoding."
```
sorki, line 567:
```suggestion
      BlockSize # "The block size (in bytes). Note, this size value is not expected to be the same as the sum of the tx sizes due to oddities in the CBOR encoding."
```

sorki (dismissed):

PR #1070, opened 2022-02-25
Replace On Delete Restrict with Cascade

kderme:
The referenced tables here `pool_hash` and `multi_asset` don't reference any other table themshelves. This means the existing behaviour should not cause any issue, since on normal execution these tables will never be deleted (since no deletion is cascaded from other tables to them). However it's good to have for possible manual deletions.


schema/migration-2-0001-20211003.sql
```diff
@@ -51,7 +51,7 @@ BEGIN
     EXECUTE 'ALTER TABLE "ada_pots" ADD CONSTRAINT "ada_pots_block_id_fkey" FOREIGN KEY("block_id") REFERENCES "block"("id") ON DELETE CASCADE  ON UPDATE RESTRICT' ;
     EXECUTE 'CREATe TABLE "pool_metadata_ref"("id" SERIAL8  PRIMARY KEY UNIQUE,"pool_id" INT8 NOT NULL,"url" VARCHAR NOT NULL,"hash" hash32type NOT NULL,"registered_tx_id" INT8 NOT NULL)' ;
     EXECUTE 'ALTER TABLE "pool_metadata_ref" ADD CONSTRAINT "unique_pool_metadata_ref" UNIQUE("pool_id","url","hash")' ;
-    EXECUTE 'ALTER TABLE "pool_metadata_ref" ADD CONSTRAINT "pool_metadata_ref_pool_id_fkey" FOREIGN KEY("pool_id") REFERENCES "pool_hash"("id") ON DELETE RESTRICT  ON UPDATE RESTRICT' ;
+    EXECUTE 'ALTER TABLE "pool_metadata_ref" ADD CONSTRAINT "pool_metadata_ref_pool_id_fkey" FOREIGN KEY("pool_id") REFERENCES "pool_hash"("id") ON DELETE CASCADE  ON UPDATE RESTRICT' ;
```
erikd, line 54: Shouldn't there be a change to `cardano-db/src/Cardano/Db/Schema.hs` to go along with this?

If is there a bug in the way these migrations are created?

kderme: Indeed it should be done through the Schema.hs.

erikd:
Will sync this from genesis to make sure all is good.

erikd (approved):

PR #1078, opened 2022-03-09
Update SqlTypes

erikd:


erikd:
Currently testing this code will merge when I am satisfied that it does not break anything.

PR #1079, opened 2022-03-09
doc/building-running.md: Add database creation step

erikd:


erikd:
No idea why this failed on MacOS, but this is a documentation only PR, so just going to merge it,

PR #1080, opened 2022-03-11
Ambit claim for ledger event rewards

erikd:


doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+part of regular transactions.
+
+Payments to stake addresses are made for one of following reasons:
+* Staking rewards for being a member of a stake pool.
+* Staking rewards for being an owner of a stake pool.
+* MIR payments from the treasury.
+* MIR payments from the reserves.
+* Refunds of the stake pool deposit payment when a pool is de-registered.
+
+Currently, the ledger provides the following reward related events:
```
JaredCorduan, line 20: It's probably worth noting that the list below is the name of the wrapped events from the Cardano API. The corresponding ledger events are currently nested according to the ledger rule hierarchy (though we may change this is the near future, a clean up is on order)
* `ShelleyLedgerEventTICK . NewEpochEvent . DeltaRewardEvent . RupdEvent`
* `ShelleyLedgerEventTICK . NewEpochEvent . MirEvent`
* `ShelleyLedgerEventTICK . NewEpochEvent . EpochEvent . PoolReapEvent`
* `ShelleyLedgerEventTICK . NewEpochEvent . TotalRewardEvent`

doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+* `LEDeltaReward`
+* `LEMirTransfer`
+* `LERetiredPools`
+* `LETotalRewards`
```
JaredCorduan, line 24:
```suggestion
* `LERewardEvent`
```

doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## The Desired Ledger Event Functionality
+
+* Every event should carry an `EpochNo` field to make debugging and validation on the `db-sync` side
+  easier.
+* A single `LETotalRewards` event should be emitted for every epoch in the Shelley era and later,
+  even if there are no rewards for that epoch (in which case the event reward map will be `mempty`).
+* `LETotalRewards` should include all payments to stake addresses for a given epoch. That means
+  all staking rewards (member and owner), all MIR payments and all stake pool deposit refunds.
```
JaredCorduan, line 37: There is currently no such `LETotalRewards` event in the Cardano API, do you meant the `LERewardEvent`? The MIR payments and pool deposit refunds come from totally different ledger rules, we cannot make the event described here.

doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+* `LEDeltaReward` should only ever contain pool membership or pool ownership rewards.
+* `LEDeltaReward` may contain rewards to stake addresses that have been de-registered.
```
JaredCorduan, line 39: Here are some extra details that might be helpful.

In the Alonzo era, and previous eras:
* any stake address inside the stake distribution snapshot which is deregestered at the time the reward calculation begins will *not* be included in this event.
* any stake address inside the stake distribution snapshot which is deregestered after the reward calculation begins, and is not registered at the time of the epoch boundary, will have it's lovelace given to the treasury.

In the Babbage era:
* any stake address inside the stake distribution snapshot which is not registered at the time of the epoch boundary will have it's lovelace given to the treasury.

doc/ledger-reward-events.md
erikd: > There is currently no such LETotalRewards event in the Cardano API,

`db-sync` currently does not use `cardano-api` for anything ledger event related.

Is there any accomodation `ledger` might make the use of these events by the current one and only consumer of these events easier?

doc/ledger-reward-events.md
erikd: Era specific behavior for any event is a huge pain in the neck for `db-sync`.

doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+* The sum of all `LEDeltaReward`, `LEMirTransfer` and `LERetiredPools` amounts for an epoch should
+  always be the same as the sum of `LETotalRewards` event amounts for that epoch.
+* `LETotalRewards` will not contain rewards to stake addresses that have been de-registered.
```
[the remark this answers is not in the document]
erikd: :cry: Do we have a different idea of the meaning of the word "total"?

doc/ledger-reward-events.md
```diff
@@ -0,0 +1,47 @@
...
+* The `LETotalRewards` must be the last reward related event for a given epoch to be emitted from
```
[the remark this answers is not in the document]
erikd: So I can theoretically get an `LEDeltaReward` for epoch `N` *after* I get the `LETotalRewards` for that same epoch? That does not make a lot of sense. More importantly, it makes the `LETotalRewards` useless from the POV of `db-sync` using it for validation.

erikd:
@JaredCorduan My attempt to ask for what would make my life easier has obviously failed. How about you actually document what you can provide me with?

doc/ledger-reward-events.md
JaredCorduan: > currently does not use cardano-api for anything ledger event related.

oh! Where does `LETotalRewards` come from? It's not from cardano-ledger.

> Is there any accomodation ledger might make the use of these events by the current one and only consumer of these events easier?

Absolutely! We just can't glue these events together in the way you are proposing, not with serious consequences

JaredCorduan:
> @JaredCorduan My attempt to ask for what would make my life easier has obviously failed. How about you actually document what you can provide me with? It would be important to document everything I can rely on and as much as possible about what I cannot rely on. Hopefully the things I can rely on is a lot longer than the list of things I cannot rely on.
> 
> Feel free to add that to the current document, below the text I wrote or add it here as a comment.

I don't think it failed at all! I think we are making great progress! the ledger team's top priority is Babbage right now, but after that we should have time to document the events.

doc/ledger-reward-events.md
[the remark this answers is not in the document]
erikd: The goal of all of these events is to be used by `db-sync` which is currently the only consumer of ledger events. If its not useful for `db-sync` is not useful.

PR #1083, opened 2022-03-15
Fix logging reward map

kderme:
The query `queryEpochRewardTotal` ignores the deposit refunds. However while printing the diff it is not ignored. So if there is some undelared mismath with events, the deposit refunds are also printed even if they don't cause the issue.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs
@@ -105,6 +105,7 @@ queryRewardMap (EpochNo epochNo) = do
         `on` (\(rwd :& saddr) ->
                 rwd ^. Db.RewardAddrId ==. saddr ^. Db.StakeAddressId)
       where_ (rwd ^. Db.RewardSpendableEpoch ==. val epochNo)
+      where_ (not_ $ rwd ^. Db.RewardType ==. val Db.RwdDepositRefund)
       orderBy [desc (saddr ^. Db.StakeAddressHashRaw)]
       pure (saddr ^. Db.StakeAddressHashRaw, rwd ^. Db.RewardType, rwd ^. Db.RewardAmount)
 


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs
```diff
@@ -105,6 +105,7 @@ queryRewardMap (EpochNo epochNo) = do
         `on` (\(rwd :& saddr) ->
                 rwd ^. Db.RewardAddrId ==. saddr ^. Db.StakeAddressId)
       where_ (rwd ^. Db.RewardSpendableEpoch ==. val epochNo)
+      where_ (not_ $ rwd ^. Db.RewardType ==. val Db.RwdDepositRefund)
```
erikd, line 108: Unfortunately, this mean we no longer have a way of validating deposit refunds.

I really do not think this is a good idea.

kderme: I don't think this function validates anything. This runs after the validation has failed.

We already don't validate deposit refunds.

erikd: The validation failure is at least in part due to the pool deposit refunds. One of the validation failures was:
```
 [2022-03-03 05:13:40.43 UTC]   e1581eab598e036f2bb6506204b5ad483e47fb817e880b1adfe5c04f09: [(RwdDepositRefund,DbLovelace 500000000)] /= []
 ```
 where the LHS is from the database and the RHS is from the ledger event.
 
 This used to work.

Comment (kderme) [reply cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs:108]: > The validation failure is at least in part due to the pool deposit refunds. 

It's not really, the deposit refunds appear in the error message, but are not the reason of failure.

See also this existing message in `queryEpochRewardTotal`
```
            -- For ... reasons ... pool deposit refunds are put into the rewards account
            -- but are not considered part of the total rewards for an epoch.
```
This was like this before. We do no validations of deposit refunds.

Comment (erikd) [reply cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs:108]: The validation detects *any* difference between what is in the db and what is provided by ledger.

The comment is almost certainly wrong (the code has gone through a lot of iterations and its easy to update the code while forgetting to update the comments). Before the latest update, this validation passed and now it does not.

Comment (kderme) [reply cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Validate.hs:108]: It's not just a comment. It's also in the query `queryEpochRewardTotal`
```
where_ (not_ $ rwd ^. Db.RewardType ==. val Db.RwdDepositRefund)
```

> Before the latest update, this validation passed and now it does not.

Which update do you mean? This fixes an existing mismatch on master: The validation filters out deposits but the comment doesn't.

erikd: I mean whichever version of `ledger` was used in the `12.0.x` release. This was working correctly until the `legder` people unilaterally changed things.

erikd:
This was merged in PR #1094 .

Issue #1032, opened 2022-01-25
Use connection pools in smash


kderme:
Currently ported smash opens a new connection for every query.
We should probably use connection pools. These are supported in `persistant`. See `runSqlPool`

erikd:
Fixed on PR #1096 .

PR #1096, opened 2022-04-08
Use connection pools in smash

tmcgilchrist:
Fix for https://github.com/input-output-hk/cardano-db-sync/issues/1032

### DIFF
--- a/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
+++ b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
@@ -142,7 +142,7 @@ queryDBSync env = Db.runWithConnectionNoLogging (getDBSyncPGPass env)
 getPoolLayer :: DBSyncEnv -> IO PoolDataLayer
 getPoolLayer env = do
   pgconfig <- orDie Db.renderPGPassError $ newExceptT $ Db.readPGPass (enpPGPassSource $ dbSyncParams env)
-  pool <- runNoLoggingT $ createPostgresqlPool (Db.toConnectionString pgconfig) 1 -- Pool size as config?
+  pool <- runNoLoggingT $ createPostgresqlPool (Db.toConnectionString pgconfig) 1 -- Pool size of 1 for tests
   pure $ postgresqlPoolDataLayer
                     nullTracer
                     pool
--- a/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
+++ b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
@@ -13,7 +13,7 @@ import           Control.Concurrent.STM.TMVar (TMVar, newEmptyTMVarIO, takeTMVar
 import           Control.Exception (SomeException, bracket)
 import           Control.Monad (void)
 import           Control.Monad.Extra (eitherM)
-import           Control.Monad.Logger (NoLoggingT)
+import           Control.Monad.Logger (NoLoggingT, runNoLoggingT)
 import           Control.Monad.Trans.Except.Exit (orDie)
 import           Control.Monad.Trans.Except.Extra (newExceptT, runExceptT)
 
@@ -24,6 +24,7 @@ import           System.Directory (createDirectoryIfMissing, removePathForcibly)
 import           System.FilePath.Posix ((</>))
 import           System.IO.Silently (hSilence)
 
+import           Database.Persist.Postgresql (createPostgresqlPool)
 import           Database.Persist.Sql (SqlBackend)
 
 import           Ouroboros.Consensus.Config (TopLevelConfig)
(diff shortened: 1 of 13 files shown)


cardano-smash-server/src/Cardano/SMASH/Server/Config.hs
```diff
@@ -1,11 +1,19 @@
...
-module Cardano.SMASH.Server.Config where
+module Cardano.SMASH.Server.Config
+  ( ApplicationUser (..)
+  , ApplicationUsers (..)
+  , SmashServerConfig (..)
+  , SmashServerParams(..)
+
+  , defaultSmashPort
+  , paramsToConfig
+  ) where
```
erikd, line 13: :100:

erikd (dismissed):

erikd:
Will merge this when @kderme approves.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
```diff
@@ -136,10 +137,13 @@ getDBSyncPGPass = enpPGPassSource . dbSyncParams
 queryDBSync :: DBSyncEnv -> ReaderT SqlBackend (NoLoggingT IO) a -> IO a
 queryDBSync env = Db.runWithConnectionNoLogging (getDBSyncPGPass env)
 
-getPoolLayer :: DBSyncEnv -> PoolDataLayer
-getPoolLayer env = postgresqlPoolDataLayer
+getPoolLayer :: DBSyncEnv -> IO PoolDataLayer
+getPoolLayer env = do
+  pgconfig <- orDie Db.renderPGPassError $ newExceptT $ Db.readPGPass (enpPGPassSource $ dbSyncParams env)
+  pool <- runNoLoggingT $ createPostgresqlPool (Db.toConnectionString pgconfig) 1 -- Pool size as config?
```
tmcgilchrist, line 143: @kderme hardcoding this as 1 seems the right option for tests.

cardano-smash-server/src/Cardano/SMASH/Server/Config.hs
```diff
@@ -22,25 +31,32 @@ data SmashServerParams = SmashServerParams
...
   , sspConfigFile :: !FilePath -- config is only used for the logging parameters.
   , sspAdminUsers :: !(Maybe FilePath)
+  , sspSmashPool :: !Int
   }
 
 defaultSmashPort :: Int
 defaultSmashPort = 3100
 
+defaultSmashPool :: Int
+defaultSmashPool = 5
```
tmcgilchrist, line 41: Arbitrarily chosen value here @kderme it really depends on how the postgres instance has been setup and how many connections it can handle. I would expect admins to override this value when setting it up.

erikd: :+1:

cardano-db/src/Cardano/Db/Run.hs
```diff
@@ -46,10 +49,11 @@ import           Database.Esqueleto.Experimental (SqlQuery)
 import           Database.Esqueleto.Internal.Internal (Mode (SELECT), SqlSelect, initialIdentState,
                    toRawSql)
 
+import           Data.Pool (Pool (..))
 import           Database.Persist.Postgresql (SqlBackend, openSimpleConn, withPostgresqlConn,
                    withPostgresqlPool)
```
kderme, line 54: Unrelated to this pr, but it seems db-sync already uses connection pools in some places even though I don't find any apparent reason, since there are no parallel queries.

cardano-smash-server/src/Cardano/SMASH/Server/Run.hs
```diff
@@ -16,27 +16,35 @@ import           Network.Wai.Handler.Warp (defaultSettings, runSettings, setBefo
...
-
     let settings =
           setPort (sscSmashPort config) $
           setBeforeMainLoop
             (logInfo trce $ "SMASH listening on port " <> textShow (sscSmashPort config))
           defaultSettings
 
-    runSettings settings =<< mkApp (sscTrace config) poolDataLayer (sscAdmins config)
+    pgconfig <- orDie renderPGPassError $ newExceptT (readPGPass PGPassDefaultEnv)
+    Db.runIohkLogging trce $ withPostgresqlPool (toConnectionString pgconfig) defaultSmashPool $ \pool -> do
```
kderme, line 44: Shouldn't this use `sspPsqlPool config` instead of `defaultSmashPool` to make it actually configurable.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
```diff
@@ -138,10 +139,13 @@ getDBSyncPGPass = enpPGPassSource . dbSyncParams
 queryDBSync :: DBSyncEnv -> ReaderT SqlBackend (NoLoggingT IO) a -> IO a
 queryDBSync env = Db.runWithConnectionNoLogging (getDBSyncPGPass env)
 
-getPoolLayer :: DBSyncEnv -> PoolDataLayer
-getPoolLayer env = postgresqlPoolDataLayer
+getPoolLayer :: DBSyncEnv -> IO PoolDataLayer
+getPoolLayer env = do
+  pgconfig <- orDie Db.renderPGPassError $ newExceptT $ Db.readPGPass (enpPGPassSource $ dbSyncParams env)
+  pool <- runNoLoggingT $ createPostgresqlPool (Db.toConnectionString pgconfig) 1 -- Pool size of 1 for tests
```
kderme, line 145: Could we use `defaultSmashPool` instead of 1 for tests?

kderme (changes requested):

cardano-db/src/Cardano/Db/Run.hs
tmcgilchrist: Changed this to not use a connection pool.

cardano-db/src/Cardano/Db/Run.hs
```diff
@@ -169,16 +172,3 @@ ghciDebugQuery query = do
...
 
--- -------------------------------------------------------------------------------------------------
-
--- from Control.Monad.Logger, wasnt exported
-defaultOutput :: Handle -> Loc -> LogSource -> LogLevel -> LogStr -> IO ()
-defaultOutput h loc src level msg =
-    BS.hPutStr h $ logStrBS loc src level msg
-  where
-    logStrBS :: Loc -> LogSource -> LogLevel -> LogStr -> BS.ByteString
-    logStrBS a b c d = fromLogStr $ defaultLogStr a b c d
```
tmcgilchrist, line 180: This function is available since monad-logger `0.3.36`

kderme (approved):
LGTM!

erikd (approved):

PR #1108, opened 2022-05-03
Bump to cardano-node 1.34.1

sorki:
* Adds dummy `localTxMonitor`
* `fromShelleyAddr` -> `fromShelleyAddrIsSbe`
* `plutus-example` was moved to `plutus-apps`

I have a follow-up PR updating to more recent `node + deps` which is a bit more involved so creating this one first.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -143,13 +143,16 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-node
-  tag: 814df2c146f5d56f8c35a681fe75e85b905aed5d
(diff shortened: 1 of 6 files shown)


sorki:
Already there. Disregard.

kderme:
We don't bump to cardano-node 1.34.1 since it has broken reward events. This is fixed in https://github.com/input-output-hk/cardano-db-sync/pull/1094

sorki:
Yeah, I've noticed that one too late and did some redundant updating. That's what happens when you forget a `git fetch`.

Good job removing bulk ops!

Issue #1081, opened 2022-03-14
Next hard fork support


kderme:
A draft and wip list of new changes for the next hard fork:
- `datum` table needs to be renamed to `plutus_data`. This is already a misnomer, since it contains not only datum values but also redeemer data. Alternatively we could have 2 separate tables `datum` and `redeemer_data`.
- `insertDatum` needs to be called for outputs containing an inline datum.
- `insertScript` needs to be called for outputs containing an inline/reference script.
- A new table `collateral_output`.
- Tx with phase 2 failures, needs to include the `collRet` in the outputs table. It can also take the fees from the `txcoll` field.
- A new table `tx_reference`
- The `tx_out` table probably doesn't need to change. The existing `data_hash` field can be used for inline datum and there is no apparent reason to differentiate between the 2 kind of datums. Alternatively we can use the `data_hash` for hashes and have a nullable reference to `datum` table for inline datums.
- `findScriptHash` needs to be extended to resolve reference inputs. It's not still clear in specs how this will work.

erikd:
Is there not some way to make tests as era independent as possible?

kderme:
Most things don't depend on the era. However the tx constructors in `Cardano.Mock.Forging.Tx.Alonzo` are coupled to a specific era.

One idea is to have a function `Tx Alonzo -> Tx Babbage` which tranlsates all txs to the new era in a default way, so we can pretty fast test it. Then we can add a new module `Cardano.Mock.Forging.Tx.Babbage` with similar constructors to add more era specific tests. After the hard fork we can slowly remove the Alonzo related tests.

PR #1110, opened 2022-05-04
Add CBOR encoded Datum

sorki:
Introduces `txDatumBytes` field
with CBOR serialized Datum.

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx.hs
@@ -151,8 +151,8 @@ data TxScript = TxScript
 
 data TxDatum = TxDatum
   { txDatumHash :: !ByteString
-  , txDatumValue :: !ByteString -- we turn this into json later.
-  , txDatumBytes :: !ByteString
+  , txDatumValue :: ByteString -- we turn this into json later.
+  , txDatumBytes :: ByteString
   }
 
 fromAllegraTx :: (Word64, ShelleyTx.Tx StandardAllegra) -> Tx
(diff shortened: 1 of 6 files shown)


kderme:
2 notes on this:
- There will probably be many changes soon on these fields with https://github.com/input-output-hk/cardano-db-sync/issues/1081
- Many of these serialisations can be avoided, since if a datum is already in db, it doesn't need to be inserted again. So we should first do a query based on the unique hash and do the serialisation only if needed. This is something we can do for many existing fields (mostly datum and scripts) and doesn't apply only to this pr.

sorki:
>    There will probably be many changes soon on these fields with [Next hard fork support #1081](https://github.com/input-output-hk/cardano-db-sync/issues/1081)

Noticed! I think that's fine, this will be especially useful with inline datums.

>  Many of these serialisations can be avoided, since if a datum is already in db, it doesn't need to be inserted again. So we should first do a query based on the unique hash and do the serialisation only if needed. This is something we can do for many existing fields (mostly datum and scripts) and doesn't apply only to this pr.

I've added two commits to address this for scripts and datums.

sorki:
> Many of these serialisations can be avoided, since if a datum is already in db, it doesn't need to be inserted again. So we should first do a query based on the unique hash and do the serialisation only if needed. This is something we can do for many existing fields (mostly datum and scripts) and doesn't apply only to this pr.

Note that it isn't possible to do this in generic way since Persistent generated types have strict fields.

```hs
λ: :i Datum
type Datum :: *
data Datum
  = Datum {datumHash :: !ByteString,
           datumTxId :: <snip>
           datumValue :: !(Maybe Text),
           datumBytes :: !ByteString}
```

It's tempting to also rename `Datum` `value` to `json` so it is similar to e.g. `Script` or `TxMetadata`. Not sure if it's worth the breakage. What do you think?

erikd:
Closed in favor of #1136 .

Issue #1113, opened 2022-05-06
Update docs with information on how to install secp256k1 


ArturWieczorek:
OS: Linux
tag/version:  `pulsing-rewards` 

When testing `pulsing-rewards` tag and trying to build it with `cabal` I ran into issue:

```sh
cabal build all
HEAD is now at 7497a29 Rename forked package
HEAD is now at ad47441e9 Merge #3697
HEAD is now at ccf5bcb99 Merge pull request #4471 from input-output-hk/bezirg/release-cache-costmodel3
HEAD is now at 00ebe72 Added support for aeson-2.0, text-2.0, and GHC 9.2
Warning: Requested index-state 2022-02-18T00:00:00Z is newer than
'hackage.haskell.org'! Falling back to older state (2022-02-17T23:17:49Z).
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: cardano-crypto-class-2.0.0 (user goal)
[__1] rejecting: cardano-crypto-class:+secp256k1-support (conflict: pkg-config
package libsecp256k1-any, not found in the pkg-config database)
[__1] rejecting: cardano-crypto-class:-secp256k1-support (manual flag can only
be changed explicitly)
[__1] fail (backjumping, conflict set: cardano-crypto-class,
cardano-crypto-class:secp256k1-support)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cardano-crypto-class,
cardano-crypto-class:secp256k1-support
```
This was because of missing dependency on my system - `libsecp256k1` library.

Probably this snippet can be copied into [db-sync istruction](https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/building-running.md):
https://github.com/input-output-hk/cardano-node/blob/master/.github/workflows/haskell.yml#L112

PR #1114, opened 2022-05-07
Extract script for checkout and building of secp256k1 library.

tmcgilchrist:
Add documentation for using script in doc/building-running.md and
reuse the script in GH Actions to validate it works. Script should
work on both Linux and OSX.

Fix for https://github.com/input-output-hk/cardano-db-sync/issues/1113


erikd:
Debian already has this library. Is there any reason not to use that?

sorki:
The lib had no official release yet so you might get too old version.

erikd (approved):

PR #1116, opened 2022-05-09
Fix outdated help in scripts/postgresql-setup.sh

sorki:
Noticed this is wrong while grepping repo for `create-migration` example.


kderme (approved):

erikd (approved):

erikd:
Thanks @sorki !

PR #1126, opened 2022-05-27
Update schema migration in stage 3 to version 8

eMCeee89:
Ensure schema migration in stage 3 to version 8 does not fail when involved indexes are already created and hence secure a smooth update to newer db-sync version once it is released. I found it common that people reporting slow rollback times have already created needed indexes.


tmcgilchrist (approved):
Looks sensible to me.

erikd:
Thanks @eMCeee89 . Closing in favor of #1127 so it runs through CI.

erikd:
And #1127 has been merged.

PR #1127, opened 2022-05-29
db: Robustify creation of indices in migrations

erikd:
Ensure schema migration in stage 3 to version 8 does not fail when involved indices
are already created.


erikd:
Already approved by @tmcgilchrist in #1126 .

Issue #905, opened 2021-10-21
Provide an option that disables ledger state


kderme:
db-sync maintains a ledger state internally and applies each block it receives to it. This is done to extract a number of data that are not included in the block, like rewards, stake distribution, deposits refunds and a number of ledger events, which are the byproduct of block application to a ledger.

Ledger has become pretty big in memory and block application takes a big chunk of the total db-sync syncing. At the same time db-sync is used for many different reason, like data analysis, by light wallets, and recently for smash. Many of the above never use the table fields that require the ledger state. Resources can be drastically reduced by running a flavour that doesn't maintain ledger state.

Changes that this will be required to provide this option:
- ledger states disk snapshots will not be necessary
- rewards validation will not be executed, since stake accounting requires ledger state.
- some table fields may need to become nullable. We can't have different schemas, so we will need to fit all options to the same schema.
- We will need good documentation to make sure users know what is and what is not available in the db and what queries are valid, based on the options provided. We already do something like this with the `extended` version.
- Probably we need to merge first https://github.com/input-output-hk/cardano-db-sync/pull/904/files which is a big simplification of the db-sync codebase.

snapshots will still be compatible with all options.

PR #1140, opened 2022-06-06
Disable ledger state option

kderme:
https://github.com/input-output-hk/cardano-db-sync/issues/905
Still requires testing and documentation

Also ports `Add option to disable cache` from a previous pr to avoid conflicts.


erikd:
This was superceded by another PR.

PR #1144, opened 2022-06-07
Update depenencies

kderme:


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Util.hs
```diff
@@ -143,3 +146,6 @@ unScriptHash (Shelley.ScriptHash h) = Crypto.hashToBytes h
 
 unTxHash :: TxId era -> ByteString
 unTxHash (TxId txid) = Crypto.hashToBytes $ Ledger.extractHash txid
+
+unAssetName :: AssetName -> ByteString
```
kderme, line 150: `AssetName` has change from being backed by a ByteString, to being backed by a ShortByteString in ledger.

erikd:
Superceded.

PR #1152, opened 2022-06-09
Handle PlutusV1 and V2 script types

sorki:
Not sure if it's better to extend the enum (makes it clear that this is a breaking change) or have this as a separate field (where there's a risk of mishandling `V2` as `V1` if the old code doesn't consult the new field). 

Opening a working draft for discussion.

Needs
- [ ] Changelog
- [x] Migration

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Alonzo.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Alonzo.hs
@@ -40,6 +40,7 @@ import           Cardano.DbSync.Era.Shelley.Generic.Witness
 
 import qualified Cardano.Ledger.Address as Ledger
 import qualified Cardano.Ledger.Alonzo.Data as Alonzo
+import qualified Cardano.Ledger.Alonzo.Language as Alonzo
 import           Cardano.Ledger.Alonzo.Scripts (ExUnits (..), txscriptfee)
 import qualified Cardano.Ledger.Alonzo.Scripts as Alonzo
 import           Cardano.Ledger.Alonzo.Tx (ValidatedTx (..))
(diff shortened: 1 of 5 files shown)


kderme:
Yes, I think extending the enum seems a bit more reasonable to me, so that we have a single field for the type. Even if some existing queries break.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Alonzo.hs
```diff
@@ -289,8 +291,10 @@ mkTxScript (hsh, script) =
     plutusCborScript =
       case script of
         Alonzo.TimelockScript {} -> Nothing
-        Alonzo.PlutusScript _lang s ->
+        Alonzo.PlutusScript Alonzo.PlutusV1 s ->
           Just . Api.serialiseToCBOR . Api.PlutusScript Api.PlutusScriptV1 $ Api.PlutusScriptSerialised s
+        Alonzo.PlutusScript Alonzo.PlutusV2 s ->
+          Just . Api.serialiseToCBOR . Api.PlutusScript Api.PlutusScriptV2 $ Api.PlutusScriptSerialised s
```
kderme, line 297: Just a personal note, this is a nice catch and intially I thought it fixes a big bug. But it doesn't change the behaviour, because the serialiser actually ignores the version tag.
```
instance IsScriptLanguage lang => SerialiseAsCBOR (Script lang) where
    serialiseToCBOR (SimpleScript SimpleScriptV1 s) =
      CBOR.serialize' (toShelleyMultiSig s)

    serialiseToCBOR (SimpleScript SimpleScriptV2 s) =
      CBOR.serialize' (toAllegraTimelock s :: Timelock.Timelock StandardCrypto)

    serialiseToCBOR (PlutusScript PlutusScriptV1 s) =
      CBOR.serialize' s

    serialiseToCBOR (PlutusScript PlutusScriptV2 s) =
      CBOR.serialize' s
```

kderme:
@sorki this looks good to far, but I'm not sure if it will make it to the next release.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Alonzo.hs
sorki: Good find! I think it's worth a comment.

erikd:
Merged as part of #1155 .

Issue #1135, opened 2022-06-04
db-sync fails to reconnect to the node after node is restarted


erikd:
**OS**
Your OS: Linux

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): branch `erikd/hf-integration` commit `a1f19fcd52895f7788c52c49be613f26fbc5e844`
PostgreSQL version: 14.3

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: `cabal build`

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): none

**Additional context**
Using git commit `ea202b7b21983140d1944ccfde8750b891b59699` of `outoborous-network`

**Problem Report**
If `db-sync` is running and the `nod` is restarted, then `db-sync` failed to reconnect correctly to the node and loops with the following error messages:
```
[db-sync-node:Error:48364] [2022-06-04 07:25:24.80 UTC] ChainSyncWithBlocksPtcl: libpq: failed (another command is already in progress)
[db-sync-node.Mux:Info:48361] [2022-06-04 07:25:24.80 UTC] Bearer on ConnectionId {localAddress = LocalAddress "", remoteAddress = LocalAddress "state-node-mainnet/node.socket"} event: State: Dead
[db-sync-node.Mux:Notice:48361] [2022-06-04 07:25:24.80 UTC] Bearer on ConnectionId {localAddress = LocalAddress "", remoteAddress = LocalAddress "state-node-mainnet/node.socket"} event: Miniprotocol MiniProtocolNum 5 InitiatorDir terminated with exception libpq: failed (another command is already in progress)
[db-sync-node.Subscription:Error:48360] [2022-06-04 07:25:24.80 UTC] Identity Application Exception: LocalAddress "../cardano-node/state-node-mainnet/node.socket" libpq: failed (another command is already in progress)
[db-sync-node.ErrorPolicy:Error:6] [2022-06-04 07:25:24.80 UTC] IP LocalAddress "../cardano-node/state-node-mainnet/node.socket" ErrorPolicyUnhandledApplicationException libpq: failed (another command is already in progress)
[db-sync-node.Subscription:Warning:51] [2022-06-04 07:25:24.81 UTC] Identity Failed to start all required subscriptions
[db-sync-node.Subscription:Notice:51] [2022-06-04 07:25:25.81 UTC] Identity Restarting Subscription after 1.027120268s desired valency 1 current valency 0
[...]

PR #1153, opened 2022-06-14
Fix node restarts

kderme:
https://github.com/input-output-hk/cardano-db-sync/issues/1135


erikd:
We really should not mix release based changes (ie "Prepare tag 13.0.0") with code fixes.

kderme:
Indeed I should have marked this as draft, since there is another pr in line before this https://github.com/input-output-hk/cardano-db-sync/pull/1144

erikd:
This has been subsumed by #1155 .

Issue #1139, opened 2022-06-06
Run interesting queries in ci


kderme:
This will make sure that queries listed in doc/interesting-queries.md ont' bit rot.

profd2004:
Closing this as stale as part of a triage pass over the older open issues — running example queries in CI was never implemented.

Feel free to reopen if still wanted.

PR #1177, opened 2022-07-04
fix typo

cardanians:


doc/interesting-queries.md
```diff
@@ -509,11 +509,11 @@ select tx.id as tx_id, tx_out.value as tx_out_value, redeemer.units_mem, redeeme
 
 ### Get all mint scripts
 ```sql
-select redeemer.tx_id as tx_id, redeemer.units_mem, redeemer.units_steps, redeemer.fee as redeemer_fee, redeemer.purpose, ma_tx_mint.policy, ma_tx_mint.name, ma_tx_mint.quantity
+select redeemer.tx_id as tx_id, redeemer.unit_mem, redeemer.unit_steps, redeemer.fee as redeemer_fee, redeemer.purpose, ma_tx_mint.policy, ma_tx_mint.name, ma_tx_mint.quantity
```
kderme, line 512: Hm this still doesn't run, because `policy` and `name` have moved to the `multi_asset` table. I think this should turn into something like this 
```
select redeemer.tx_id as tx_id, redeemer.unit_mem, redeemer.unit_steps, redeemer.fee as redeemer_fee, redeemer.purpose, multi_asset.policy, multi_asset.name, ma_tx_mint.quantity
  from redeemer 
    join multi_asset on redeemer.script_hash = multi_asset.policy 
    join ma_tx_mint on ma_tx_mint.ident = multi_asset.id and redeemer.tx_id = ma_tx_mint.tx_id 
      where purpose = 'mint';
```

kderme:
Thanks for this fix. We should really run these queries in ci, so that they don't bit-rot https://github.com/input-output-hk/cardano-db-sync/issues/1139

doc/interesting-queries.md
cardanians: oh sorry for that, this second query is really only partial fix (as I fixed mainly unit(s) and step(s).

cardanians:
> Thanks for this fix. We should really run these queries in ci, so that they don't bit-rot #1139

I even think some of the queries aren't quite accurate.

When I have time I will try to look at it too and add some interesting ones, there are never enough of these samples.

doc/interesting-queries.md
cardanians: @kderme fixed this one too

kderme:
Squashed and merged in https://github.com/input-output-hk/cardano-db-sync/pull/1180

PR #1187, opened 2022-07-09
Update CODEOWNERS

tmcgilchrist:
Fixes devops team access

@input-output-hk/docs-access exists but needs write access to the repository.


erikd:
I have added write access for @input-output-hk/docs-access .

PR #1188, opened 2022-07-09
Add schema ERD

kderme:
Entity Relationship Diagram was created by pgAdmin.

I believe this is interesting to have. Apps like pgAdmin/dbeaver make it quite easy to create those and we can update them with every major release.


erikd (changes requested):
Shouldn't this be linked into say the top level README?

Would also be nice to document how this file is generated.

erikd:
Pulled into https://github.com/input-output-hk/cardano-db-sync/pull/1203 .

PR #1200, opened 2022-07-19
Add an error message when no ledger state files are found

erikd:


erikd:
This is not yet tested, and should not be merged until it is.

PR #1202, opened 2022-07-19
change mktemp parameter for BSD compat

reqlez:
mktemp has 2 available options for directories:
mktemp --directory  and  mktemp -d

First option does not work on FreeBSD, second option works on Linux and FreeBSD.


erikd:
The nice thing about `--recursive` and `--directory` is that they are self documenting in a way that `-r` and `-d` are not.

reqlez:
Output: FreeBSD

I get it, i'm just trying to make it more POSIX compliant.

Actually, if we truly wanted to make it posix compliant, we need to change the template name as well as per:
https://pubs.opengroup.org/onlinepubs/009695399/functions/mktemp.html

reqlez:
For example... this is an issue also, but at least it does not prevent the script from running:

This is the file that gets created when i run that mktemp command on FreeBSD ( with the -d instead of course )
/tmp/db-sync-snapshot-XXXXXXXXXX.RKAUqG00

erikd:
I am in favor or making it POSIX compliant just not at the cost of making it less self-documented.

I was thinking of something along the lines of:
```sh
case "$(uname)" in
  FreeBSD)
     recursive="-r"
     directory="-d"
     ;;
  *)
    recursive="--recursive"
    directory="--directory"
    ;;
  esac
```
and then using the variables instead of using the switches directly.

reqlez:
That could work. I didn't realize the importance of that self-documentation since i saw other places where rm -rf is used for example.

Let me see what else is broken, if anything, i'm pretty sure it's good tho. I was able to restore a snapshot into my bsd server and that went well.

erikd:
I have hacked something up, but do not have spare machine capacity to test it currently. Here it is:

```diff
diff --git a/scripts/postgresql-setup.sh b/scripts/postgresql-setup.sh
index 5d5d79c8..0bcc462e 100755
--- a/scripts/postgresql-setup.sh
+++ b/scripts/postgresql-setup.sh
@@ -8,6 +8,22 @@ IFS=$'\n\t'
 
 progname="$0"
 
+case "$(uname)" in
+  FreeBSD)
+     recursive="-r"
+     directory="-d"
+     ;;
+  Linux)
+    # Linux tools have long names for these options.
+    recursive="--recursive"
+    directory="--directory"
+    ;;
+  *)
+    # These shoud work on any POSIX system
+    recursive="-r"
+    directory="-d"
+  esac
+
 
 function die {
 	echo "$1"
@@ -134,11 +150,11 @@ function create_snapshot {
 	tgz_file=$1.tgz
 	dbfile=$1.sql
 	ledger_file=$2
-	tmp_dir=$(mktemp --directory -t db-sync-snapshot-XXXXXXXXXX)
+	tmp_dir=$(mktemp "${directory}" -t db-sync-snapshot-XXXXXXXXXX)
 	echo $"Working directory: ${tmp_dir}"
 	pg_dump --no-owner --schema=public "${PGDATABASE}" > "${tmp_dir}/$1.sql"
 	cp "$ledger_file" "$tmp_dir/$(basename "${ledger_file}")"
-	tar zcvf - --directory "${tmp_dir}" "${dbfile}" "$(basename "${ledger_file}")" \
+	tar zcvf - "${directory}" "${tmp_dir}" "${dbfile}" "$(basename "${ledger_file}")" \
 	| tee "${tgz_file}.tmp" | sha256sum | sed -e "s/-/${tgz_file}/" > "${tgz_file}.sha256sum"
 	mv "${tgz_file}.tmp" "${tgz_file}"
 	rm -rf "${tmp_dir}"
@@ -156,13 +172,13 @@ function restore_snapshot {
 	  echo "Ledger state directory ($2) is not empty. Please empty it and then retry."
 	  exit 1
 	  fi
-	tmp_dir=$(mktemp --directory -t db-sync-snapshot-XXXXXXXXXX)
-	tar -zxvf "$1" --directory "$tmp_dir"
+	tmp_dir=$(mktemp "${directory}" -t db-sync-snapshot-XXXXXXXXXX)
+	tar -zxvf "$1" "${directory}" "$tmp_dir"
 	db_file=$(find "$tmp_dir/" -iname "*.sql")
 	lstate_file=$(find "${tmp_dir}/" -iname "*.lstate")
 	mv "${lstate_file}" "$2"
 	psql --dbname="${PGDATABASE}" -f "${db_file}"
-	rm --recursive "${tmp_dir}"
+	rm "${recursive}" "${tmp_dir}"
 }
 
 function usage_exit {
```

reqlez:
Thanks! I will try this out tomorrow!

Also, if you ever interested in getting cardano-node / cardano-db-sync installed on BSD we made some ports here:
cardano-node is release, but cardano-db-sync is right now "WIP".
https://github.com/cardano-bsd-alliance/freebsd-ports-haskell/tree/cardano/net-p2p/cardano-node
https://github.com/cardano-bsd-alliance/freebsd-ports-haskell/tree/cardano/net-p2p/cardano-db-sync
https://university.psilobyte.io/en/how-to/cardano-freebsd-relay-setup

erikd:
Please feel free to take my suggestions and tweak it to make it work.

erikd:
I will test this as soon as I am able. May be as late as early next week (July 25/26).

reqlez:
Totally fine, maybe by that time i have fixed an issue with source filename not appearing inside the sha256sum file on BSD.

reqlez:
This is the best solution i can come up with, maybe you can find another one... 
I replaced with: | sha256sum | head -c 64 | sed -e "s/$/  ${tgz_file}/" > "${tgz_file}.sha256sum"

Since for whatever reason, in linux ... a "-" is added at end, but in BSD, nothing is added, so i cannot replace the "-" instead i'm just taking the first 64 characters ( sha256 ) and then adding the filename after that.

reqlez:
Sorry i don't know what the heck happened to this branch... i'll fix

erikd:
Closed in favor of https://github.com/input-output-hk/cardano-db-sync/pull/1210 .

PR #1205, opened 2022-07-22
un-hardcode arch + bsd cmd output mod

reqlez:
"prepare-snapshot" command output is always hardcoded to x86_64, un-hardcoded arch.

Due to the generic name of the "postgresql-setup.sh" script, we cannot install this to the OS in our FreeBSD port "as-is" hence we are renaming it to: cardano-db-sync-pgsql-setup in our port. Added a modification that detects "freebsd" and outputs correct  "--create-snapshot" command so that users are not confused. Outputs "normal" command for all other OSes. This logic can later be used on other OSs as well that might require changing the default snapshot creation instructions.

Tested build on FreeBSD 13.1/ARM ( aarch64 )

"prepare-snapshot" command output seems correct:
image

### DIFF
--- a/cardano-db-tool/src/Cardano/DbTool/PrepareSnapshot.hs
+++ b/cardano-db-tool/src/Cardano/DbTool/PrepareSnapshot.hs
@@ -18,6 +18,7 @@ import           Ouroboros.Network.Block hiding (blockHash)
 
 import           Paths_cardano_db_tool (version)
 
+import           System.Info (arch, os)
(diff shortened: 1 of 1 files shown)


erikd (changes requested):
Please squash these two commits down into one.

reqlez:
> Please squash these two commits down into one.

Done.

erikd:
Closed in favor of https://github.com/input-output-hk/cardano-db-sync/pull/1210 .

PR #1208, opened 2022-07-25
Release/13.0.x

kderme:

PR #1215, opened 2022-08-03
Update dependencies

erikd:


erikd:
I do not see how these dependencies fail here but pass in https://github.com/input-output-hk/cardano-db-sync/pull/1208 .

kderme:
> I do not see how these dependencies correctly fail here but pass (incorrectly) in https://github.com/input-output-hk/cardano-db-sync/pull/1208 .

I think it should pass here also. `b443f540133888442e5475c05890dc2eee13d43e` is now in a release branch.

kderme:
Oh this is node not ouroboros-network. Probably the node `release/1.35` branch was forced pushed, so this revision is no longer available and ci on our side started to fail.

PR #1223, opened 2022-08-09
Reduce use of foreign key constraints and depend more heavily on BlockNo

erikd:


erikd:
I have been running this for nearly two weeks, on top of snapshotting and restoring the database. Seems fine.

PR #1235, opened 2022-08-18
Improve the way state snapshots are taken

erikd:
@reqlez I have modified the `scripts/postgresql-setup.sh` but my mod is Linux only. Do you want to add a FreeBSD code path?

### DIFF
--- a/scripts/postgresql-setup.sh
+++ b/scripts/postgresql-setup.sh
@@ -14,13 +14,23 @@ case "$(uname)" in
     recursive="--recursive"
     directory="--directory"
     force="--force"
+    test="--test"
     ;;
   *)
     # These shoud work on any POSIX system
     recursive="-r"
     directory="-d"
     force="-f"
+    test="-t"
   esac
+
+# This should work on all Linux variants as well as FreeBSD.
+numcores=$(getconf _NPROCESSORS_ONLN)
+if test "${numcores}" -le 2 ; then
+  numcores=1
+else
+  numcores=$(( numcores - 1 ))
+  fi
 
 function die {
 	echo "$1"
@@ -145,18 +155,20 @@ function dump_schema {
 
 function create_snapshot {
 	tgz_file=$1.tgz
-	dbfile=$1.sql
 	ledger_file=$2
 	tmp_dir=$(mktemp "${directory}" -t db-sync-snapshot-XXXXXXXXXX)
 	echo $"Working directory: ${tmp_dir}"
-	pg_dump --no-owner --schema=public "${PGDATABASE}" > "${tmp_dir}/$1.sql"
-	cp "$ledger_file" "$tmp_dir/$(basename "${ledger_file}")"
-	tar zcvf - --directory "${tmp_dir}" "${dbfile}" "$(basename "${ledger_file}")" | tee "${tgz_file}.tmp" \
-	| sha256sum | head -c 64 | sed -e "s/$/  ${tgz_file}\n/" > "${tgz_file}.sha256sum"
+	pg_dump --no-owner --schema=public --jobs="${numcores}" "${PGDATABASE}" --format=directory --file="${tmp_dir}/db/"
+	lstate_gz_file=$(basename "${ledger_file}").gz
+	gzip --to-stdout "${ledger_file}" > "${tmp_dir}/$(basename "${ledger_file}").gz"
+	tree "${tmp_dir}"
+	# Use plain tar here because the database dump files and the ledger state file are already gzipped.
+	tar cvf - --directory "${tmp_dir}" "db" "${lstate_gz_file}" | tee "${tgz_file}.tmp" \
+		| sha256sum | head -c 64 | sed -e "s/$/  ${tgz_file}\n/" > "${tgz_file}.sha256sum"
 	mv "${tgz_file}.tmp" "${tgz_file}"
 	rm "${recursive}" "${force}" "${tmp_dir}"
-	if test "$(gzip --test "${tgz_file}")" ; then
-	  echo "Gzip reports the snapshot file as being corrupt."
+	if test "$(tar "${test}" --file "${tgz_file}")" ; then
+	  echo "Tar reports the snapshot file as being corrupt."
 	  echo "It is not safe to drop the database and restore using this file."
 	  exit 1
 	  fi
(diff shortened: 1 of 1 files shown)


reqlez:
> @reqlez I have modified the `scripts/postgresql-setup.sh` but my mod is Linux only. Do you want to add a FreeBSD code path?

Thanks for the heads up. Let me test this out :)

scripts/postgresql-setup.sh
```diff
@@ -171,10 +178,19 @@ function restore_snapshot {
...
-	psql --dbname="${PGDATABASE}" -f "${db_file}"
+	if test -d "${tmp_dir}/db/" ; then
+	  # New pg_dump format
+      lstate_file=$(find "${tmp_dir}/" -iname "*.lstate")
+	  mv "${lstate_file}" "$2"
+	  drop_db
+	  pg_restore --jobs="${numcores}" --format=directory --dbname=cexplorer "${tmp_dir}/db/"
+	else
+	  # Old pg_dump format
+	  db_file=$(find "$tmp_dir/" -iname "*.sql")
```
reqlez, line 189: Is this a typo ?

scripts/postgresql-setup.sh
```diff
@@ -145,13 +153,12 @@ function dump_schema {
...
 	tgz_file=$1.tgz
-	dbfile=$1.sql
 	ledger_file=$2
 	tmp_dir=$(mktemp "${directory}" -t db-sync-snapshot-XXXXXXXXXX)
 	echo $"Working directory: ${tmp_dir}"
-	pg_dump --no-owner --schema=public "${PGDATABASE}" > "${tmp_dir}/$1.sql"
+	pg_dump --no-owner --schema=public --jobs="${numcores}" "${PGDATABASE}" --format=directory --file="${tmp_dir}/db/"
 	cp "$ledger_file" "$tmp_dir/$(basename "${ledger_file}")"
-	tar zcvf - --directory "${tmp_dir}" "${dbfile}" "$(basename "${ledger_file}")" | tee "${tgz_file}.tmp" \
+	tar zcvf - --directory "${tmp_dir}" "db" "$(basename "${ledger_file}")" | tee "${tgz_file}.tmp" \
```
reqlez, line 161: "--format=directory" is compressed by default no? does it make sense to re-compress ? Another way is to remove "z" from here. Only thing that is not compressed is the ledger state in this case ( hopefully eliminated one day as per your suggestion), but it also could be compressed separately. Just an idea, could save a bunch of extra CPU time potentially.

scripts/postgresql-setup.sh
```diff
@@ -22,6 +22,14 @@ case "$(uname)" in
     force="-f"
   esac
 
+# Only Linux is officially supported so this should be fine.
+numcores=$(grep -c processor /proc/cpuinfo)
```
reqlez, line 26: Another way of doing this that works everywhere ( I think... tried ubuntu, rocky linux and freebsd at least ... ) is: getconf _NPROCESSORS_ONLN

reqlez (commented):
Some things i found for now, I will get back to it tomorrow.

scripts/postgresql-setup.sh
reqlez: Actually, another benefit, is that gzip is single core process, unless you use pigz for example, so while the parallelism introduced by new pg_dump is great, now you are getting slowed down by the single-threaded compression.

scripts/postgresql-setup.sh
erikd: In the comment? Yes.

Changed to 
```
# Old snapshot format produced by this script
```

scripts/postgresql-setup.sh
erikd: I will play with this some more.

erikd: We probably don't want to gzip the dababase files (because they are already gzipped) but probably do want to gzip the ledger state file.

scripts/postgresql-setup.sh
reqlez: I was more curious about db_file=$(find "$tmp_dir/" -iname "*.sql")  VS.  lstate_file=$(find "${tmp_dir}/" -iname "*.lstate")

the missing {}

scripts/postgresql-setup.sh
```diff
@@ -145,13 +153,14 @@ function dump_schema {
...
 	tmp_dir=$(mktemp "${directory}" -t db-sync-snapshot-XXXXXXXXXX)
 	echo $"Working directory: ${tmp_dir}"
-	pg_dump --no-owner --schema=public "${PGDATABASE}" > "${tmp_dir}/$1.sql"
-	cp "$ledger_file" "$tmp_dir/$(basename "${ledger_file}")"
-	tar zcvf - --directory "${tmp_dir}" "${dbfile}" "$(basename "${ledger_file}")" | tee "${tgz_file}.tmp" \
+	pg_dump --no-owner --schema=public --jobs="${numcores}" "${PGDATABASE}" --format=directory --file="${tmp_dir}/db/"
+	lstate_gz_file=$(basename "${ledger_file}").gz
+	gzip --to-stdout "$ledger_file" > "$tmp_dir/$(basename "${ledger_file}").gz"
+	# Use plain tar here because the database dump files and the ledger state file are already gzipped.
+	tar cvf - --directory "${tmp_dir}" "db" "$(basename "${ledger_file}")" | tee "${tgz_file}.tmp" \
```
reqlez, line 163: Also should this not be ( to add the .gz )?:

tar cvf - --directory "${tmp_dir}" "db" "$(basename "${ledger_file}").gz" | tee "${tgz_file}.tmp" \

erikd: Still debugging this. Only pushed to the repo because I needed to pull it on another machine.

erikd:
Ok, I think I am finally happy wit this.

reqlez:
> Ok, I think I am finally happy with this. @reqlez ?

Nice! If you are not in a rush to merge this, let me take it for a spin and I will report back some time end of day.

scripts/postgresql-setup.sh
```diff
@@ -22,6 +22,14 @@ case "$(uname)" in
     force="-f"
   esac
 
+# This should work on all Linux variants as well as FreeBSD.
+numcores=$(getconf _NPROCESSORS_ONLN)
+if test "${numcores}" -le 2 ; then
+  numcores=1
+else
+  numcores=$(( numcores - 1 ))
+  fi
```
tmcgilchrist, line 31:
```suggestion
fi
```
Alignment on the closing `fi`

scripts/postgresql-setup.sh
```diff
@@ -145,18 +153,20 @@ function dump_schema {
...
+	gzip --to-stdout "${ledger_file}" > "${tmp_dir}/$(basename "${ledger_file}").gz"
+	tree "${tmp_dir}"
+	# Use plain tar here because the database dump files and the ledger state file are already gzipped.
+	tar cvf - --directory "${tmp_dir}" "db" "${lstate_gz_file}" | tee "${tgz_file}.tmp" \
+		| sha256sum | head -c 64 | sed -e "s/$/  ${tgz_file}\n/" > "${tgz_file}.sha256sum"
 	mv "${tgz_file}.tmp" "${tgz_file}"
 	rm "${recursive}" "${force}" "${tmp_dir}"
-	if test "$(gzip --test "${tgz_file}")" ; then
-	  echo "Gzip reports the snapshot file as being corrupt."
+	if test "$(tar --test --file "${tgz_file}")" ; then
```
tmcgilchrist, line 168: macOS tar does not have the `--test` flag and same for FreeBSD.

scripts/postgresql-setup.sh
erikd: But the current indention matches the rest of the file.

scripts/postgresql-setup.sh
erikd: :sob:

reqlez: Sorry I was dealing with Cardano FUD today instead of testing :(

reqlez:
So i did a test, and of course script failed at executing "tree" since it's not standard command. I guess I will have to add the "tree" package as requirement for the port. This might be a problem in macos as well, possibly ?

erikd:
Patches will definitely be considered.

reqlez:
> Patches will definitely be considered.
Cool :)  I did take a snapshot today using this new method here ( from 7 hours ago ) and for sure it is more efficient:
https://psb-dbsync.s3.ca-central-1.amazonaws.com/browser/index.html#13-aarch64/

However that db has the datumfix code, just heads up.

PR #1236, opened 2022-08-20
Pass various disable flags into nix service.

tmcgilchrist:


blaggacao:
I'm wondering, shouldn't we be all consolidating all packaging efforts via `cardano-world`? After all, those are the artifacts, that we ultimately consume...

tmcgilchrist:
I wasn't aware of [cardano-world](https://github.com/input-output-hk/cardano-world) until now, thanks. 
It might make sense to pull together the docker usage information into a single place including setting up `cardano-node` and `postgres`.

For now I am going to keep it in place here because it is easier for the team to maintain. Our usage of docker is split between the notes kept to be used by developers of `cardano-db-sync` and that used by consumers of `cardano-db-sync` running the docker images. That second group does need a more detailed set of documentation about running cardano using docker images. cc @kderme

blaggacao:
Absolutely! Cardano World is primarily meant for the second group, where SRE is being the driving force of integration.

But I've been meaning to explore local dev service manager targets such as docker-compose, foreman, hivemind et al.

The use of docker(-compose) in such scenarios is nice for interop, but it's also slow and probably most suited for ancillary third party services, such as a database, not necessarily the primary development target.

I've been looking at https://bob.build recently as a way to integrate build-steps seamlessly with that repo-local service management perspective, currently, they support docker compose, only.

However, while they use the Nix Package Collection under the hood, they don't currently expose an interface into the Nix Language (as configuration language).

doc/docker.md
```diff
@@ -66,7 +66,19 @@ To download and restore a snapshot include `RESTORE_SNAPSHOT`:
...
+## Disable options
 
+Consult the configuration [docs](docs/configuration.md) for what these options mean, assuming you have read that
+they can be accessed via env variables passed to docker-compose. Leave out any that do not make sense, eg if
+you just want to disable the ledger use `DISABLE_LEDGER=--disable-ledger docker-compose up`.
+
+``` shell
+DISABLE_LEDGER=--disable-ledger \
+DISABLE_CACHE=--disable-cache \
+DISABLE_LEDGER=--disable-ledger \
```
kderme, line 80:
```suggestion
DISABLE_EPOCH=--disable-epoch \
```

kderme:
Other than the typo is this ready to go?

jbgi (dismissed):

kderme (dismissed):

PR #1252, opened 2022-09-08
Add file doc/what_why_how.md

erikd:


doc/what_why_how.md
```diff
@@ -0,0 +1,29 @@
...
+### Why
+
+The database provided and maintained by `cardano-db-sync` allows anybody with some SQL expertise
+to extract information about the Cardano chain (including some data that is not on the chain but
+rather is part of ledger state). This information includes things like
+
+* The transactions in a specific block.
+* The balance of a specific address.
+* The staking rewards earned by a pool or address for a specific epoch.
+* ...
```
angerman, line 20: can we annotate (ledger state / chain), where these information come from?

angerman (commented):
Can we add a tiny note on how a blockchain and the database relate to event sourcing and state hydration? Does the database provide a view on the /current/ state? Or does it include historical information as well?

Issue #1190, opened 2022-07-11
Avoid rollbacks on restarts


kderme:
On restarts DBSync finds the latest point for which there is both a snapshot and its data are already inserted in the db. It deletes newer snapshots and data of newer blocks and starts syncing from this point. Deleting newer data can be avoided, if DBSync could instead apply the ledger rules from the snapshot until it reaches a compatible db tip. Before applying the rules, db-sync should first find the db tip and make sure it's on the node chain.

Other ways to avoid big rollbacks on restarts is to store more frequent snapshots or to store snapshots on signal handlers while exiting.

erikd:
> Other ways to avoid big rollbacks on restarts is to store more frequent snapshots

Yes, it will reduce the need for rollbacks, but done naively it would slow down sync speeds even further. In epoch `319` I see:
```
[2022-07-21 05:03:43.35 UTC] Took a ledger snapshot at ledger-state/mainnet/52823074-07f01c5bd4.lstate. It took 127.062317205s.
```
That is over 6 minutes for a singe operation of storing the ledger state on disk which it us currently doing about 3 times per epoch. More frequent snapshots, means more time taken storing them on disk. A partial solution to this is to off load it to another thread, but since `db-sync` is already I/O bound this will almost certainly do very little to help.

The correct solution is for `db-sync` to drop its maintenance of ledger state (which I believe was always the idea) and have `ledger`/`consensus`/`node` trickle feed ledger state data to `db-sync` in a fully deterministic way. This means that `db-sync` would not have any ledger state that need to be stored on disk, and would not need to rollback more than a couple of blocks (worst case) on restart.

erikd:
> to store snapshots on signal handlers while exiting.

There are a bunch of real world scenarios where signal handlers and on-exit functions do not get called. For example the machine looses power or the process is killed by the kernel OOM handler. I am sure there are many others.

kderme:
> There are a bunch of real world scenarios where signal handlers and on-exit functions do not get called. For example the machine looses power or the process is killed by the kernel OOM handler. I am sure there are many others.

Sure, we're trying to catch the most common case with a normal exit an improve it.

> The correct solution is for db-sync to drop its maintenance of ledger state (which I believe was always the idea) and have ledger/consensus/node trickle feed the required ledger state data to db-sync in a fully deterministic way. 

That something we should push to the node teams, but can take quite a while to design and implement. In the meantime we can try to improve the situation on our side.

[...]

PR #1266, opened 2022-09-18
Avoid Rollbacks during restart

kderme:
Fixes https://github.com/input-output-hk/cardano-db-sync/issues/1190
This should drastically speed up restart times.
Still draft as it needs a lot of testing.

Currently when there is a restart, db-sync goes back to the latest on disk ledger snapshot. It also deletes all data in the db back to this point as it maintains the property that ledger and db are always at the same point. However in most cases deleting the db is unnecessary and db-sync has to reinsert all the deleted data, which may take long if many blocks are deleted.

With this pr we change the property: it is now that the db tip point is never behind the ledger. So on a relatively long rollback received from the chainsync server, db-sync won't delete any db-blocks. It will apply each new block to the ledger state and it will check if the block is already in the db. If it's already there it just goes to the next block. If it's not, it first cleans up the db for any existing blocks with greater or equal BlockNo and then inserts the block. After that the initial property, ledger and db at the same point, is restored.

As an optimization, when db-sync doesn't do a full rollback, ie doesn't delete the db, it will change its `ConsistentLevel` to `DBAheadOfLedger`. When the consistency between ledger and db is restored it will change the level back to `Consistent`. The query I mentioned above only happens with `DBAheadOfLedger`, so in the normal syncing case no additional queries are executed.


erikd:
How does it avoid rollbacks?

kderme:
> How does it avoid rollbacks?

I added a description. In short db-sync will not delete a block from the db until the time that itself recognises that there is a fork, ie different block with the same BlockNo and not when the node sends a Rollback message.

erikd:
Would be really nice to delete the description above and put in in the commit message so we have a permanent record of it in the Git repo rather than here on Github

> .It will apply each new block to the ledger state and it will check if the block is already in the db. If it's already there it just goes to the next block.

I assume it also needs to act on all the ledger events. Is that what it does? If it gets a block that is already in the db, but the ledger state emits and event containing say rewards, wouldn't those rewards already be in the db? Does it check to see if they are already there and avoid inserting them again?

kderme:
If the block hash is there it assumes ledger events have also been inserted, so it does nothing else with the block and ignores the events.

erikd:
ok, would like to see a notes to this effect added to the commit message. The commit message can be a lot briefer than what you have above.

Issue #1274, opened 2022-09-29
Remaining Tech debt


kderme:
- [ ] So far we've tried to split imports based on the package they come from. This has caused the import list to be chaotic, it makes it hard to maintain, hard to find import etc. We should list all imports alphabetically without any empty lines 
- [ ] Queries in `Cardano.Db.Query` have many queries that are used from db-sync, many that are used in tests and many that are never used. We should delete what is never used db-sync and split in different module what is not used by db-sync. This will make it more clear what queries and indexes db-sync needs during syncing, so we can delete the others. Related to https://github.com/input-output-hk/cardano-db-sync/issues/1087
- [ ] We should rename many modules, which include the name DB/Database to Storage as they're also related to ledger state not just the DB. Examples are `Cardano.DbSync.Database`, `Cardano.DbSync.DbAction` 
- [ ] Module `Cardano.DbSync.Default` is very generic and a left over from the plugins. Other plugins names should also be removed.
- [ ] We should do more splitting between the chainsync client thread and the db thread, using the `ActionQueue` more.
- [ ] We should use a uniform format everywhere.

profd2004:
Closing this as stale as part of a triage pass over the older open issues — some listed items remain but are better re-filed individually.

Feel free to reopen or file fresh issues for anything still relevant.

PR #1275, opened 2022-09-29
Tech debt

kderme:
https://github.com/input-output-hk/cardano-db-sync/issues/1274

### DIFF
--- a/cardano-chain-gen/Setup.hs
+++ b/cardano-chain-gen/Setup.hs
@@ -1,18 +1,14 @@
 
 import           Cardano.Crypto.Hash
 import           Control.Monad (foldM, forM)
-
 import qualified Data.ByteString.Char8 as BS
-
 import qualified Data.List as List
-
 import           Distribution.PackageDescription (extraSrcFiles)
 import           Distribution.Simple (UserHooks (..), defaultMainWithHooks, simpleUserHooks)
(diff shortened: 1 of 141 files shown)


erikd:
> * Remove unused queries

These queries you a deleting were designed to be examples. I can't see the point of removing compiler checked documentation.

> *Use a single import group instead of splitting them with empty lines

This was done for readability.

> * Annotate queries based on their usage

I would actually call this document or comment rather than "annotate" but sure.

kderme:
> > * Remove unused queries
> 
> These queries you a deleting were designed to be examples. I can't see the point of removing compiler checked documentation.

I think they are mostly leftovers and identical to other queries, so not sure how much they offer in terms of documentation. Also devs can always resurect them from git history.

If there is some specific query that you believe is worth keeping, of course we can.

> > *Use a single import group instead of splitting them with empty lines
> 
> This was done for readability.

Yes, but I think since db-sync has a big amount of dependencies, this is getting really hard to maintain.

erikd:
> Yes, but I think since db-sync has a big amount of dependencies, this is getting really hard to maintain.

A lot of this is due to `ledger`, `consensus` and `network` actually lacking usable APIs.

kderme:
Superseded by https://github.com/input-output-hk/cardano-db-sync/pull/1334

Issue #1087, opened 2022-03-22
Remove unused Indexes, Unique and Primary Keys


kderme:
In the same spirit as https://github.com/input-output-hk/cardano-db-sync/issues/1082 we could continue simplifying the schema trying to optimize insertions.

- Unique and autoincremental Primary Id keys on tables like `StakeRegistration`, `Delegation`, `TxIn`, `MaTxOut`, `MaTxMint` which are not referenced by any other table could probably be deleted. 
- Tables like `Reward` and `EpochStake` need to maintain their Unique keys, since we can have duplicate insertions on rollbacks. However their Primary Id is never used.
- Tables like ~`Tx`~, `Redeemer` which are referenced by other tables need to maintain their Primary key, but their Unique key can be removed. We already know that their hashes are unique and the specs make sure there will never be any duplicates. We already use `insertUnchecked` on them (or should use), so if there is a violation the whole thing crashes anyway. There is no query on their hashes, but we should audit if they are used in other queries while inserting (removing unique keys would make these slower).
- Tables like `StakeAddress`, `MultiAsset`, `Script` need to maintain both primary and unique keys. The difference with previous tables is that they can be inserted in different places and db-sync may try to insert them multiple times. We also need a way to get their Primary id from their hash.

We could also try to audit indexes that are important for syncing and disable all others. Applications that build on top of db-sync could insert their own indexes based on their needs.

The above list probably needs a better audit to check if it's correct for all tables and the mentioned tables can probably be extended.

erikd:
I am testing something right now which just removes the foreign key constraints. Currently on mainnet epoch `289` . While there did seem to be a small speed improvement in the Byron era, I want to see it sync the full chain before I pass judgement.

erikd:
Removing foreign key constraints (and nothing else) provides significant sync speed improvements when testing on mainnet. For Byron era epochs, the speed improvement is about 25%. For the latest era (Alonzo) the sync speed improvement are from 50% to 70%. I synced mainnet from genesis to epoch `308` in about 52 hours (still running). I would be a little surprised if the complete sync would take more than 72 hours (3 days).

erikd:
@rdirt There is still a lot of work to be done on this idea.

rdlrt:
> @rdirt There is still a lot of work to be done on this idea.

[...]

PR #1282, opened 2022-10-19
Loosen unique constraint on `extra_key_witness`

FinleyMcIlwaine:
The previous unique constraint made `extra_key_witness` only hold key witnesses for the first time a key was used as a "required signer" ("extra key witness"). Subsequent uses of keys as required signers would fail to insert due to the constraint. This loosens the constraint so the combination of the `hash` and `tx_id` columns must be unique. Since this new constraint is unlikely to trigger, I changed the insertion of key witnesses to use `insertUnchecked`. Now any transaction which has "required signers" will have those included in the `extra_key_witness` table.

### DIFF
--- a/cardano-db/src/Cardano/Db/Insert.hs
+++ b/cardano-db/src/Cardano/Db/Insert.hs
@@ -133,8 +133,8 @@ insertEpochParam = insertCheckUnique "EpochParam"
 insertEpochSyncTime :: (MonadBaseControl IO m, MonadIO m) => EpochSyncTime -> ReaderT SqlBackend m EpochSyncTimeId
 insertEpochSyncTime = insertReplace "EpochSyncTime"
 
-insertExtraKeyWitness :: (MonadBaseControl IO m, MonadIO m) => ExtraKeyWitness -> ReaderT SqlBackend m ExtraKeyWitnessId
-insertExtraKeyWitness = insertCheckUnique "ExtraKeyWitness"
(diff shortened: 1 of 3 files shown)


cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -134,7 +134,7 @@ insertEpochSyncTime :: (MonadBaseControl IO m, MonadIO m) => EpochSyncTime -> Re
 insertEpochSyncTime = insertReplace "EpochSyncTime"
 
 insertExtraKeyWitness :: (MonadBaseControl IO m, MonadIO m) => ExtraKeyWitness -> ReaderT SqlBackend m ExtraKeyWitnessId
-insertExtraKeyWitness = insertCheckUnique "ExtraKeyWitness"
+insertExtraKeyWitness = insertUnchecked "ExtraKeyWitness"
```
erikd, line 137: If we remove the `txId` from the `UniqueWitness` constraint, it is probably unnecessary to use `insertUnchecked`.

FinleyMcIlwaine:
That makes sense, thanks. I've updated this PR to remove the unique constraint and have `insertExtraKeyWitness = insert` now. I'm not totally sure that's how you'd want the insert to be done with the constraint removed. I also understand if you'd like to close this and batch it in with your #1087 changes.

erikd:
Closing in favor of https://github.com/input-output-hk/cardano-db-sync/pull/1283 so it runs through CI.

PR #1283, opened 2022-10-19
Remove unique constraint on `extra_key_witness`

erikd:


erikd:
Going to ignore the Cicero CI error.

PR #1329, opened 2022-12-29
Add tx_in indexes for reference_tx_in and collateral_tx_in

sorki:
Identified as having significant performance impact in https://github.com/blockfrost/blockfrost-backend-ryo/issues/42

Should I create a PR for `release/13.1.0.x` as well?


kderme:
This was closed by a bot for reasons..
Anyway, I think it shouldn't be a responsibility of db-sync to create all possible indexes, since it doesn't know which ones each user/client may need. Also indexes add delays during syncing.
In the coming release `13.1.0.0`, users can add their indexes in stage 3 or stage 4 migration which is documented in https://github.com/input-output-hk/cardano-db-sync/blob/release/13.1.0.x/doc/migrations.md#upgrading-to-13100

PR #1334, opened 2023-01-11
Add fourmolu code formatter

kderme:
Merge https://github.com/input-output-hk/cardano-db-sync/pull/1332 first


scripts/fourmolize.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+fourmolu --indentation 2 --indent-wheres true --haddock-style single-line \
```
jbgi, line 5: would a `fourmolu.yaml` config be clearer? would also permit to use fourmolu safely outside of this script.
Also cardano-ledger adopted fourmolu recently: https://github.com/input-output-hk/cardano-ledger/blob/master/fourmolu.yaml
so I'm wondering if a uniform fourmolu config across project would be easier for people working on multiple project?

kderme: The config file definitely makes sence.
The only difference I see with leger is the `record-brace-space`, for which I use the default `false`
I'm ok with this diff, so I'll adopt it
```
-        throwIO err404{errBody = encode err}
+        throwIO err404 {errBody = encode err}
```
```
-  toJSON failure@DbLookupPoolMetadataHash{} =
+  toJSON failure@DbLookupPoolMetadataHash {} =
```

olgahryniuk (approved):

PR #1335, opened 2023-01-12
Removed `./` from `customConfig` input `url`

aciceri:
Having a flake input of type `path` starting with `./` was causing errors like this:

```
error: cannot fetch input 'path:./custom-config?lastModified=1&narHash=sha256-Zd5w1I1Dwt783Q4WuBuCpedcwG1DrIgQGqabyF87prM=' because it uses a relative path
```

I was simply trying to build the docker image using this repo as a flake (so without `nix-build` as indicated in the docs) with `nix build .#dockerImage` when I got this error.

Can't tell if this appeared only with more recent Nix versions (I'm on `2.12.0`), by the way this seems to fix it.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1111,11 +1111,11 @@
       "locked": {
         "lastModified": 1,
         "narHash": "sha256-Zd5w1I1Dwt783Q4WuBuCpedcwG1DrIgQGqabyF87prM=",
-        "path": "./custom-config",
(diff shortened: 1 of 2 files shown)


jbgi (approved):

aciceri:
@jbgi I think this is not needed anymore (?)
Now `customConfig` is empty by default:
```nix

PR #1336, opened 2023-01-12
Fixup various nix and CHaP issues

michaelpj:
Still building this, but it looks promising.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -157,3 +157,11 @@ source-repository-package
   location: https://github.com/input-output-hk/cardano-config
   tag: e9de7a2cf70796f6ff26eac9f9540184ded0e4e6
   --sha256: 1wm1c99r5zvz22pdl8nhkp13falvqmj8dgkm8fxskwa9ydqz01ld
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/iohk-monitoring-framework
+  tag: 066f7002aac5a0efc20e49643fea45454f226caa
+  --sha256: 0s6x4in11k5ba7nl7la896g28sznf9185xlqg9c604jqz58vj9nj
+  subdir:
+    plugins/scribe-systemd
--- a/cabal.project
+++ b/cabal.project
@@ -122,26 +122,6 @@ allow-newer:
 
 source-repository-package
   type: git
-  location: https://github.com/input-output-hk/cardano-node
-  tag: ebc7be471b30e5931b35f9bbc236d21c375b91bb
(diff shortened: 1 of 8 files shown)


flake.nix
```diff
@@ -39,6 +39,8 @@
       supportedSystems = import ./supported-systems.nix;
       defaultSystem = head supportedSystems;
 
+      inputMap = { "https://input-output-hk.github.io/cardano-haskell-packages" = CHaP; };
```
angerman, line 42: Maybe we should be more explicit how this works?

We want to ensure that we do get a pinned version of CHaP (as we want to minimise the risk of someone
impersonating GitHub, and subtly giving us some modified CHaP). Thus we use `nix` to lock the CHaP reference,
and then need to pass that locked reference to haskell.nix

1. We reference iog/cardano-haskell-packags in the flake and give it the name `CHaP`.
2. We need to inform haskell.nix about this custom hackage. For this we tell haskell.nix to map the `repository` to the location of `CHaP` from the flake. We record this mapping in the `inputMap`
3. Finally we need pass that `inputMap` to haskell.nix, so it knows where to find the custom hackage.

angerman (approved):

cabal.project
```diff
@@ -120,26 +92,6 @@ allow-newer:
 -- --------------------------- 8< --------------------------
 -- Please do not put any `source-repository-package` clause above this line.
 
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/cardano-node
```
kderme, line 125: This changes the pinned cardano-node version to 1.35.3 though right? My idention was to use node 1.35.4 and the same version dependencies as node 1.35.4. I think this cannot be achived simply by the index-state date, can it?

michaelpj: You can just use normal version bounds to get that, I think.

michaelpj:
@kderme feel free to take what you like from this, e.g. you can revert the one removing the node s-r-p if you want to do it that way.

kderme:
Many thanks. Though I get this very strange error

michaelpj:
What command are you running?

kderme:
Any nix-build and nix-shell command like
`nix-build -A cardano-db-sync -o db-sync-node`

michaelpj:
Yikes. You appear to have two parallel setups in `flake.nix` and `default.nix`, that's very surprising. I'll fix up `default.nix` also.

michaelpj:
Done, I think.

kderme:
I cherry-picked most of this here https://github.com/input-output-hk/cardano-db-sync/pull/1332. Thanks!

PR #1353, opened 2023-02-08
Fix queryPoolFetchRetry

erikd:


schema/migration-2-0026-20230208.sql
```diff
@@ -0,0 +1,19 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 26 THEN
+    EXECUTE 'ALTER TABLE "reverse_index" ALTER COLUMN "min_ids" SET NOT NULL' ;
```
kderme, line 9: This migration seems irrelevant with this pr. Strange that it wasn't already in place as there is no `Maybe` annotation in the persistent TH. Could we skip the migration for now?

kderme (dismissed):
LGTM, modulo the migration comment.
Personal note that this doesn't result into any sql schema changes, even though the TH changes.

schema/migration-2-0026-20230208.sql
erikd: I always tried to make sure the migration files were always up-to-date.

The problem is that if we do not do this now, this will become part of the next schema change even though its comepletely unrelated.

kderme: I actually think this shouldn't be a schema change, but a change to the TH and Haskell types.

erikd: This is not a change.

Its fixing an inconsistency between the Haskell schema definition and the migration files.

I think there should be a CI test added to prevent this ever happening again.

kderme: It definitely changes the schema and this shouldn't be done if there is not an important reason, as the pr template suggests [Migrations](https://github.com/input-output-hk/cardano-db-sync/blob/master/.github/PULL_REQUEST_TEMPLATE.md#migrations)

For example for the next release we don't have any schema changes in place. If it ends up being a schema change bump just for this change, then new snapshots need to be create (ie 13.2) which won't be compatible with 13.1 as migrations are not reversible.

kderme: > I think there should be a CI test added to prevent this ever happening again.

I agree that a test like this would be great, to avoid similar issues in the future.

erikd: > It definitely changes the schema and this should be done only if there is an important reason, as the pr template suggests

The code that is shipped now, means that the Haskell code assumes that this column is never a `Maybe` whereas the SQL schema thinks that column is nullable. I think that is dangerous.

kderme: Yes in general such mismatches are dangerous. But since only db-sync writes to disk, there will never be a case where a null field is inserted, which could cause issues.

I created a ticket for this https://github.com/input-output-hk/cardano-db-sync/issues/1358 and added it to the next schema changing planned release (13.2.0.0) so we don't forget it https://github.com/orgs/input-output-hk/projects/52/views/6

erikd:
Hydra is dead!

PR #1370, opened 2023-03-01
Fix rollbacks for epoch table

erikd:


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -220,6 +221,21 @@ queryBlockNo blkNo = do
...
+    blk <- from $ table @Block
+    where_ (blk ^. BlockBlockNo ==. just (val blkNo))
+    pure (blk ^. BlockId, blk ^. BlockEpochNo)
+  pure $ convert (listToMaybe res)
+  where
+  convert :: Maybe (Value (Key Block), Value (Maybe Word64)) -> Maybe (BlockId, Word64)
+  convert mr =
+    case mr of
+      Nothing -> Nothing
+      Just (_, Value Nothing) -> Nothing -- Should not ever happen.
```
kderme, line 236: Personal note this can't happen because it's only null for genesis which doesn't have a blockNo in the first place.

kderme (approved):
LGTM! I think this fixes a big part of the problem

erikd:
I intend to check the that this fixes if when the epoch rolls over in about an hour.

erikd:
Seemed to work over the most recent epoch transition. Will merge this and keep an eye on it.

Issue #1259, opened 2022-09-13
--disable-ledger should not be compatible or need --state-dir


kderme:
When `--disable-ledger` is enabled, db-sync still requires the `--state-dir`, even though it's not used anywhere and it has no meaning, as there are not ledger snapshots. We should improved the cli parsers here.

PR #1376, opened 2023-03-13
1259 - Make ledger state dir optional to handle disable ledger option

Cmdv:
# Description

This PR fixes issue #1259

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -137,43 +137,46 @@ runSyncNode ::
...
+              ranAll
+              (enpForceIndexes enp)
+              runMigration
+
+        -- If the DB is empty it will be inserted, otherwise it will be validated (to make
+        -- sure we are on the right chain).
+        lift $ Db.runIohkLogging trce $ withPostgresqlConn dbConnString $ \backend -> do
+          liftIO $ unless (enpHasLedger enp) $ do
+            logInfo trce "Migrating to a no ledger schema"
+            Db.noLedgerMigrations backend trce
```
Cmdv, line 173: @kderme am I right in thinking this is all it does for `enpHasLedger` being false?
Just that the rest of the actions in `runSyncNode` function would still happen. I've stuck a `case` at the top but unsure if everything is needed again.

Cmdv:
Closing as superseded by https://github.com/input-output-hk/cardano-db-sync/pull/1378

PR #1378, opened 2023-03-15
Make ledger state dir optional to handle disable ledger option

Cmdv:
# Description

Alternate effort to solving #1259 

# Checklist

- [x] Self-reviewed the diff

# Migrations


kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -187,9 +184,6 @@ generateNewEpochEvents env details = do
         , esEpochNo = Strict.Just currentEpochNo
         }
 
-getTrace :: SyncEnv -> Trace IO Text
```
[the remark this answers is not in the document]
Cmdv: yes but unfortunately it won't type check because `envLedger` is a `Maybe` and it seems that `getTrace` isn't solely used in places that don't required and `LedgerEnv` 😞

cardano-db-sync/src/Cardano/DbSync/Database.hs
```diff
@@ -40,24 +40,25 @@ data NextState
 
 runDbThread ::
   SyncEnv ->
+  LedgerEnv ->
   MetricSetters ->
   DbActionQueue ->
   IO ()
-runDbThread env metricsSetters queue = do
+runDbThread syncEnv ledgerEnv metricsSetters queue = do
```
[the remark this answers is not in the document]
Cmdv: But if it's a `Maybe` what do I make `leTrace` if it's `Nothing`?

I have a feeling that trace needs to be inside `SyncEnv` not `LedgerEnv`??

cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -137,43 +138,54 @@ runSyncNode ::
...
-      lift $ orDie renderSyncNodeError $ insertValidateGenesisDist trce backend (dncNetworkName enc) genCfg (useShelleyInit enc)
-      liftIO $ epochStartup (enpExtended enp) trce backend
-
-    case genCfg of
-      GenesisCardano {} -> do
-        liftIO $ runSyncNodeClient metricsSetters syncEnv iomgr trce (enpSocketPath enp)
+      lift $ orDie renderSyncNodeError $ insertValidateGenesisDist trce backend (dncNetworkName syncNodeConfig) genCfg (useShelleyInit syncNodeConfig)
+      liftIO $ epochStartup (enpExtended syncNodeParams) trce backend
+    -- TODO: Vince: a lot of the downstream functions require LedgerEnv but is this right?
+    whenJust (envLedger syncEnv) $ \ledgerEnv ->
```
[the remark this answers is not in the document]
Cmdv: I thought that was the case. But if there is no ledger directory or `--disable-leger` then `LedgerEnv` has to be `Nothing`. We either pull out the values that are used by both state (Has Ledger, Has no Ledger) which live inside `LedgerEnv` like `leTrace`. Because the only way you can get access to those is to take them out of `Maybe` but seems like they are used in places that don't require a `LedgerEnv`?

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -187,8 +197,23 @@ generateNewEpochEvents env details = do
         , esEpochNo = Strict.Just currentEpochNo
         }
 
+getTopLevelConfig :: SyncEnv -> TopLevelConfig CardanoBlock
```
Cmdv, line 200: I'm unsure about this one as it's used in places that seem like it should only work if there is a Ledger

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -187,8 +197,23 @@ generateNewEpochEvents env details = do
...
+getTopLevelConfig syncEnv =
+  case envLedgerEnv syncEnv of
+    HasLedger hasLedgerEnv -> Consensus.pInfoConfig $ leProtocolInfo hasLedgerEnv
+    NoLedger noLedgerEnv -> Consensus.pInfoConfig $ nleProtocolInfo noLedgerEnv
```
Cmdv, line 204: I'm unsure about this one as it's used in `runSyncNodeClient` when creating `codecConfig` so had to add it `NoLedgerEnv` 🤷

cardano-db-sync/src/Cardano/DbSync/LocalStateQuery.hs
```diff
@@ -52,12 +52,16 @@ import Ouroboros.Network.Protocol.LocalStateQuery.Client (
...
-  , nlsSystemStart :: !SystemStart
-  , nlsQueryVar :: StateQueryTMVar CardanoBlock CardanoInterpreter
-  , nlsHistoryInterpreterVar :: StrictTVar IO (Strict.Maybe CardanoInterpreter)
+data NoLedgerEnv = NoLedgerEnv
+  { nleTracer :: Trace IO Text
+  , nleSystemStart :: !SystemStart
+  , nleQueryVar :: StateQueryTMVar CardanoBlock CardanoInterpreter
+  , nleHistoryInterpreterVar :: StrictTVar IO (Strict.Maybe CardanoInterpreter)
+  , nleNetwork :: !Ledger.Network
+  , nleProtocolInfo :: !(Consensus.ProtocolInfo IO CardanoBlock)
```
Cmdv, line 64: The addition of these two values, I hope is ok? was trying to make it match the parts that are shared in `HasLedgerEnv`

cardano-db-sync/src/Cardano/DbSync/Database.hs
```diff
@@ -113,32 +113,35 @@ runActions env actions = do
...
+                  [ "Ledger "
+                  , textShow statePoint
+                  , " and ChainSync "
+                  , textShow point
+                  , " don't match."
+                  ]
+            pure Nothing
+          Left lsfs ->
+            Just . fmap fst <$> verifySnapshotPoint syncEnv (OnDisk <$> lsfs)
+    NoLedger _ -> pure Nothing
```
Cmdv, line 136: Not sure what the scenario should be here of is what I've done is in fact correct!

Cmdv, line 136: Not sure what the scenario should be here of is what I've done is in fact correctby returning `NoLedger _ -> pure Nothing`?

kderme: I think that's fine. And since the separate cases are handled in this function, the caller place can be siplified, by removing the check
```
            if hasLedgerState env
              then lift $ rollbackLedger env chainSyncPoint
              else pure Nothing
```

doc/configuration.md
```diff
@@ -30,7 +30,7 @@ When this flag is enabled, some features are missing and some DB tables are left
 
 Warning: Running db-sync with this flag and then restarting it without the flag will cause crashes and should be avoided.
 
-Known issues: When this flag in enabled, the `--state-dir` option still needs to be provided, even though db-sync will completely ignore it.
+Warning: It was previously required to still have a `--state-dir` option provided when in conjunction with `--disable-ledger`. This is no longer the case and now an error will occure if they are both present at the same time.
```
kderme, line 33: 👍

kderme (changes requested):

kderme (commented):
Just a few comments added. Looks great otherwise!

PR #1380, opened 2023-03-22
Improve epoch calculations by introducing cache

### LINKED ISSUE #708: Improve epoch table calculation
The `Epoch` table contains a row per epoch for like the following:

```sql
select * from epoch where no = 250 ;
 id  |      out_sum      |    fees     | tx_count | blk_count | no  |     start_time      |      end_time       
-----+-------------------+-------------+----------+-----------+-----+---------------------+---------------------
 251 | 48999055712564141 | 41831556274 |   192598 |     20991 | 250 | 2021-02-24 21:44:51 | 2021-03-01 21:44:44
(1 row)

```

When `db-sync` is syncing the row is updated once per epoch via an SQL query that aggregates the data for the epoch and then writes a row to the database.

When `db-sync` is following the chain tip, the row for the current epoch is updated after every block, also via an SQL query.

The syncing query (once per epoch) is of zero concern. However, when following the chain tip, running this query for every single new block is suboptimal.

When following the chain we would like to avoid doing an SQL query per block. It should be possible to grab the info out of each block to update a cache and the cached value is written to the database. At the end of an epoch the SQL query can be run again to validate that the epoch row from the query matches the value accumulated during the epoch.

**Probably need to get real numbers on how long it currently takes to get the epoch table entry via an SQL query,**

Cmdv:
# Description

This fixes #708 which introduces a Cache to minimise the expensive DB lookup / Calculations used when working out new Epoch which happens as a reaction to new Blocks being added and in following mode.

To achieve this we're keeping Epoch and last know blockId in cache. Then on each new lookup instead of calculating from zero we can do a lookup of only the blocks we are missing and add the results to our epoch in cache. Thus resulting in a new Epoch that can then both be updated in cache and db with minimal overhead.
 

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -84,75 +96,129 @@ epochInsert trce (BlockDetails cblk details) = do
...
+    Nothing -> queryCalc
+    Just _ ->
+    -- Just cacheEp ->
+      case ceLastKnownBlock cachedEpoch of
+        Nothing -> queryCalc
+        Just _ ->
+        -- Just lastKnowBlockId ->
+          -- if we have both a block and epoch then lets use them
+          -- to do a new less expensive query of updating the epoch
+          queryCalc
```
Cmdv, line 164: Might need some pointers in regards to the DB query that can utilise the last known block in cache 🤔

Cmdv: from a top level I think the new query should:

1.  use last stored block in cache and retrieve all blocks that came after it (either by ID or time stamp)
2. sum the needed parameters of the blocks.
3. merger the result to the exiting epoch both in cache and DB.

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -46,6 +46,16 @@ rollbackFromBlockNo syncEnv blkNo = do
     lift $ do
       DB.deleteBlocksBlockId trce blockId
       DB.deleteEpochRows epochNo
+      -- clear any Epoch related state in cache
```
kderme, line 49: Possibly this should live in `rollbackCache`

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -95,6 +98,11 @@ data CacheStatistics = CacheStatistics
   , prevBlockQueries :: !Word64
   }
 
+data CacheEpoch = CacheEpoch
+  { ceEpoch :: !(Maybe DB.Epoch)
```
kderme, line 102: Is it ever the case where this is Nothing and BlockId is Just? If not then we could change the type to `Maybe (CacheEpoch DB.Epoch (Maybe DB.BlockId))

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -70,7 +82,7 @@ epochInsert trce (BlockDetails cblk details) = do
           -- the Ouroboros Classic era.
           pure $ Right ()
         Byron.ABOBBlock _blk ->
-          insertBlock trce details
+          insertEpoch trce cache details
```
Cmdv, line 85: we need to pass the _blk instead of looking it up from db downstream

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -84,75 +96,157 @@ epochInsert trce (BlockDetails cblk details) = do
...
-latestCachedEpochVar :: IORef (Maybe Word64)
-latestCachedEpochVar = unsafePerformIO $ newIORef Nothing -- Gets updated later.
+updateEpochNum ::
+  (MonadBaseControl IO m, MonadIO m) =>
+  Cache ->
+  Word64 ->
+  Trace IO Text ->
+  ReaderT SqlBackend m (Either SyncNodeError ())
+updateEpochNum cache slotEpochNum trce = do
+  mid <- queryForEpochId slotEpochNum
```
Cmdv, line 138: get this from the cache first and if it fails then get it from db

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -228,69 +232,51 @@ calculateEpochUsingCache cBlock cEpoch slotDetails = do
...
+      qResInputRes <- lift $ mapM queryResolveInput (Generic.txInputs tx)
+
       let cacheFee = fromIntegral $ DB.unDbLovelace $ DB.epochFees cEpoch
-          otS = fromInteger outS
-          inSum = sum $ map (DB.unDbLovelace . thrd3) resolvedInputs
-          diffSum = if inSum >= otS then inSum - otS else 0
-          newFees = maybe diffSum (fromIntegral . unCoin) (Generic.txFees t)
+          -- get all of the values that returned and sum them up
+          inSum = sum [DB.unDbLovelace dbl | Right (_, dbl) <- qResInputRes]
+          diffSum = if inSum >= outS then inSum - outS else 0
```
kderme, line 275: I don't think this calculation of fees is correct. The equation `inputs + fees = outputs` doesn't hold as there are also deposits, withdrawals etc

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -228,69 +232,51 @@ calculateEpochUsingCache cBlock cEpoch slotDetails = do
...
+    -- calculating the fees for current block using Tx's
+    calNewFees ::
+      (MonadBaseControl IO m, MonadIO m) =>
+      Generic.Tx ->
+      Word64 ->
+      ExceptT SyncNodeError (ReaderT SqlBackend m) Word64
+    calNewFees tx outS = do
+      -- taking all the txInputs and getting their values from a DB lookup.
+      -- Errors are ignored but might actually need to handle this gracefully
+      qResInputRes <- lift $ mapM queryResolveInput (Generic.txInputs tx)
```
kderme, line 270: This aproach goes against the incentive of reducing the db queries necessary. We already resolve the inputs in `insertTx` and calculate the tx fees. Isn't it possible to resuse these results?

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
kderme: Actually the next line fixes this. But still I think we can avoid replicating the logic by reusing the result in `insertTx`

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -144,6 +157,53 @@ getCacheStatistics cs =
...
+  Cache ->
+  CacheEpoch ->
+  ReaderT SqlBackend m ()
+writeCacheEpoch cache cacheEpoch =
+  case cache of
+    UninitiatedCache -> pure ()
+    Cache ci -> liftIO $ atomically $ writeTVar (cEpoch ci) $ Just cacheEpoch
+
+-- write a block into the cache
+writeBlockToCacheEpoch ::
```
kderme, line 177: I'm assuming this won't be used eventually

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -114,8 +117,13 @@ insertShelleyBlock syncEnv shouldLog withinTwoMins withinHalfHour blk details is
 
     let zippedTx = zip [0 ..] (Generic.blkTxs blk)
     let txInserter = insertTx tracer cache (getInsertOptions syncEnv) (getNetwork syncEnv) isMember blkId (sdEpochNo details) (Generic.blkSlotNo blk)
-    grouped <- foldM (\grouped (idx, tx) -> txInserter idx tx grouped) mempty zippedTx
-    minIds <- insertBlockGroupedData tracer grouped
+    (blockGroupedData, txFees) <- foldM (\gp (idx, tx) -> txInserter idx tx gp) (mempty, 0) zippedTx
+    minIds <- insertBlockGroupedData tracer blockGroupedData
+
+    -- now that we've inserted all the txs for a ShellyBlock lets put what we need in the cache
+    latestEpochFromDb <- lift queryLatestEpoch
```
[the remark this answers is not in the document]
Cmdv: yeah I was thinking the same. Technically it's there once the epoch has been put into the db, put I put this here as a fail safe. Maybe I should abstract this out as a function and it can call the db when it isn't in cache.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -255,13 +264,13 @@ insertTx tracer cache iopts network isMember blkId epochNo slotNo blockIndex tx
       !txOutsGrouped <- mapM (prepareTxOut tracer cache iopts (txId, txHash)) (Generic.txOutputs tx)
 
       let !txIns = map (prepareTxIn txId Map.empty) resolvedInputs
-      pure $ grouped <> BlockGroupedData txIns txOutsGrouped [] []
+      pure (blockGroupData <> BlockGroupedData txIns txOutsGrouped [] [], groupedFees + fees)
```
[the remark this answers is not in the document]
Cmdv: ahhh so instead of returning a tuple. Yeah that's a good idea. I couldn't spot when there Semigroup instance was but I'll have another look.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
Cmdv: yeah think now I write everything in one go to the cache not the block on it's own. Kept it around just incase but I'll make a note to delete it if I don't.

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -70,89 +61,270 @@ epochInsert trce (BlockDetails cblk details) = do
...
+  case ceBlock cacheEpoch of
+    BlockByron blk ->
+      calculateEpochByron
+        blk
+        ceE
+        cacheEpoch
+        slotDetails
+    BlockShelley blk ->
+      calculateEpochGeneric
+        (Generic.fromShelleyBlock blk)
```
kderme, line 241: I think it isn't necessary to redo the calculations of the Generic Block using `fromShelleyBlock`, `fromAllegraBlock` etc and take cases bassed on the era. This is already done when the block is received. Any needed parts of the block (txs count, output sum etc) can be passed similarly to the fees. In fact I'm not sure it's necessary to pass the whole blk.

cardano-db-sync/src/Cardano/DbSync/Cache/Epoch.hs
```diff
@@ -0,0 +1,90 @@
...
+  ReaderT SqlBackend m ()
+writeCacheEpoch cache cacheEpoch =
+  case cache of
+    UninitiatedCache -> pure ()
+    Cache ci -> liftIO $ atomically $ writeTVar (cEpoch ci) $ Just cacheEpoch
+
+writeBlockAndFeeToCacheEpoch ::
+  MonadIO m =>
+  Cache ->
+  CardanoBlock ->
```
kderme, line 49: Like in the comments above, instead of `CardanoBlock` you could pass directly the fields that you need. Or alternatively the `Generic.Block`. This can avoid us duplicating the logic.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/Grouped.hs
```diff
@@ -68,6 +69,7 @@ instance Semigroup BlockGroupedData where
       (groupedTxMetadata tgd1 <> groupedTxMetadata tgd2)
       (groupedTxMint tgd1 <> groupedTxMint tgd2)
       (groupedTxFees tgd1 <> groupedTxFees tgd2)
```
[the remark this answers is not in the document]
Cmdv: I was unsure if that would break any Semigroup laws 😅 certainly saves me using `sum` on the result 🤦

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -70,89 +45,177 @@ epochInsert trce (BlockDetails cblk details) = do
...
+insertOrReplaceEpoch ::
+  (MonadBaseControl IO m, MonadIO m) =>
+  Cache ->
+  Word64 ->
+  Trace IO Text ->
+  Maybe DB.Epoch ->
+  Maybe EpochInternal ->
+  ReaderT SqlBackend m (Either SyncNodeError ())
+insertOrReplaceEpoch cache slotEpochNum trce latestEpochCache internalEpochCache = do
+  -- get the epoch id using a slot number
```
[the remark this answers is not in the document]
Cmdv: ah yeah course good catch!

angerman:
force-pushed to get hydra re-exicted.

kderme:
I tries to sync with this branch, but the `epoch` table seems to stay unpopulated. Not sure what's causing it yet.
```
# In branch
> select * from epoch;
 id | out_sum | fees | tx_count | blk_count | no | start_time | end_time 
----+---------+------+----------+-----------+----+------------+----------
(0 rows)

# Release
> select * from epoch;
 id  |      out_sum       |     fees     | tx_count | blk_count | no  |     start_time      |      end_time       
-----+--------------------+--------------+----------+-----------+-----+---------------------+---------------------
   1 |  10378568796482912 |      3458053 |       33 |     21587 |   0 | 2017-09-23 21:44:51 | 2017-09-28 21:44:31
   2 | 101402912214214219 |   1033002678 |    12870 |     21590 |   1 | 2017-09-28 21:44:51 | 2017-10-03 21:44:31
   3 |  63479608908424764 |    503075439 |     4292 |     21599 |   2 | 2017-10-03 21:44:51 | 2017-10-08 21:44:31
   4 |  41495794265422302 |    426081777 |     3293 |     21598 |   3 | 2017-10-08 21:44:51 | 2017-10-13 21:44:31

**Comment by Cmdv** [comment]
This issues seems to only happen on Mainnet, so investigating why that is.

**Comment by Cmdv** [comment]
@kderme found the source of the issue which was to do with Boundary blocks being the first block which is ignored in Byron. As we use the first block in new epoch to calculate and create the pervious block. This was never been triggered.

There was also a bug with the counting of the blocks in epochs due to the boundary blocks and how they no longer existed after epoch 175.

This is now ready to go 👍

**Review by kderme — APPROVED** [review]
Good work! I left some comments, but they're mostly minor. I'm fine to merge this as long as this fix https://github.com/input-output-hk/cardano-db-sync/pull/1432 is merged to the pr and the pr is squashed.

File: cardano-db-sync/src/Cardano/DbSync/Epoch.hs [hunk 15]
```diff
@@ -1,159 +1,313 @@
...
+  SyncEnv ->
+  Cache ->
+  Maybe EpochBlockDiff ->
+  Maybe DB.Epoch ->
+  Word64 ->
+  Bool ->
+  ReaderT SqlBackend m (Either SyncNodeError ())
+updateEpochWhenSyncing syncEnv cache mEpochBlockDiff mLastMapEpochFromCache epochNo isBoundaryBlock = do
+  let trce = getTrace syncEnv
+      isFirstEpoch = epochNo == 0
```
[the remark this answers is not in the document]
Cmdv: It was always getting the block counts for the first epoch incorrect on Mainnet but think you are right this must have been left from previous experimentation `isBoundaryBlock` should do the trick right here

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -1,159 +1,313 @@
...
+        Nothing -> do
+          let calculatedEpoch = initCalculateNewEpoch epochBlockDiffCache additionalBlockCount
+          _ <- makeEpochWithDBQuery syncEnv cache (Just calculatedEpoch) epochNo "updateEpochWhenSyncing"
           pure $ Right ()
+        -- simply use cache
+        Just lastMapEpochFromCache -> do
+          let calculatedEpoch = initCalculateNewEpoch epochBlockDiffCache additionalBlockCount
+          void $ writeToMapEpochCache syncEnv cache calculatedEpoch
+          mEpochID <- DB.queryForEpochId epochNo
+          case mEpochID of
```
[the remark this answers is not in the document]
Cmdv: this is to do with server restarts when syncing. It's cases we didn't cover before.

cardano-db-sync/src/Cardano/DbSync/Epoch.hs
```diff
@@ -1,159 +1,313 @@
...
+
+    -- If there isn't an epoch in cache, let's see if we can get one from the db. Otherwise
+    -- calculate the epoch using the expensive db query.
+    Nothing -> do
+      mNewestEpochFromDb <- DB.queryLatestEpoch
+      case mNewestEpochFromDb of
+        -- no latest epoch in db (very unlikely) so lets use expensive db query
+        Nothing -> do
+          makeEpochWithDBQuery syncEnv cache Nothing epochNo "handleEpochWhenFollowing"
+        Just newestEpochFromDb -> do
```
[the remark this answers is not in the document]
Cmdv: yeah you are right will change 👍

Cmdv:
closing due to squash hell in favour of #1433

PR #1381, opened 2023-03-29
Readme.md: Update System Requirements

erikd:


Cmdv (approved):
LGTM :shipit:

erikd:
I updated the `Readme` and CI fails????

erikd:
Since this is only the Readme I am going to merge this anyway.

Cmdv:
CI always fails it seems, which is indicative of a CI but it would be nice for it not to fail 😅

PR #1382, opened 2023-03-29
Improve reward insert speeds by limiting bulking

### LINKED ISSUE #1292: Speedup Reward and EpochStake insertions
Syncing with `disable-leger` provides a considerable speedup. This is because ledger rules are not applied and because `Reward` and `EpochStake` tables are not populated. This provides the motivation to investigate why these tables take so long to populate.

The reward table has a unique key which is checked every time a new entry is inserted. It exists because the reward table is never deleted. After a rollback the same values will be reattempted and the unique key guards against it. Since https://github.com/input-output-hk/cardano-db-sync/issues/1190 is done, there are no rollbacks when db-sync syncs, they only happen when db-sync reaches the tip of the chain. So for the most part, the unique constaint checks provides delays without any benefit. We should try and find a way to delay the migration which creates this unique key untill db-sync reaches the tip of the chain.

Also rewards are added in bug chunks. We should consider limit the size of the insertions. Persistant cache is also affected by it

Similar things apply to EpochStake.

Depends on https://github.com/input-output-hk/cardano-db-sync/issues/1087 https://github.com/input-output-hk/cardano-db-sync/issues/1293

Cmdv:
# Description

Improving reward insert speeds by limiting the size of the insert to hundreds of thousands rather than `n`.
Look at when to add constraints to the `awards` table as they are also meant to slow things down.
- outstanding question [here](https://github.com/input-output-hk/cardano-db-sync/issues/1292#issuecomment-1488583902)

This will fix #1292 

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db/AlterTable.hs
```diff
@@ -0,0 +1,97 @@
...
+  handle alterTableExceptHandler (rawExecute query [])
+  where
+    query :: T.Text
+    query =
+      T.concat
+        [ "ALTER TABLE "
+        , unEntityNameDB (entityDB entity)
+        , " DROP CONSTRAINT "
+        , unConstraintNameDB cname
+        ]
```
Cmdv, line 74: unsure if the check is required, would be nice if it could be a type check not a runtime check just unsure how I would do that!

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -82,17 +88,55 @@ applyAndInsertBlockMaybe syncEnv cblk = do
         rollbackFromBlockNo syncEnv (blockNo cblk)
         insertBlock syncEnv cblk applyRes True tookSnapshot
         liftIO $ setConsistentLevel syncEnv Consistent
+        -- now that we have caught up with the tip of the chain
+        -- we can put the constraints on Reward + EpochStake tables
+        -- as a speed improvement.
+        lift addRewardTableConstraint
```
kderme, line 94: I think there should be an `unless` check in case the unique key exists already. Otherwise there would be `constraind already exists errors`

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -213,7 +257,7 @@ insertLedgerEvents syncEnv currentEpochNo@(EpochNo curEpoch) =
     handler ev =
       case ev of
         LedgerNewEpoch en ss -> do
-          lift $ do
+          lift $
```
kderme, line 260: Was this change proposed by fourmolu? We usually avoid `do` when there is a single line afterwards

cardano-db/src/Cardano/Db/AlterTable.hs
```diff
@@ -0,0 +1,95 @@
...
+  if checkAllFieldsValid entity cols
+    then handle alterTableExceptHandler (rawExecute query [])
+    else liftIO $ throwIO (DbAlterTableException "Constraint field does not exist" sqlError)
+  where
+    query :: T.Text
+    query =
+      T.concat
+        [ "ALTER TABLE "
+        , unEntityNameDB (entityDB entity)
+        , " ADD CONSTRAINT IF NOT EXISTS "
```
kderme, line 54: Is `ADD CONSTRAINT IF NOT EXISTS` a valid close? I can't seem to find it in https://www.postgresql.org/docs/current/sql-altertable.html

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -326,7 +331,7 @@ insertManyUncheckedUnique vtype records =
             . Util.commaSeparated
             $ placeholders
         , " ON CONFLICT ON CONSTRAINT "
-        , unConstraintNameDB (uniqueDBName $ onlyOneUniqueDef (Proxy @record))
+        , unConstraintNameDB foreignConstraintNameDBName
```
kderme, line 334: foreign keys is something different from unique keys. DBSync no longer maintains any foreign keys.

I think you should take cases for this query, since sometimes the unique key constaint will be there but sometimes it won't. I'm not sure if there is a single query that can work for all cases, so db-sync needs to store the current state somewhere in memory.

schema/migration-2-0001-20211003.sql
```diff
@@ -88,7 +88,6 @@ BEGIN
     EXECUTE 'ALTER TABLE "tx_metadata" ADD CONSTRAINT "unique_tx_metadata" UNIQUE("key","tx_id")' ;
     EXECUTE 'ALTER TABLE "tx_metadata" ADD CONSTRAINT "tx_metadata_tx_id_fkey" FOREIGN KEY("tx_id") REFERENCES "tx"("id") ON DELETE CASCADE  ON UPDATE RESTRICT' ;
     EXECUTE 'CREATe TABLE "reward"("id" SERIAL8  PRIMARY KEY UNIQUE,"addr_id" INT8 NOT NULL,"type" rewardtype NOT NULL,"amount" lovelace NOT NULL,"earned_epoch" INT8 NOT NULL,"spendable_epoch" INT8 NOT NULL,"pool_id" INT8 NULL)' ;
-    EXECUTE 'ALTER TABLE "reward" ADD CONSTRAINT "unique_reward" UNIQUE("addr_id","type","earned_epoch","pool_id")' ;
```
kderme, line 91: This will need to be a new migration file which drops the unique key rather than deleting this line. This would allow us to have proper migration, so that users don't need to resync from genesis when upgrading. To create this new file, you could check https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/schema-management.md#creating-a-migration as well as https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/migrations.md for general context about migrations.

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Default.hs
Cmdv: ah no, I took them out my self

cardano-db-sync/src/Cardano/DbSync/Default.hs
Cmdv: ah ok, maybe I do that check inside the function it self. Actually your'e right would be more obvious if that check happens here 👍

cardano-db/src/Cardano/Db/AlterTable.hs
Cmdv: I'm sure I saw it somewhere when I was researching `ADD CONSTRAINT...` I thought above is what you meant when you wrote this in the issue:

> Also before creating the Unique Key it's necessary to first check if it has already been created before. Postgres queries usually have a IF NOT EXISTS clause for this. Alternatively a separate query needs to check it.

Have I got it in the wrong order?

kderme: If they cannot be in the same query, then it should be two separate queries.

cardano-db/src/Cardano/Db/Insert.hs
Cmdv: I couldn't see another way to get to the `Text` representation of a constraint name. I got it from [here](https://github.com/yesodweb/persistent/blob/eaf9d561a66a7b7a8fcbdf6bd0e9800fa525cc13/persistent-test/src/TreeTest.hs#L60)

schema/migration-2-0001-20211003.sql
Cmdv: That's what I thought we did, but I must of got mixed up when we talked about it as I made a note that line should be delete. 😅

cardano-db/src/Cardano/Db/AlterTable.hs
Cmdv: might be able to have it in the same query, just unsure if the way/order I've written it is right. Will search again to see how `IF NOT EXISTS` should be used in this scenario.

schema/migration-2-0001-20211003.sql
kderme: Actualy it may be possible to just delete it, but it needs more special attention.

If there is an upgrade from an existing `13.1.x.x` version and the Unique key is already there this won't cause any issues. What could cause issues is using the next release schema with an existing `13.1.x.x` release. This can happen on a downgrade or when someone uses snapshots created after the change with older release.

So bottom line, yes you could leave it as is, but the unique key should be created when a snapshot is created, in the `runPrepareSnapshot` function.

cardano-db/src/Cardano/Db/AlterTable.hs
Cmdv: You're right it doesn't exists only `IF EXISTS`.

cardano-db-sync/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Revision history for cardano-db-sync
 
-## 13.2.0.0
+## 13.1.1.3
```
kderme, line 3: I think this was changed accidentally.

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -85,6 +91,10 @@ applyAndInsertBlockMaybe syncEnv cblk = do
           void $ migrateStakeDistr syncEnv (apOldLedger applyRes)
           insertBlock syncEnv cblk applyRes True tookSnapshot
           liftIO $ setConsistentLevel syncEnv Consistent
+          -- now that we have caught up with the tip of the chain
+          -- we can put the constraints on rewards table
+          lift addRewardTableConstraint
+          lift addEpochStakeTableConstraint
```
kderme, line 97: This rollback should only happens if at least one block is deleted in `rollbackFromBlockNo`

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Util/Constraint.hs
```diff
@@ -0,0 +1,84 @@
...
+      , dbConstraintEpochStake = resEpochStake
+      }
+
+alterConstraints ::
+  forall m.
+  (MonadBaseControl IO m, MonadIO m) =>
+  ManualDbConstraints ->
+  ReaderT SqlBackend m ()
+alterConstraints ManualDbConstraints {..} = do
+  when dbConstraintRewards addRewardTableConstraint
```
kderme, line 54: I think `unless` should be used instead of `when`. Also this function is never used, but I think it can, see comment below. It couls also be renamed to something like `addConstraintsIfNotExist`

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -28,26 +33,36 @@ import Ouroboros.Network.Point
...
+      (minIds, txInDeleted, deletedBlockCount) <- DB.deleteBlocksBlockId trce blockId
       whenConsumeOrPruneTxOut syncEnv $
         DB.setNullTxOut trce (DB.minTxInId minIds) txInDeleted
       DB.deleteEpochRows epochNo
+      when (deletedBlockCount > 0) $ do
+        -- We use custom constraints to improve input speeds when syncing.
+        -- If they don't already exists we add them here as once a rollback has happened
+        -- we always need a the constraints.
+        unless (dbConstraintRewards dbConstraints) addRewardTableConstraint
+        unless (dbConstraintEpochStake dbConstraints) addEpochStakeTableConstraint
```
kderme, line 63: `alterConstraints` can be used here instead, to avoid repetetion.

cardano-db/src/Cardano/Db/AlterTable.hs
```diff
@@ -0,0 +1,146 @@
...
+        , unConstraintNameDB cname
+        , " UNIQUE("
+        , Text.intercalate "," $ map unFieldNameDB cols
+        , ")"
+        ]
+
+    throwErr :: forall m'. MonadIO m' => [Char] -> ReaderT SqlBackend m' ()
+    throwErr e = liftIO $ throwIO (DbAlterTableException e sqlError)
+
+alterTableDropUniqueConstraint ::
```
kderme, line 85: I assume this can be removed. But also fine to keep it if you prefer.

kderme (changes requested):
`alterConstraints` could be used in two additional places. One is mentioned in the comments above. The other is at the place where the indexes are forced https://github.com/input-output-hk/cardano-db-sync/blob/master/cardano-db-sync/src/Cardano/DbSync/Default.hs#L215 (above or below this unless). The `envDbConstraints` should be updated in both cases.

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -213,8 +218,9 @@ insertBlock syncEnv cblk applyRes firstAfterRollback tookSnapshot = do
       when withinHalfHour $ do
         ranIndexes <- liftIO $ getRanIndexes syncEnv
         unless ranIndexes $ do
+          addConstraintsIfNotExist syncEnv
           unless commited DB.transactionCommit
-          liftIO $ runIndexMigrations syncEnv
+    -- liftIO $ runIndexMigrations syncEnv
```
Cmdv, line 223: @kderme is this what you had in mind?

cardano-db-sync/src/Cardano/DbSync/Util/Constraint.hs
```diff
@@ -0,0 +1,90 @@
...
+addConstraintsIfNotExist ::
+  forall m.
+  (MonadBaseControl IO m, MonadIO m) =>
+  SyncEnv ->
+  ReaderT SqlBackend m ()
+addConstraintsIfNotExist syncEnv = do
+  ManualDbConstraints {..} <- liftIO . readTVarIO $ envDbConstraints syncEnv
+  unless dbConstraintRewards addRewardTableConstraint
+  unless dbConstraintEpochStake addEpochStakeTableConstraint
+  when (dbConstraintRewards && dbConstraintEpochStake) $
```
kderme, line 59: This should always update. The constraints are always created at this point, so the in memory flags should match this.

cardano-db-sync/src/Cardano/DbSync/Util/Constraint.hs
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+addConstraintsIfNotExist ::
+  forall m.
+  (MonadBaseControl IO m, MonadIO m) =>
+  SyncEnv ->
+  ReaderT SqlBackend m ()
+addConstraintsIfNotExist syncEnv = do
+  ManualDbConstraints {..} <- liftIO . readTVarIO $ envDbConstraints syncEnv
+  unless dbConstraintRewards addRewardTableConstraint
```
kderme, line 57: I think we can log before and after `addRewardTableConstraint`, when it's executed. Same for `addEpochStakeTableConstraint`.

cardano-db-sync/src/Cardano/DbSync/Default.hs
kderme: Not exactly. This is different from the indexes creation. `addConstraintsIfNotExist` can  be called after the indexes are created and shouldn't be affected by `ranIndexes` flag.

kderme: For context: Indexes are automatically created by db-sync when it reaches 30min before the tip of the chain. This allows users to run fast queries. We can also create the unique keys at the same point if they are missing. They are not yet needed by db-sync, but unique keys create automatically indexes, so they can speed up queries by users.

Cmdv: ah ok the previous comment stated "(above this unless)" did you mean before `unless ranIndexes $ do`?

cardano-db-sync/src/Cardano/DbSync/Util/Constraint.hs
Cmdv: yeah cool I'll add some logs 👍

cardano-db-sync/src/Cardano/DbSync/Default.hs
kderme: Yes, it can go above or below (not inside)

kderme (approved):

Issue #1333, opened 2023-01-10
Split out addresses table from tx_out


rdlrt:
**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.0.5
PostgreSQL version: 14.4

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: N.A.

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): systemd

**Additional context**
The background for this request comes from Koios , for an endpoint that allows users to query addresses associated with stake_address causing too long (ones which have millions of transaction entries) - but  this applies to almost all query layers that use dbsync. When _not_ part of Postgres cache , simply querying observations themselves could take a while (at times can take 15-50s):

``` bash
cexplorer=# select count(1) from tx_out where stake_address_id = 3350960;
  count
---------
 2648923
(1 row)

Time: 23476.114 ms (00:23.476)
```
Ofcourse - once in cache, this would run in a few milliseconds.
However, the outcome (unique address list) being sought from same query was simply 12 addresses.

This made us think of starting to maintain seperate cache table for stake_address to address mapping, but it makes little sense to only do it on client side rather than dbsync itself, as the advantages will be higher than this scope.

**Problem Report**

The request here is to move address field to it's own table, alongwith duplicated information that currently resides in tx_out table.
A new table `addresses` that might contain columns below, that are currently in `tx_out`:
```
|id|address | address_raw (optional) | payment_cred (optional) | stake_address_id | address_has_script |
```
Similarly, the above columns could be removed from `tx_out` and replaced with `addr_id` which would link to address.id field.

**Advantages:**

1. Space and IOPs:
  Currently , size of tx_out table is ~40G and it's indexes is ~20G. A huge chunk of this size is due to repitition of address information across multiple rows every time a transaction references an address. In turn, this also increases index size which could would much larger footprint to process, replacing it with a single address_id field saves on both fronts as well as lookup in tx_out table. The average savings achieved might be ~25-30G atm ( + indexes)
2. Reduction in duplications:
  Columns `address`, `address_raw` and `payment_cred` can all be derived from address itself. While it might have been initially for UX, the duplication of each of these long char fields for tx does more harm than good in tx_out table. They can still be maintained in separate address table.
3. Transient state of `address_hash_script`:
  My understanding here might be incorrect, but `address_has_script` seems like a permanent state rather than a transient state - thus, does not seem to be right candidate in tx_out table.
[...]

PR #1396, opened 2023-04-25
Introduce a separate Address table

kderme:
# Description

Fixes https://github.com/input-output-hk/cardano-db-sync/issues/1333
# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/schema/migration-2-0026-20230422.sql
+++ b/schema/migration-2-0026-20230422.sql
@@ -1,0 +1,24 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 26 THEN
+    EXECUTE 'ALTER TABLE "tx_out" ADD COLUMN "address_id" INT8 NOT NULL' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address_raw"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address_has_script"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "payment_cred"' ;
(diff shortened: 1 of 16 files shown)


schema/migration-2-0026-20230422.sql
```diff
@@ -0,0 +1,24 @@
...
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 26 THEN
+    EXECUTE 'ALTER TABLE "tx_out" ADD COLUMN "address_id" INT8 NOT NULL' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address_raw"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "address_has_script"' ;
+    EXECUTE 'ALTER TABLE "tx_out" DROP COLUMN "payment_cred"' ;
+    EXECUTE 'CREATe TABLE "address"("id" SERIAL8  PRIMARY KEY UNIQUE,"address" VARCHAR NOT NULL,"address_raw" BYTEA NOT NULL,"has_script" BOOLEAN NOT NULL,"payment_cred" hash28type NULL)' ;
```
rdlrt, line 14: Could this table have an additional field inserted (`stake_address_id`)? Useful to filter used addresses contributing towards a stake.
Absence of this field would require querying `tx_out` table and then doing a [distinct] fetch on `address_id` together

infnada:
Is this and/or should this MR also take into account `collateral_tx_out`? 

Also, i don't know if it's possible to have an `address` inside `collateral_tx_out` without having it in `tx_out`.

Not really familiar with collaterals, i would say it is possible since a collateral output can be sent to any account (assumption), and at the same time a collateral must have a `tx_out` where to use and reference it for "a future `tx_in`" in the chain, so i'm a little confused :). If that's the case and we don't handle `collateral_tx_out` we will end up with an `address` table not really containing all the addresses.

rdlrt:
@kderme - are there plans for this PR to progress? would be nice to get an insight around any blockers for this one

kderme:
Done here https://github.com/IntersectMBO/cardano-db-sync/pull/1820

Issue #1397, opened 2023-05-01
[Feature Request] - Flag to prune used UTxOs


rdlrt:
**Problem Report**

This is a feature available in other chain indexers, and while it might add complexity (or may not be suitable), would be good to have - atleast as a backlog.
As the size of chain increases, it will become increasingly improbable that we could keep scaling the entire blockchain database, atleast for specific use cases that do not care about history. It would be a nice addition to dbsync flags if it could operate with only current set of UTxOs (discarding old used UTxOs that are finalized, ones that do not qualify for rollback). This would tie in especially nicely with #1389 , to allow those who want to manage the queries (while considering rollback for more recent transactions) to build using the additional column `tx_out.consumed_by_tx_in_id`.

This change would heavily help out with resource usage of the underlying database, will massively speedup the restore from snapshots as well as query times - should one opt to run in this mode. The target audience will likely be wallets, marketplaces, dApps, specific analytics, lite mode for query layers like Koios.

Ofcourse - this might defer from original design of dbsync which is append-only, so understand if it's too big of an impact to go through in shorter timeframe, but it would be a useful addition should there be a consideration in future

kderme:
This is an interesting option. Indeed it defers from the original append-only design, but the start has been made with https://github.com/input-output-hk/cardano-db-sync/pull/1389 and this is the next step. 

I think it could work with a garbage collection of the `tx_out` table, which will happen ie once every epoch. It will delete `tx_out`entries which are already consumed, by checking the `tx_out.consumed_by_tx_in_id` field. Queries will still have to check the  `tx_out.consumed_by_tx_in_id` being null to find utxo, since recently consumed outputs will thill exist in the table.

Similarly to https://github.com/input-output-hk/cardano-db-sync/pull/1389 it will be an optional extension of the schema.

kderme:
This is done in https://github.com/input-output-hk/cardano-db-sync/pull/1416. It's an optional migration, so it doesn't affect released snapshots or breaks compatibility with users that don't use the related flag.

PR #1389, opened 2023-04-04
Introduce tx_out.consumed_by_tx_in_id

kderme:
# Description

Fixes https://github.com/input-output-hk/cardano-db-sync/issues/1314

# Checklist

- [ ] Commit sequence broadly makes sense
- [ ] Commits have useful messages
- [ ] New tests are added if needed and existing tests are updated
- [ ] Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)
- [ ] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`
- [ ] Self-reviewed the diff

# Migrations

- [x] The pr causes a [breaking change](https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/migrations.md) of type a,b or c
- [ ] If there is a breaking change, the pr includes a database migration and/or a fix process for old values, so that upgrade is possible
- [ ] Resyncing and running the migrations provided will result in the same database semantically

It will eventually be a b level breaking change. It needs a proper migration

PR #1398, opened 2023-05-01
Prune consumed tx_out

kderme:
# Description

This is a prototype for https://github.com/input-output-hk/cardano-db-sync/issues/1397. It doesn't correctly handles rollbacks or migrations, but it allows to benchmark syncing and querying the feauture. It builds on top of https://github.com/input-output-hk/cardano-db-sync/pull/1389

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Byron/Genesis.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Byron/Genesis.hs
@@ -152,7 +152,7 @@ validateGenesisDistribution tracer networkName cfg bid =
       Left err -> dbSyncNodeError $ "validateGenesisDistribution: " <> textShow err
       Right expectedSupply ->
         when (expectedSupply /= totalSupply) $
-          dbSyncNodeError $
+          liftIO $ logInfo tracer $
             Text.concat
               [ "validateGenesisDistribution: Expected total supply to be "
               , DB.renderAda expectedSupply


cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -146,6 +146,8 @@ insertBlock syncEnv cblk applyRes firstAfterRollback tookSnapshot = do
           Generic.fromBabbageBlock (ioPlutusExtra iopts) (getPrices applyResult) blk
   insertEpoch details
   lift $ commitOrIndexes withinTwoMin withinHalfHour
+  when (unBlockNo blkNo `mod` 21600 == 0) $ do
```
Cmdv, line 149: think that's meant to be `2160` not `21600`.

I ended up using something like bellow rather than hard coding this value:

```haskell
getSecurityParam :: SyncEnv -> Word64
getSecurityParam syncEnv =
  case envLedgerEnv syncEnv of
    HasLedger hle -> getSecurityParameter $ leProtocolInfo hle
    NoLedger nle -> getSecurityParameter $ nleProtocolInfo nle          
```
not sure that would be of use 🤷

kderme: 21600, or 10 * securityParameter, is the expected number of blocks per epoch on mainnet, so by doing this periodicaly every 21600 it's done once per epoch.
Initially I wanted to do this pruning at the first block of the epoch, however the block itself doesn't carry this information, it requires the ledger state and we have options that disable it. So this was the easiest check I can make to have a realistic benchmark just by looking at the block itself (no additional state).

rdlrt:
Apologies as it's not very clear to me, will the delete 'cascade' the references for tx_out from ma_tx_out as well (as it has a max of 900 1-to-many mapping from tx_out)?

kderme:
> Apologies as it's not very clear to me, will the delete 'cascade' the references for tx_out from ma_tx_out as well (as it has a max of 900 1-to-many mapping from tx_out)?

Indeed this is not done in this pr yet, though it should. I was running benchmarks with `disable-multiassets` and related flags to see the performance diff related to tx_out.

rdlrt:
Gotcha - thanks. Just wanted to be sure it's not missed 🙂

kderme:
Superseded by https://github.com/input-output-hk/cardano-db-sync/pull/1416

PR #1419, opened 2023-05-30
Get security param from config

### LINKED ISSUE #1418: Remove hardcoded use of securityParam in code
In places there is the use of `2160` which should ideally be taken from the value `securityParam` used in the config files.

This is especially helpful when running tests so we can minimise the rollback size.

Cmdv:
# Description

This fixes #1418 
In the future we'll need to split this out as `Api.hs` is getting rather big and with lots of mixed functionality. 

# Checklist

- [ ] Self-reviewed the diff

# Migrations


kderme (approved):
LGTM!

Issue #1426, opened 2023-06-01
Feature request: Make writing of ledger state to risk asynchronous


erikd:
Currently the ledger state is written to disk once per epoch and this is done synchronously with the maintenance of ledger state.

SInce writing this state can take up to 10 minutes on my machine, I think it would be really helpful if this was done asynchronously.

erikd:
I think this can just be done in an `async` there the created thread just does the write and then terminates. Meanwhile the main thread can just continue on its way. Its a bit naive but there should not be any real horrible corner cases.

PR #1427, opened 2023-06-01
Write the ledger state file asynchronously

erikd:
In epoch above 340 or so, encoding and writing a ledger state file to disk can take 10 minutes or more. Better to do this asynchronously.

Closes: https://github.com/input-output-hk/cardano-db-sync/issues/1426

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/DbAction.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/DbAction.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
 module Cardano.DbSync.DbAction (
@@ -48,7 +47,9 @@ newDbActionQueue :: IO DbActionQueue
 newDbActionQueue =
   -- Use an odd number here so that the db_tip_height metric increments by this odd number
   -- when syncing, instead of incrementing by say 100.
-  DbActionQueue <$> TBQ.newTBQueueIO 117
+  -- The pipeline queue in the LocalChainSync machinery is 50 elements long
+  -- so we should not exceed that.
+  DbActionQueue <$> TBQ.newTBQueueIO 47
 
 writeDbActionQueue :: DbActionQueue -> DbAction -> STM ()
 writeDbActionQueue (DbActionQueue q) = TBQ.writeTBQueue q
--- a/cardano-db-sync/src/Cardano/DbSync/Sync.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Sync.hs
@@ -32,7 +32,7 @@ import qualified Cardano.Crypto as Crypto
 import Cardano.Db (runDbIohkLogging)
 import qualified Cardano.Db as Db
 import Cardano.DbSync.Api
-import Cardano.DbSync.Api.Types (ConsistentLevel (..), FixesRan (..), LedgerEnv (..), RunMigration, SyncEnv, SyncOptions (..), envConnString, envLedgerEnv, envNetworkMagic, envOptions)
+import Cardano.DbSync.Api.Types (ConsistentLevel (..), FixesRan (..), RunMigration, SyncEnv, SyncOptions (..), envConnString, envLedgerEnv, envNetworkMagic, envOptions)
 import Cardano.DbSync.Config
 import Cardano.DbSync.Database
 import Cardano.DbSync.DbAction
@@ -41,6 +41,7 @@ import Cardano.DbSync.Era
 import Cardano.DbSync.Error
 import Cardano.DbSync.Fix.PlutusDataBytes
 import Cardano.DbSync.Fix.PlutusScripts
+import Cardano.DbSync.Ledger.State
 import Cardano.DbSync.LocalStateQuery
 import Cardano.DbSync.Metrics
 import Cardano.DbSync.Tracing.ToObjectOrphans ()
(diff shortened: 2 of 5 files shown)


erikd:
Currently running a mainnet sync from epoch 346 to chain tip as a test.

erikd:
Funny thing is that I am now seeing:
```
[2023-06-01 07:12:36.04 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/64540777-106ba5f091-346.lstate. It took 56.596637642s.
[2023-06-01 07:14:22.52 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/64972791-b879ba7b08-347.lstate. It took 57.049584059s.
[2023-06-01 07:16:02.64 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/65404734-b0779b7199-348.lstate. It took 56.068380846s.
[2023-06-01 07:17:47.38 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/65836796-b3082749a7-349.lstate. It took 57.538157427s.
[2023-06-01 07:19:29.54 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/66268791-374128a874-350.lstate. It took 59.209895878s.
[2023-06-01 07:21:49.13 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/66700756-3f0facd76c-351.lstate. It took 86.673485992s.
[2023-06-01 07:23:38.23 UTC] Asynchronously wrote a ledger snapshot to ledger-state/mainnet/67132766-9921b7c5fb-352.lstate. It took 59.453359303s.
```
This files are being written **MUCH** faster than they were previously (epoch 340 - 345) where it was often 10 minutes each.

I wonder if part of the problem is memory fragmentation.

kderme:
What stops from having multiple threads storing a snapshot concurrently? This could cause important issues. I believe there should be some guard which for example doesn't fork a new thread when ther is already one running.

erikd:
> What stops from having multiple threads storing a snapshot concurrently?

As long as they are storing different snapshots to different files, I am pretty certain that cannot happen.

Anyway, that's why this is a draft/wip PR. I will think about this some more.

erikd:
> I believe there should be some guard which for example doesn't fork a new thread when there is already one running.

I think the obvious solution (in line with the rest of `db-sync`) is to have the DB thread put the ledger state on a `TBQueue` (a shallow one, maybe 5 elements deep) and then have a thread that reads the queue and writes the files. If the `TBQueue` is full, the DB thread will block until there is room for a new entry.

erikd:
The code now uses a `TBQueue` (of depth 5) in a separate thread.

I am testing this by deleting all files in the ledger state directory, forcing it to replay ledger state from genesis.

cardano-db-sync/src/Cardano/DbSync/LedgerState.hs
```diff
@@ -408,7 +421,27 @@ saveCurrentLedgerState env ledger mEpochNo = do
...
               ["File ", Text.pack file, " exists"]
-        else do
+        else
+          atomically $ writeTBQueue (leStateWriteQueue env) (file, lState)
+
+runLedgerStateWriteThread :: Trace IO Text -> LedgerEnv -> IO ()
+runLedgerStateWriteThread tracer lenv =
+  case lenv of
+    HasLedger le -> ledgerStateWriteLoop tracer (leStateWriteQueue le) (configCodec $ getTopLevelconfigHasLedger le)
+    NoLedger _ -> forever $ threadDelay 600000000 -- 10 minutes
```
Cmdv, line 431: Out of interest how come we wait 10mins when there is no ledger? (don't actually know the ledger state part of the code yet)

cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -334,13 +336,15 @@ dbSyncProtocols syncEnv metricsSetters _version codecs _connectionId =
                             (runDbThread syncEnv metricsSetters actionQueue)
                             (runOfflineFetchThread syncEnv)
                         )
-                        ( runPipelinedPeer
-                            localChainSyncTracer
-                            (cChainSyncCodec codecs)
-                            channel
-                            ( chainSyncClientPeerPipelined $
-                                chainSyncClient metricsSetters tracer (fst <$> latestPoints) currentTip actionQueue
+                        ( race_
```
Cmdv, line 339: Another inquisitive question, how come we race here? which ever finishes first would cancel the other but not grasping which is likely to win/lose?

cardano-db-sync/src/Cardano/DbSync/LedgerState.hs
erikd: This is one branch of a `race_` . If there is no ledger, there is no ledger file to write. If we return immediately, the `race_` will terminate the other thread.

cardano-db-sync/src/Cardano/DbSync/Sync.hs
erikd: They both should run forever, so we don't care which wins of loses.

Cmdv: ah cool, so it's just about being able to do them concurrently 👍

cardano-db-sync/src/Cardano/DbSync/DbAction.hs
```diff
@@ -48,7 +47,9 @@ newDbActionQueue :: IO DbActionQueue
 newDbActionQueue =
   -- Use an odd number here so that the db_tip_height metric increments by this odd number
   -- when syncing, instead of incrementing by say 100.
-  DbActionQueue <$> TBQ.newTBQueueIO 117
+  -- The pipeline queue in the LocalChainSync machinery is 50 elements long
+  -- so we should not exceed that.
+  DbActionQueue <$> TBQ.newTBQueueIO 47
```
kderme, line 52: Why is it an issue to exceed the LocalChainSync queue?

erikd: The the local chain sync protocol exceeds its depth, it causes a disconnect and the disconnect/reconnect cycle takes a non-zero amount of time. Before this change, with just a ledger state replay, I was seeing this for nearly every epoch.

kderme (approved):

PR #1429, opened 2023-06-02
Test new tx-out flags

### LINKED ISSUE #1417: Add tests for consumed and pruned tx_out migrations
https://github.com/input-output-hk/cardano-db-sync/pull/1416 introduces 2 additional optional migrations that are executed when flags `consumed-tx-out` or `prune-tx-out` are on. Related tickets are https://github.com/input-output-hk/cardano-db-sync/issues/1397 and https://github.com/input-output-hk/cardano-db-sync/issues/1314. Due to lack of testing we only support them experimentally. It would be great to have some unit tests for these options, which would give us more confidence.

### LINKED ISSUE #1418: Remove hardcoded use of securityParam in code

Cmdv:
# Description

This fixes #1417 adding tests to new flags (`--consumed-tx-out` , `--prune-tx-out` ) that were introduced which deal with tx-out consuming and pruning.

Also a slight restructure to start tests on feature flags

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db.hs
```diff
@@ -7,14 +7,13 @@ module Cardano.Db (
...
   TxOut (..),
   gitRev,
-  migrateTxOut,
 ) where
 
 import Cardano.Db.Delete as X
 import Cardano.Db.Error as X
 import Cardano.Db.Insert as X
 import Cardano.Db.Migration as X
-import Cardano.Db.Migration.Extra.CosnumedTxOut.Queries (migrateTxOut)
```
kderme, line 17: I would prefer to keep the scope of the Extra.Migration queries as limited as possible, since it defines queries on a different schema, that can cause issues if executed against the default schema.

Cmdv: cool, I'll just export the function I need for the tests 👍

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Tx.hs
```diff
@@ -59,3 +62,53 @@ consumeSameBlock =
...
+
+    -- add a couple of transactions
+    -- void $ withBabbageFindLeaderAndSubmit interpreter mockServer $ \st -> do
+    --   tx0 <- Babbage.mkPaymentTx (UTxOIndex 0) (UTxOIndex 1) 20000 20000 st
+    --   let utxo0 = head (Babbage.mkUTxOBabbage tx0)
+    --   tx1 <- Babbage.mkPaymentTx (UTxOPair utxo0) (UTxOIndex 2) 10000 500 st
+    --   pure [tx0, tx1]
+
+    -- add more blocks to
+    b2 <- fillUntilNextEpoch interpreter mockServer
```
kderme, line 107: `forgeAndSubmitBlocks` could be used instead to create a specific number of blocks.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Tx.hs
```diff
@@ -59,3 +62,53 @@ consumeSameBlock =
...
+  let txOutParam =
+        TxOutParam
+          { paramMigrateConsumed = True,
+            paramPruneTxOut = True
+          }
+  withTxOutParamConfig txOutParam babbageConfigDir testLabel $ \interpreter mockServer dbSync -> do
+
+    startDBSync dbSync
+    -- fill in some blocks
+    b1 <- fillEpochPercentage interpreter mockServer 50
```
kderme, line 94: forgeAndSubmitBlocks could be used instead to create a specific number of blocks.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Alonzo/Tx.hs
```diff
@@ -39,3 +68,107 @@ consumeSameBlock =
     assertBlockNoBackoff dbSync 1
   where
     testLabel = "consumeSameBlock-alonzo"
+
+consumeTxOutMigratedAlz :: IOManager -> [(Text, Text)] -> Assertion
+consumeTxOutMigratedAlz = do
```
kderme, line 73: Probably I missed to mention, we try to do new test development only on the latest era (currently Babbage). We keep some tests from older eras just to be sure we didn't break something. So could these tests be migrated to the Babbage.hs module and using Babbage utilities? Eventually we want to have a seemless way to write tests for all eras, which will also reduce the cost of migrating the tests when a new era appears.

Cmdv: ah yeah cool I just assumed they both needed them. I have these tests in Babbage era already as that's where I started them. I'll just remove theses as they are exactly the same really. You're right though we should probably have a way to write one test that works across multiple eras :)

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage.hs
```diff
@@ -56,6 +56,10 @@ unitTests iom knownMigrations =
         [ test "simple tx" BabTx.addSimpleTx
         , test "simple tx in Shelley era" BabTx.addSimpleTxShelley
         , test "consume utxo same block" BabTx.consumeSameBlock
+        , test "migrate to consume tx-out" BabTx.consumeTxOutMigrated
```
kderme, line 59: Maybe it's worth adding these tests in a new testGroup dedicated to the flags? This group is for simple transactions without a rollback and there is a separate group for rollbacks. In that case though they would be scattered

kderme (approved):
This looks like great start! We can either try to merge this and then you can add more tests on a separate pr, or you can append more tests in this pr.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Tx.hs
```diff
@@ -2,60 +2,170 @@ module Test.Cardano.Db.Mock.Unit.Babbage.Tx (
...
+    startDBSync dbSyncEnv
+    -- add 50 block
+    b1 <- forgeAndSubmitBlocks interpreter mockServer 50
+
+    -- add 2 blocks with tx
+    void $ withBabbageFindLeaderAndSubmitTx interpreter mockServer $ Babbage.mkPaymentTx (UTxOIndex 0) (UTxOIndex 1) 10000 10000
+    void $ withBabbageFindLeaderAndSubmitTx interpreter mockServer $ Babbage.mkPaymentTx (UTxOIndex 1) (UTxOIndex 0) 10000 10000
+    assertBlockNoBackoff dbSyncEnv (fromIntegral $ length b1 + 2)
+    -- check tx-out count before any pruning has happened
+    assertEqQuery dbSyncEnv DB.queryTxOutCount 14 "new epoch didn't prune tx_out column that are null"
```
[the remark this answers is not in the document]
Cmdv: it's checking the count "before" any has happened

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Tx.hs
```diff
@@ -2,60 +2,170 @@ module Test.Cardano.Db.Mock.Unit.Babbage.Tx (
...
       tx1 <- Babbage.mkPaymentTx (UTxOPair utxo0) (UTxOIndex 2) 10000 500 st
       pure [tx0, tx1]
-    assertBlockNoBackoff dbSync 1
+    assertBlockNoBackoff dbSyncEnv 1
   where
     testLabel = "consumeSameBlock"
+
+consumeTxOutMigrated :: IOManager -> [(Text, Text)] -> Assertion
+consumeTxOutMigrated = do
+  let txOutParam =
```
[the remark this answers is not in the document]
Cmdv: ah ok used to having variables at the top but I'll stick them in a `where` 👍

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Tx.hs
```diff
@@ -2,60 +2,170 @@ module Test.Cardano.Db.Mock.Unit.Babbage.Tx (
...
+    blk1 <- forgeNext interpreter mockBlock1
+    blk2 <- forgeNext interpreter mockBlock2
+    atomically $ addBlock mockServer blk0
+    startDBSync dbSyncEnv
+    atomically $ addBlock mockServer blk1
+    atomically $ addBlock mockServer blk2
+    assertBlockNoBackoff dbSyncEnv 3
+
+    atomically $ rollback mockServer (blockPoint blk1)
+    assertBlockNoBackoff dbSyncEnv 3 -- rollbacks effects are now delayed
```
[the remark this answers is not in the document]
Cmdv: I just used existing rollback tests, ideally I'd like to insert 110+ blocks let a prune happen rollback and check everything was removed correctly. Just wasn't sure on doing that correctly. Will try again

kderme (approved):
Looks like a great first step for https://github.com/input-output-hk/cardano-db-sync/pull/1429. The last commit [add pruning tests to Alonzo era](https://github.com/input-output-hk/cardano-db-sync/pull/1429/commits/31b1b177aad4d4852160d5f6953e67592fd72505) maybe needs a better name or a squash? Since there are no tests in Alonzo era

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage.hs
```diff
@@ -33,6 +34,10 @@ unitTests iom knownMigrations =
         , test "node restart" BabSimple.nodeRestart
         , test "node restart boundary" BabSimple.nodeRestartBoundary
         ]
+    , testGroup
+        "flags"
+        [ test "conssumed-tx-out" Flags.flagConsumedTxOut
+        ]
```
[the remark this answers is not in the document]
Cmdv: ahh I see what you're saying. Yeah good idea, we can nest the type of flag when dealing with the other flags too 👍

sgillespie (approved):
LGTM

PR #1431, opened 2023-06-05
Reduce cyclic errors by separating out Ledger & Api

Cmdv:
# Description

Due to some of the structure of the files we'd end up with scenarios like:
```
mkSyncEnvFromConfig ::
  Trace IO Text ->
  ConnectionString ->
  SyncOptions ->
  Maybe LedgerStateDir ->
  Bool ->
  GenesisConfig ->
  Bool ->
  Bool ->
  Bool ->
  Bool ->
  RunMigration ->
```
Having 4 `Bools` might cause issues if inserted in an incorrect order. The solution is to pass `SyncNodeParams` but doing so causes a cyclic dependency error.  This PR separates files apart to fix the cyclic dependency error.
 
# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/cardano-db-sync.cabal
+++ b/cardano-db-sync/cardano-db-sync.cabal
@@ -43,6 +43,7 @@ library
   exposed-modules:      Cardano.DbSync
 
                         Cardano.DbSync.Api
+                        Cardano.DbSync.Api.Types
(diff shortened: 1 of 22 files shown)


kderme (approved):
LGTM!

PR #1433, opened 2023-06-06
Cache epoch calculations to minimise use of DB queries

Cmdv:
# Description
_This was originally https://github.com/input-output-hk/cardano-db-sync/pull/1380 but everything blew up when I tried to squash the commit history 😢_

This fixes https://github.com/input-output-hk/cardano-db-sync/issues/708 which introduces a Cache to minimise the expensive DB lookup / Calculations used when working out new Epoch which happens as a reaction to new Blocks being added and in following mode.

To achieve this we're keeping Epoch and last know blockId in cache. Then on each new lookup instead of calculating from zero we can do a lookup of only the blocks we are missing and add the results to our epoch in cache. Thus resulting in a new Epoch that can then both be updated in cache and db with minimal overhead.

This PR also includes "Disabling the epoch table when there is no cache".

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -80,6 +80,11 @@ jobs:
         ls -al /var/run/postgresql/.s.PGSQL.5432 || true
         ls -al || true
 
+    - name: Debug Pkgconfig
+      run: |
+        echo $PKG_CONFIG_PATH
+        pkg-config --list-all
+
     - name: Haskell versions
       run: |
         ghc --version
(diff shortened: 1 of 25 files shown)


kderme (approved):
LGTM! Some remaining issues can be resolved on a future pr, like
- queryEpochNoForLastTwoBlocks is not used anywhere
- tests to be added 
- fix the case with --disable-cache

kderme (approved):

kderme (approved):

PR #1436, opened 2023-06-14
Allow to run docker with DISABLE_LEDGER (#1430)

### LINKED ISSUE #1430: Allow to run docker without `--state-dir`
db-sync can be executed with `--disable-ledger` which disables the ledger state. When it is on the `state-dir` option shouldn't be provided, since the directory is not needed to store snapshots. Release 13.1.1.0 initially crashed correctly if both options were enabled, but due to an issue with docker https://github.com/input-output-hk/cardano-db-sync/issues/1409 the crash was replaced by a warning in 13.1.1.2 and master

It would be great to have a way to run db-sync from docker with `--disable-ledger`, but without the `state-dir` option set.

sgillespie:
# Description

Fixes #1430. This change allows users to pass an environment variable `DISABLE_LEDGER` to the docker container (and the NixOS module). If this flag is non-empty, it will pass `--disable-ledger` and omit the `--state-dir ...` option. 

Note that even when `services.cardano-db-sync.disableLedger` is true, `stateDir` is still defaulted and required, because we write `pgpass` there.

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -136,9 +136,13 @@ let
         set -euo pipefail
         ${scripts.mainnet.db-sync.passthru.service.restoreSnapshotScript}
 
-        exec $DBSYNC \
-          --schema-dir ${../schema} \
-          --state-dir ${scripts.mainnet.db-sync.passthru.service.stateDir} $@
+        if [[ "''${DISABLE_LEDGER:-N}" == "Y" ]]; then
+          LEDGER_OPTS="--disable-ledger"
+        else
+          LEDGER_OPTS="--state-dir ${scripts.mainnet.db-sync.passthru.service.stateDir}"
+        fi
+
+        exec $DBSYNC --schema-dir ${../schema} ''${LEDGER_OPTS} $@
       ${clusterStatements}
       else
         echo "Managed configuration for network "$NETWORK" does not exist"
--- a/nix/nixos/cardano-db-sync-service.nix
+++ b/nix/nixos/cardano-db-sync-service.nix
@@ -126,6 +126,14 @@ in {
         type = lib.types.nullOr lib.types.str;
         default = "/var/lib/cexplorer";
       };
+      disableLedger = lib.mkOption {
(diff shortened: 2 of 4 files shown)


nix/docker.nix
```diff
@@ -136,9 +136,13 @@ let
         set -euo pipefail
         ${scripts.mainnet.db-sync.passthru.service.restoreSnapshotScript}
 
-        exec $DBSYNC \
-          --schema-dir ${../schema} \
-          --state-dir ${scripts.mainnet.db-sync.passthru.service.stateDir} $@
+        if [[ -n "$DISABLE_LEDGER" ]]; then
```
andreabedini, line 139: I understand you just want to check whether `DISABLE_LEDGER` is set or not but, given you write `DISABLE_LEDGER=true` in the documentation above, I am worried someone will do `DISABLE_LEDGER=false` :joy:.
Maybe checking for an explicit value of `DISABLE_LEDGER` (e.g. "true" or 1) could be a safer option?

nix/nixos/cardano-db-sync-service.nix
```diff
@@ -177,6 +185,20 @@ in {
             (if (cfg.environment.nodeConfig.RequiresNetworkMagic == "RequiresNoMagic" )
               then "--mainnet"
               else "--testnet-magic $(jq '.networkMagic' ${cfg.environment.nodeConfig.ShelleyGenesisFile})");
```
andreabedini, line 187: Just a thought, could we get the networkMagic in nix rather than going through `jq`?

nix/docker.nix
sgillespie: That's a good  point. After looking around it, it looks like there's an existing pattern (`RECREATE_DB=Y`), so I think it would make sense to follow that pattern. Source: https://github.com/input-output-hk/cardano-db-sync/blob/master/scripts/postgresql-setup.sh#L311

kderme (approved):

PR #1443, opened 2023-06-21
fourmolu everything

Cmdv:
# Description

looks like most of the project was fourmolu-rised but `./cardano-chain-gen/` after running `fourmolu --mode inplace $(find . -name '*.hs')` in root directory.

Added `single-constraint-parens=never` to `fourmolu.yaml` file as the default in version 0.12 was changed to use `always` by default.

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/fourmolu.yaml
+++ b/fourmolu.yaml
@@ -11,4 +11,5 @@ let-style: auto
 in-style: right-align
 unicode: never
 respectful: true
+single-constraint-parens: never
 fixities: []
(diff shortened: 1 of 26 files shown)


config/pgpass-mainnet
```diff
@@ -1 +1 @@
-/var/run/postgresql:5432:cexplorer:*:*
+localhost:5432:cexplorer:*:*
```
kderme, line 1: Possibly leftover local diff?

Cmdv: good spot yes it was!! there will be another too, let me take them out :)

fourmolu.yaml
```diff
@@ -11,4 +11,5 @@ let-style: auto
 in-style: right-align
 unicode: never
 respectful: true
-fixities: []
\ No newline at end of file
+single-constraint-parens: never
```
sgillespie, line 14: This won't have an effect until after we upgrade to `0.12.0.0`, and it will cause another mass change. I have no problem with this, just noting.

sgillespie (approved):

fourmolu.yaml
Cmdv: In `0.10.1.0` it defaults to `never` but in `0.12.0.0` it's `always` so doing this just means our code won't change once we upgrade 😄

kderme (approved):

disassembler (approved):

PR #1444, opened 2023-06-21
remove tullia

dermetfan:
# Description

Remove tullia/cicero jobs as we are going to use Hydra instead. Hydra jobs are already in place and we can start using them without any changes.

# Checklist

- ~~New tests are added if needed and existing tests are updated~~
- ~~Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)~~
- ~~Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) (which can be run with `scripts/fourmolize.sh`~~
- [x] Self-reviewed the diff

# Migrations

- ~~The pr causes a [breaking change](https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/migrations.md) of type a,b or c~~
- ~~If there is a breaking change, the pr includes a database migration and/or a fix process for old values, so that upgrade is possible~~
- ~~Resyncing and running the migrations provided will result in the same database semantically~~


Cmdv (approved):
LGTM 👍

edit: oops still in draft, I was too keen!! 😂

Cmdv (approved):
Still LGTM :shipit:

angerman (approved):
LGTM

PR #1445, opened 2023-06-22
Upload docker images to ghcr.io after release

sgillespie:
# Description

TODO[sgillespie]

# Checklist

- [ ] Self-reviewed the diff

# Migrations


angerman (approved):
This looks great!

just one question: why do we add the detached docker compose up command in this PR? Is this related?

sgillespie:
> just one question: why do we add the detached docker compose up command in this PR? Is this related?

Not related, it just looks like a fairly obvious typo to me. If it's `docker-compose up && docker-compose logs -f`, the second command will never run because the services are long-running.

PR #1446, opened 2023-06-22
Update CODEOWNERS

sgillespie:
# Description

Update CODEOWNERS. In particular:

 * Add a new group _core-tech-dbsync_
 * Add core-tech-devx to DevOps and Docs files
 * Only use groups

# Checklist

- [x] Self-reviewed the diff

# Migrations


kderme (approved):
Thanks!

PR #1448, opened 2023-06-23
Fix GitHub CI build

sgillespie:
# Description

Add `icu4c` to PKG_CONFIG_PATH for macOS

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -60,7 +60,7 @@ jobs:
       run: |
         brew install postgresql@14 libpq openssl@1.1 tree
         tree /usr/local/opt/postgresql@14/lib/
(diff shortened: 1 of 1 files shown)


sgillespie:
> Wonder why it worked then not as nothing was update in this file

No idea. I couldn't reproduce it in my fork build either. The only thing I know is this seemed to fix  it: https://github.com/input-output-hk/cardano-db-sync/actions/runs/5359842716

angerman (approved):
Looks ok. Though I can't shake the feeling that this should rather be in the base action?

sgillespie:
> Looks ok. Though I can't shake the feeling that this should rather be in the base action?

Should it be in the base or haskell action?

sgillespie:
> Looks ok. Though I can't shake the feeling that this should rather be in the base action?

@angerman I added it over here: https://github.com/input-output-hk/actions/pull/1. I initially put it in actions/haskell, we can discuss the proper place there.

Issue #1453, opened 2023-06-29
Remove cardano-api dependency


kderme:
db-sync mainly uses directly the ledger and the ledger-api. However it still depends on cardano-api on some parts. Cardano-api has recently moved to a separate repository https://github.com/input-output-hk/cardano-api and it's questionable how fast it will be upgraded after node is released. We should try to completely drop our dependency to it, so that it doesn't block future node integrations.

rdlrt:
Interesting, I thought node & CLI had strong dependency on cardano-api, so the work on cardano-api would be locked prior to node being tagged

PR #1458, opened 2023-07-11
Remove cardano-api dependency

sgillespie:
# Description

Take steps to remove dependency on cardano-api altogether (#1453):

 * Move all calls to cardano-api calls into a single module
 * Reimplement bech32 serialisation
 * Add some pure property tests

A few notes:

 * I put the functionality into `Cardano.DbSync.Util` mostly because I'm not sure where else to put it
 * I needed to test some pure functions, so I added a Hedgehog testsuite to cardano-db-sync
 * There's no type classing or overloading of any kind

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-db-sync/cardano-db-sync.cabal
```diff
@@ -296,3 +307,48 @@ executable test-http-get-json-metadata
                       , text
                       , transformers
                       , transformers-except
+
+test-suite test
```
Cmdv, line 311: 🎉

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Util.hs
```diff
@@ -4,9 +4,10 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE NoImplicitPrelude #-}
+
 -- Need this because both ghc-8.6.5 and ghc-8.10.2 incorrectly warns about a redundant constraint
 -- in the definition of renderAddress.
-{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+-- {-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
Cmdv, line 10: assume it works now so probably worth removing comments?

Cmdv (commented):
Looks good, just a very minor comment that has been left behind.
Otherwise 🚢

kderme (approved):

PR #1460, opened 2023-07-13
Tests more command line args

Cmdv:
# Description

Add more tests for command line arguments:
`disable-epoch`
`config`
Also rename `*extended` to `*epochDisabled + *epochAndCacheEnabled`

# Checklist

- [ ] Self-reviewed the diff

# Migrations


sgillespie (approved):
LGTM

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/CommandLineArg/ConfigFile.hs
```diff
@@ -0,0 +1,37 @@
...
+where
+
+import Cardano.Mock.ChainSync.Server (IOManager)
+import Data.Text (Text)
+import Test.Cardano.Db.Mock.Config (CommandLineArgs (..), babbageConfigDir, withCustomConfigAndLogs)
+import Test.Tasty.HUnit (Assertion)
+
+mkCommandLineArgs :: Bool -> CommandLineArgs
+mkCommandLineArgs hasConfigFile =
+  CommandLineArgs
```
kderme, line 13: It may be less verbose to have a default `CommandLineArgs` and write here only the values that are different, using a notation of this form
```
defaultCommandLineArgs {claHasConfigFile =  claHasConfigFile, ...} 
```

Cmdv: good spot I've actually got one already too `initCommandLineArgs`!!

kderme (approved):
LGTM! It would be great to have some tests where these flags change on a second start of db-sync.

Cmdv:
@kderme yeah great idea I'll try that out in another PR for sure 👍

Issue #1457, opened 2023-07-11
Wrong tx_count and blk_count numbers in the epoch table


dmitrystas:
**OS**
Your OS: Debian 11

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.1.0.2
PostgreSQL version: 15
 
**Build/Install Method**
The method you use to build or install `cardano-db-sync`: binaries download

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): systemd

**Additional context**
A snapshot (https://update-cardano-mainnet.iohk.io/cardano-db-sync/index.html#13.1/) was used

**Problem Report**
Almost all rows in the table `epoch` have wrong tx_count and blk_count. Usually they differ by 1 block

 ```
SELECT COUNT(*) AS tx_count, COUNT(distinct block.id) AS blk_count
FROM block
LEFT JOIN tx ON tx.block_id = block.id
WHERE block.epoch_no = 386;

  tx_count | blk_count
----------+-----------
    429066 |     20893
(1 row)


SELECT tx_count, blk_count
FROM epoch
WHERE no = 386;

  tx_count | blk_count
----------+-----------
    428477 |     20892
(1 row)
```

but for some epochs it looks even more strange

```
SELECT COUNT(*) AS tx_count, COUNT(distinct block.id) AS blk_count
FROM block
LEFT JOIN tx ON tx.block_id = block.id
WHERE block.epoch_no = 387;

 tx_count | blk_count
----------+-----------
   389880 |     21065
(1 row)


SELECT tx_count, blk_count
FROM epoch
WHERE no = 387;

 tx_count | blk_count
----------+-----------
    1303  |       43
(1 row)

```

Fell-x27:
I have the same problem.

Fell-x27:
Here is a table showing inconsistent data. Everything is accurate up until entry 348, but after that, the data is almost always incorrect.

| no | count in blocks table | count in epochs table | deviation |
| :--- | :--- | :--- | :--- |
| 0 | 21587 | 21587 | 0 |
| 1 | 21590 | 21590 | 0 |
| 2 | 21599 | 21599 | 0 |
| 3 | 21598 | 21598 | 0 |
| 4 | 21601 | 21601 | 0 |
| 5 | 21599 | 21599 | 0 |
| 6 | 21601 | 21601 | 0 |
| 7 | 21600 | 21600 | 0 |
[...]

PR #1468, opened 2023-07-24
doc/interesting-queries.md: Add new query

erikd:
This query finds incorrect epoch table entries.

Related to: https://github.com/input-output-hk/cardano-db-sync/issues/1457


Cmdv (commented):
LGTM, though we should maybe add a caveat that this query is likely to be very slow over a big data set?

erikd:
This query runs in under 10 seconds on a full `mainnet` database.

Cmdv:
yeah cool, as mentioned in the issue I "think" that when my PR is released then this issue shouldn't exist.

Cmdv (approved):
LGTM 👍

Issue #1459, opened 2023-07-12
Replace panic by exceptions


kderme:
This becomes possible after https://github.com/input-output-hk/cardano-db-sync/issues/1438. Before that, simple exceptions were caught by the mini-protocol subscriber and all the threads were restarted. Now an exception at any storage thread is enough to kill db-sync.

Cmdv:
Could we discuss which route we are going to take with errors in general ? There seems to be a mishmash of different ways we do it through out this repo and it would be great if we could align into a single method (if possible).

@kderme @sgillespie what do you both think in regards to how we handle errors?

Cmdv:
Ok I'm going to attempt to use https://github.com/fpco/safe-exceptions as it seems to be used in other project at IOG

kderme:
I wasn't aware of the library, but do we need something more complicated than `Control.Exception` for this? The library is based around `MonadCatch`, `MonadMask` monads, that we don't use in db-sync. Actually from a quick search, more panics are in pure code, where a `error` may suffice.

sgillespie:
> I wasn't aware of the library, but do we need something more complicated than Control.Exception for this? 

AFAIK `Control.Exception` does not have `MonadThrow` (also provided by `exceptions` package), which I think is easier when working with transformers. `exceptions` is also built on top of `Control.Exception`. If we're going for GHC runtime exceptions, I think it's worth looking at.

Cmdv:
@kderme no I'm happy with `Control.Exception`, just trying to have a unified way of dealing with errors. 

Here are some of the various ways we do things:
`Control.Monad.Trans.Except.Exit (orDie)` 
`Control.Monad.Trans.Except.Extra (left, firstExceptT, hoistEither, handleIOExceptT)` 
`GHC.Base (error)`
`Control.Exception`
`throwIO e`
`panic`
`Either Text (..)`
`Either SyncNodeErrors ()`

Error types:
`SyncNodeErrors`
`FetchError`
`PoolOfflineFetchError`

Also lots of
`ExceptT SyncNodeError IO ()`

from reading [this blog post](https://www.fpcomplete.com/haskell/tutorial/exceptions/), it states:
> making large sum types containing all possible exceptions, and wrapping all your IO code in ExceptT, is a major anti-pattern. It has three major downsides:
It slows down your code.
It introduces confusion: your type signatures imply that it has only one way to fail (the explicit error case), but in reality runtime exceptions still exist.
[...]

PR #1471, opened 2023-07-26
Replace panic with exceptions

Cmdv:
# Description

Replacing `panic` with exceptions and better error messages/types.

There are still some panic's left in the project in :
`./cardano-smash-server/`
`./cardano-chain-gen/`

But the all have been replaced in `./cardano-db-sync/`

fixes https://github.com/input-output-hk/cardano-db-sync/issues/1459

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db/Multiplex.hs
```diff
@@ -85,7 +85,7 @@ runExtraMigrations trce blockNoDiff consumed pruned = do
       liftIO $ logInfo trce "Extra migration consumed_tx_out already executed"
     (True, False, False) -> do
       liftIO $ logError trce migratedButNotSet
-      panic migratedButNotSet
+      liftIO $ throwIO $ userError (show migratedButNotSet)
```
kderme, line 88: These lines look a lot like `logAndThrowIO` except it's not IO. See comment on `logAndThrowIO` on how to reuse it more.

cardano-smash-server/src/Cardano/SMASH/Server/Types.hs
```diff
@@ -399,32 +419,42 @@ instance ToJSON DBFail where
...
-renderDBFail (DbInsertError err) =
-  "The database got an error while trying to insert a record. Error: " <> err
-renderDBFail (DbLookupPoolMetadataHash poolId poolMDHash) =
-  "The metadata with hash " <> show poolMDHash <> " for pool " <> show poolId <> " is missing from the DB."
-renderDBFail (TickerAlreadyReserved ticker) =
-  "Ticker name " <> getTickerName ticker <> " is already reserved"
-renderDBFail RecordDoesNotExist =
-  "The requested record does not exist."
-renderDBFail (DBFail lookupFail) = renderLookupFail lookupFail
+-- renderDBFail :: DBFail -> Text
```
kderme, line 450: Forgotten comments

cardano-db-sync/src/Cardano/DbSync/Error.hs
```diff
@@ -66,55 +156,35 @@ renderSyncInvariant ei =
...
     Right txt -> txt
+
+runOrThrowIO :: forall e a. Exception e => IO (Either e a) -> IO a
+runOrThrowIO ioEither = do
+  et <- ioEither
+  case et of
+    Left err -> throwIO err
+    Right a -> pure a
+
+logAndThrowIO :: Trace IO Text -> Text -> IO ()
```
kderme, line 172: This could work for any `MonadIO m`, to allow more usages.

kderme (commented):
Looks good! Just a few comments

cardano-db-sync/src/Cardano/DbSync/Error.hs
```diff
@@ -20,22 +27,105 @@ import Cardano.DbSync.Util
...
+  | SNErrLedgerState !String
+  | SNErrNodeConfig NodeConfigError
+  | SNErrLocalStateQuery !String
+
+instance Exception SyncNodeError
+
+instance Show SyncNodeError where
+  show =
+    \case
+      SNErrDefault t -> "Error SNErrDefault: " <> Text.show t
```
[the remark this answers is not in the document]
Cmdv: Think I was getting clashes between the different `show`s but just making them `show` and also removing `textShow` and using `show` and it all compiles now. Maybe I copy pasted too haphazardly when converting to an instance.

cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
```diff
@@ -208,48 +211,59 @@ applyBlock env blk = do
   atomically $ do
     !ledgerDB <- readStateUnsafe env
     let oldState = ledgerDbCurrent ledgerDB
-    let !result = applyBlk (ExtLedgerCfg (getTopLevelconfigHasLedger env)) blk (clsState oldState)
+    !result <- applyBlk (ExtLedgerCfg (getTopLevelconfigHasLedger env)) blk (clsState oldState)
```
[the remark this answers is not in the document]
Cmdv: ah yes good idea. I added the `STM` so I could `throwSTM` but you're right I can make a utility function and do it there thus keeping `applyBlk` pure. 👍

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Offline/Http.hs
```diff
@@ -31,6 +31,7 @@ import qualified Data.Text.Encoding as Text
 import Network.HTTP.Client (HttpException (..))
 import qualified Network.HTTP.Client as Http
 import qualified Network.HTTP.Types as Http
+import qualified GHC.Show as GHCS
```
sgillespie, line 34: Is this used?

cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -234,7 +234,7 @@ dbSyncProtocols syncEnv metricsSetters tc _version codecs _connectionId =
                     ( Client.chainSyncClientPeer $
                         chainSyncClientFixScripts backend tracer ls
                     )
-              when onlyFix $ panic "All Good! This error is only thrown to exit db-sync" -- TODO fix.
+              when onlyFix $ error "All Good! This error is only thrown to exit db-sync"
```
sgillespie, line 237: I know you didn't introduce this, but I think it would be better to find a way to exit gracefully here.

cardano-db-sync/src/Cardano/DbSync/Util.hs
```diff
@@ -140,29 +134,9 @@ nullMetricSetters =
     , metricsSetDbSlotHeight = const $ pure ()
     }
 
--- The network code catches all execptions and retries them, even exceptions generated by the
--- 'error' or 'panic' function. To actually force the termination of 'db-sync' we therefore
--- need a custom panic function that is guaranteed to abort when we want it to.
--- However, we may not want to abort in production, so we make it optional by use of an
--- environment variable.
```
sgillespie, line 147: Should we also add this as an option to the NixOS module?

sgillespie (commented):
Well done! I've added a few comments and questions

sgillespie:
hlint failure: https://ci.iog.io/build/538220/nixlog/4

cardano-db-sync/src/Cardano/DbSync/Sync.hs
kderme: I think it's impossible to exit gratefully and this is a limitation of either the protocol or the network-client library that we use. 
This is one of the cases where `panic` should remain. Otherwise the connection will keep restarting.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Offline/Http.hs
Cmdv: this is weird when removing it gave me an error on the `show` inside the show instance. I removed the import and used `import GHC.Show (show)`

cardano-db-sync/src/Cardano/DbSync/Util.hs
Cmdv: I was wondering if it should be a flag rather than a env variable. But that would cause a breaking change so I left it.

sgillespie: Why would a flag make it breaking?

sgillespie (approved):
LGTM

kderme (approved):

PR #1474, opened 2023-07-27
fix: Hardcode Byron software version

### LINKED ISSUE #1473: Byron application params no longer used since Shelley still required
**Description**
There are legacy byron parameters `ApplicationName` and `ApplicationVersion` which are no longer in use since Shelley era and these have been hard coded into other cardano products ([example 1](https://github.com/input-output-hk/cardano-node/pull/5240), [example 2](https://github.com/input-output-hk/cardano-api/blob/102e790b68aa56985436bfbd78f1f08af0d0ec67/cardano-api/internal/Cardano/Api/SpecialByron.hs#L205-L213)) and subsequently removed from the general iohk-nix environment descriptions ([link](https://github.com/input-output-hk/iohk-nix/commit/443d36860c7ecc2711101dfa703a55dfcc1b3c9c)).  However, cardano-db-sync still requires these parameters and throws a fatal error without them explicitly defined in the config.

**OS**
NixOS 23.05

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.1.1.3
PostgreSQL version: 12.13

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: nix

**Run method**

**Problem Report**
Not having these parameters present in the config results in this fatal error:
```
FatalError {fatalErrorMessage = "Error parsing node config: AesonException \"Error in $: key \\\"ApplicationName\\\" not found\""}
```
Example CI failure with this message [here](https://ci.iog.io/build/392010/nixlog/39), when iohk-nix environment is used without these parameters present in the config.

Slack discussion [here](https://input-output-rnd.slack.com/archives/CC0CLNEH1/p1690307421615979).

sgillespie:
# Description

Cardano Node no longer provides this as configuration. Fixes #1473 

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-chain-gen/test/testfiles/config-alonzo/test-config.json
+++ b/cardano-chain-gen/test/testfiles/config-alonzo/test-config.json
@@ -1,8 +1,6 @@
 {
   "AlonzoGenesisFile": "genesis.alonzo.json",
   "AlonzoGenesisHash": "7e94a15f55d1e82d10f09203fa1d40f8eede58fd8066542cf6566008068ed874",
-  "ApplicationName": "cardano-sl",
-  "ApplicationVersion": 0,
   "ByronGenesisFile": "genesis.byron.json",
   "ByronGenesisHash": "865a9da19944fc00cefe5a0cdcb6dc9d3964a74681d9488ab79967f13a6400b3",
   "LastKnownBlockVersion-Alt": 0,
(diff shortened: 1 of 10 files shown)


Cmdv (approved):
LGTM, though it looks like we use version `0` in the `.json` files used for the tests (found them by searching for `ApplicationVersion` in `cardano-chain-gen/`)

sgillespie:
> LGTM, though it looks like we use version `0` in the `.json` files used for the tests (found them by searching for `ApplicationVersion` in `cardano-chain-gen/`)

Nice find! I've removed those from the json files as well

Cmdv (approved):
👌

PR #1476, opened 2023-08-05
Update dependencied to Node-8.2.0-pre

kderme:
# Description

Allows db-sync to follow sanchonet.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-chain-gen/src/Cardano/Mock/Forging/Tx/Babbage.hs
```diff
@@ -643,13 +637,13 @@ mkFullTx n m sta = do
...
 
     certs =
-      [ DCertDeleg $ RegKey $ Prelude.head unregisteredStakeCredentials
-      , DCertPool $ RegPool poolParams0
-      , DCertPool $ RegPool poolParams1
-      , DCertPool $ RetirePool (Prelude.head unregisteredPools) (EpochNo 0)
-      , DCertDeleg $ DeRegKey $ unregisteredStakeCredentials !! 2
-      , DCertDeleg $ Delegate $ Delegation (unregisteredStakeCredentials !! 1) (unregisteredPools !! 2)
-      , DCertMir $
+      [ ShelleyTxCertDelegCert $ ShelleyRegCert $ Prelude.head unregisteredStakeCredentials
```
Cmdv, line 639: Just checking I assume there will never be a scenario where the list is empty in the tests, since there is the use of `head`?

cardano-db-sync/src/Cardano/DbSync/Config/Cardano.hs
```diff
@@ -131,7 +133,7 @@ mkProtocolInfoCardano ge shelleyCred =
         (Consensus.ProtocolTransitionParamsShelleyBased () $ dncMaryHardFork dnc)
         (Consensus.ProtocolTransitionParamsShelleyBased alonzoGenesis $ dncAlonzoHardFork dnc)
         (Consensus.ProtocolTransitionParamsShelleyBased () $ dncBabbageHardFork dnc)
-        (Consensus.ProtocolTransitionParamsShelleyBased (ConwayGenesis (GenDelegs mempty)) Consensus.TriggerHardForkNever) -- TODO: Conway Fix
+        (Consensus.ProtocolTransitionParamsShelleyBased (ConwayGenesis (GenDelegs mempty)) (Consensus.TriggerHardForkAtEpoch (EpochNo 34))) -- TODO: Conway Fix
```
Cmdv, line 136: maybe a little note as to why hard fork at epoch 34?

Cmdv (approved):
LGTM 🎉 Just a couple of basic questions

cardano-db-sync/src/Cardano/DbSync/Config/Cardano.hs
kderme: Yes this needs to change. It's hardcoded 34 for sanchonet, but needs to be parsed from the config file before this is merged.

Cmdv: Cool thanks, so it's usually an option in the config file

sgillespie:
LGTM, except for the CI failures. I believe this is related to the updated dependencies and not the changes introduced here.

sgillespie:
It looks like there were some hlint failures: https://ci.iog.io/build/567612/nixlog/89

PR #1479, opened 2023-08-10
Make it build with ghc-9.6

erikd:
# Checklist

- [x] Self-reviewed the diff

# Migrations

No breaking change


angerman (dismissed):
Lgtm

kderme (dismissed):

sgillespie:
One other thing I just realized is we'll probably want to add a variant for ghc962 here: https://github.com/input-output-hk/cardano-db-sync/blob/master/nix/haskell.nix#L60

PR #1481, opened 2023-08-11
Fix Hydra failures

sgillespie:
# Description

This change fixes the CI build failures, and [for now] removes `ghc962` from nix builds. We will continue to work to get GHC 9.6.2 supported.

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-chain-gen/cardano-chain-gen.cabal
+++ b/cardano-chain-gen/cardano-chain-gen.cabal
@@ -30,12 +30,14 @@ library
   hs-source-dirs:       src
 
   ghc-options:          -Wall
+                        -Werror
                         -Wcompat
-                        -fwarn-redundant-constraints
-                        -fwarn-incomplete-patterns
-                        -fwarn-unused-imports
+                        -Wredundant-constraints
+                        -Wincomplete-patterns
                         -Wincomplete-record-updates
                         -Wincomplete-uni-patterns
+                        -Wunused-imports
+                        -Wunused-packages
 
   exposed-modules:
                         Cardano.Mock.Chain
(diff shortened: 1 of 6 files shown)


cardano-chain-gen/cardano-chain-gen.cabal
```diff
@@ -108,29 +90,13 @@ library
...
                       , strict-stm
-                      , swagger2
                       , text
-                      , time
                       , transformers
-                      , transformers-except
                       , typed-protocols
-                      , unix
-                      , vector
-                      , yaml
```
kderme, line 133: I hadn't realised this has so many unused packages

Cmdv (approved):
🎉

nix/haskell.nix
```diff
@@ -56,8 +56,8 @@ let
     src = ../.;
     # our current release compiler is 8107
     compiler-nix-name = lib.mkDefault "ghc8107";
-    # but we also build for 962.
-    flake.variants = lib.genAttrs ["ghc962"] (x: {compiler-nix-name = x;});
+    # but we also build for 927.
+    flake.variants = lib.genAttrs ["ghc927"] (x: {compiler-nix-name = x;});
```
angerman, line 60: Can't we have `["ghc927" "ghc962"]` here? I'm a bit concerned that we'd regress on all the changes @erikd already made?

angerman (approved):
Conceptually ok to make CI green, but I'm a bit concerned we should have 9.6.2 green as well.

nix/haskell.nix
sgillespie: We're stuck in an awkward place on this one for the time being.  Postgres won't build with musl since this commit: https://github.com/NixOS/nixpkgs/commit/8d4a81b1851be7823e23d26bd02abcfb902d2bef, or at least any version < 14, which means we can't update nixpkgs-unstable. The revision of nixpkgs we're using does not contain ghc962 (and I'm having trouble getting ghc961 to build).

I've submitted an issue and PR to nixpkgs, hopefully someone will look at it. I'm tracking everything here: https://github.com/input-output-hk/cardano-db-sync/issues/1483

sgillespie: @angerman nixpkgs-unstable should be fixed now, how often does haskell.nix update?

angerman: @sgillespie whenever @hamishmack bumps it 🙈

PR #1482, opened 2023-08-13
Update doc/schema.md to fix description of tx_id in datum and redeemer

WubbleWobble:
doc/schema.md previously stated for both `datum` and `redeemer_data` tables that `tx_id` was "The Tx table index for the transaction where this script first became available.", neither of which seems to make sense. Rather the `tx_id` fields indicate the first transaction in which the pieces of data (datum / redeemer_data) were used.


doc/schema.md
```diff
@@ -564,7 +564,7 @@ A table containing Plutus Datum, found in witnesses or inlined in outputs
 |-|-|-|
 | `id` | integer (64) |  |
 | `hash` | hash32type | The Hash of the Datum |
-| `tx_id` | integer (64) | The Tx table index for the transaction where this script first became available. |
```
kderme, line 567: Indeed this should say `datum` instead of `script`. However I'm not sure if `was first used` is correct. At least in my mind a datum is used when it runs inside a script. But an inline datum is not used when it becomes available on chain.

WubbleWobble: That makes sense (to my limited understanding - still very much learning!).

Maybe `where this datum was first observed` (or "mentioned", or "seen") then?

Cmdv (approved):

Cmdv:
@kderme is this good to go or do we also need review from docs-access ?

Issue #1325, opened 2022-12-14
Use Deposits ledger event


kderme:
With this event we can reduce the queries that db-sync does during syncing
pr for ledger is https://github.com/input-output-hk/cardano-ledger/pull/3212

kderme:
During integration of the ledger, it was realised that the event is missing from most eras, only appeared in Shelley. This pr adds it to other eras https://github.com/input-output-hk/cardano-ledger/pull/3404. This issue is blocked until the event extension is merged and released from ledger

PR #1484, opened 2023-08-16
Stake Distribution, Deposits and Conway

kderme:
# Description

This pr combines 3 different stuff :
- dbsync no longer queries the TxOut table for every input. This was currently used only to compute the deposits of the tx, which are now taken through a ledger event. Fixes https://github.com/input-output-hk/cardano-db-sync/issues/1325
- stake-distr is computed earlier. Fixes https://github.com/input-output-hk/cardano-db-sync/issues/1406
- initial Conway integration

These are combined in a single pr, since testing on master any of these 3 tickets is necessary and it's more efficient to test them at once.

# Checklist

- [x] Any changes are noted in the [changelog](https://github.com/input-output-hk/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)
- [ ] Self-reviewed the diff

# Migrations

- [x] The pr causes a [breaking change](https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/migrations.md) of type a,b or c

This pr adds migrations files and is a b type breaking change. 
In addition an adjusting procedure is created for the stake distribution changes. On an upgrade this procedure will insert the missing entries that should have been inserted earlier.


cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/CommandLineArg/ForceIndex.hs
```diff
@@ -20,7 +20,7 @@ checkForceIndexesArg =
     startDBSync dbSyncEnv
     -- assertBlockNoBackoff dbSyncEnv 0
     threadDelay 3_000_000
-    assertEqQuery dbSyncEnv DB.queryPgIndexesCount 143 "there wasn't the correct number of indexes"
+    assertEqQuery dbSyncEnv DB.queryPgIndexesCount 161 "there wasn't the correct number of indexes"
```
Cmdv, line 23: these were failing on me today too (they were out by 1) but if I turned on consumed + purge to True they would pass 🤷

kderme: There is an additional index when this flag is used, so this is expected
```
CREATE INDEX IF NOT EXISTS idx_tx_out_consumed_by_tx_in_id ON tx_out (consumed_by_tx_in_id)
```

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -541,21 +628,21 @@ insertConwayDelegCert ::
   SlotNo ->
   ConwayDelegCert StandardCrypto ->
   ExceptT SyncNodeError (ReaderT SqlBackend m) ()
-insertConwayDelegCert tracer cache network txId idx mRedeemerId epochNo slotNo dCert =
+insertConwayDelegCert _tracer cache network txId idx mRedeemerId epochNo slotNo dCert =
```
Cmdv, line 631: maybe we can remove `_tracer` if it's not being used. 

totally not related to this PR and would probably require way too much work but it would be nice to have have an app wide monad transformer with all our heavily used configs eg: `Trace IO Text + SyncNodeParams` etc...

cardano-db-sync/src/Cardano/DbSync/Fix/PlutusScripts.hs
```diff
@@ -147,7 +147,7 @@ scrapScriptBlock cblk = case cblk of
   BlockShelley _ -> error "No Plutus Scripts in Shelley"
   BlockAllegra _ -> error "No Plutus Scripts in Allegra"
   BlockMary _ -> error "No Plutus Scripts in Mary"
-  _ -> mempty -- error "TODO: Conway not supported"
+  _ -> mempty -- This bug existed in a version that didn't support Conway
```
Cmdv, line 150: is it worth being explicit? by having both:
```
BlockConway _ -> mempty
_ -> mempty -- This bug existed in a version that didn't support Conway
```

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -250,19 +253,36 @@ insertTx ::
...
       !withdrawalSum = fromIntegral $ unCoin $ Generic.txWithdrawalSum tx
-  !resolvedInputs <- mapM (resolveTxInputs hasConsumed (fst <$> groupedTxOut blockGroupData)) (Generic.txInputs tx)
-  let !inSum = sum $ map (unDbLovelace . forth4) resolvedInputs
-  let diffSum = if inSum >= outSum then inSum - outSum else 0
-  let !fees = maybe diffSum (fromIntegral . unCoin) (Generic.txFees tx)
-  let !txHash = Generic.txHash tx
+  hasConsumed <- liftIO $ getHasConsumed syncEnv
+  (resolvedInputs, fees', deposits) <- case (mdeposits, unCoin <$> Generic.txFees tx) of
+    (Just deposits, Just fees) -> do
+      (resolvedInputs, _) <- splitLast <$> mapM (resolveTxInputs hasConsumed False (fst <$> groupedTxOut grouped)) (Generic.txInputs tx)
```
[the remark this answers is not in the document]
Cmdv: oh no take it back just seen it later in the review, is it worth having explicit import for it?

Cmdv (commented):
LGTM just a couple of questions 👍

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/StakeDist.hs
```diff
@@ -73,44 +73,50 @@ getSecurityParameter = maxRollbacks . configSecurityParam . pInfoConfig
...
+  | index + size > delegationsLen = Slice (mkSlice (delegationsLen - index)) True
+  | otherwise = Slice (mkSlice size) False
   where
-    -- We use 'ssStakeSet' here instead of 'ssStateMark' because the stake addresses for the
-    -- later may not have been added to the database yet. That means that when these values
+    epoch :: EpochNo
+    epoch = 1 + Shelley.nesEL (Consensus.shelleyLedgerState lstate)
+
+    minSliceSize :: Word64
+    minSliceSize = 2000
```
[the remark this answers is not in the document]
kderme: We don't want to allow very low values here, so it may not be safe to expose it to users.

cardano-db-sync/src/Cardano/DbSync/Fix/PlutusScripts.hs
kderme: I don't think this would work, as the _ would raise an Unreachable case warning. The current approach is also future proof, maybe the comment can also become 
```suggestion
  _ -> mempty -- This bug existed in a version that didn't support Conway or later eras
```

Cmdv: That works for me. I was unsure without going to the library what other "Block" types there were, thanks for explaining.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/StakeDist.hs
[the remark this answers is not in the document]
Cmdv: ah cool, I get scared 😱 when I see hardcoded values as I spotted it's also hardcoded in the test.

Cmdv (approved):

PR #1493, opened 2023-08-25
Refactor: Nix flake cleanup

sgillespie:
# Description

Clean up Nix code. In particular, I've moved the `haskell.nix` stuff into `flake.nix`, and removed most of the code that's unused or overly complex.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


angerman:
@sgillespie what's up with the schema-rollback change in the cabal file and the related haskell code?

flake.nix
```diff
@@ -3,19 +3,20 @@
...
-    utils.url = "github:numtide/flake-utils";
+    hackageNix = {
+      url = "github:input-output-hk/hackage.nix";
+      flake = false;
+    };
     iohkNix = {
       url = "github:input-output-hk/iohk-nix";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+    devxShell.url = "github:input-output-hk/devx";
```
angerman, line 19: I'm still unsure if we want to _embed_ this or tell people to just go directly `nix develop github:input-output-hk/devx#...`?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
+                      LC_CTYPE = "C";
+                    }))
+                    .override {
+                      enableSystemd = stdenv.hostPlatform.isLinux && !stdenv.hostPlatform.isMusl;
+                      gssSupport = stdenv.hostPlatform.isLinux && !stdenv.hostPlatform.isMusl;
+                    };
+                })
 
-        project = pkgs.cardanoDbSyncProject;
+                (_: cardanoNodeOverlay)
```
angerman, line 62: should we add a comment that says what the cardanoNodeOverlay adds?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
 
-        hlint = pkgs.callPackage pkgs.hlintCheck {
-          inherit (project.args) src;
-        };
+                (final: prev: {
+                  # The NixOS module expects these to be here
```
angerman, line 65:
```suggestion
                  # The NixOS module for cardano-db-sync expects these to be here
```
I assume?

Can we add a pointer where the module is defined?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
+          cardanoNodeOverlay = pkgs:
+            let
+              nodeFlakeCompat = import inputs.flake-compat {
+                inherit (project.hsPkgs.cardano-node) src;
+                inherit pkgs;
+              };
+            in {
+              inherit (nodeFlakeCompat.defaultNix.packages.${system})
+                cardano-cli cardano-node;
+            };
```
angerman, line 83: Can we reuse any of the https://github.com/input-output-hk/cardano-parts?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
-        checks = project.flake'.checks // {
-          inherit hlint fourmolu shellcheck;
-        } // lib.optionalAttrs hostPlatform.isLinux (prefixNamesWith "nixosTests/" nixosTests);
+          project = (nixpkgs.haskell-nix.cabalProject' ({ config, lib, ... }: rec {
+            src = ./.;
+            name = "cardano-db-sync";
+            compiler-nix-name = lib.mkDefault "ghc8107";
+            flake.variants =
+              lib.genAttrs
+                ["ghc962"]
```
angerman, line 134: I guess, we can bump this to ghc963 now 😅

flake.nix
```diff
@@ -24,211 +25,299 @@
...
+            inputMap = {
+              "https://input-output-hk.github.io/cardano-haskell-packages" = inputs.CHaP;
+            };
 
-        scripts = flattenTree pkgs.scripts;
+            shell.buildInputs = with nixpkgs.pkgsBuildBuild; [
+              haskell-language-server
+              gitAndTools.git
+            ];
+            shell.withHoogle = true;
```
angerman, line 145: this is a different shell from the devx shell, do we need to add devx as a flake dependency?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
+              ({ lib, pkgs, ... }: {
+                # Ignore version bounds
+                packages.katip.doExactConfig = true;
+                # Split data to reduce closure size
+                packages.ekg.components.library.enableSeparateDataOutput = true;
+                # Use our libsodium, secp256k1, libblst forks
+                packages.cardano-crypto-praos.components.library.pkgconfig =
+                  lib.mkForce [[ pkgs.libsodium-vrf ]];
+                packages.cardano-crypto-class.components.library.pkgconfig =
+                  with pkgs; lib.mkForce [[ libsodium-vrf secp256k1 libblst ]];
```
angerman, line 157: I thought we had most of this resolved with the crypto-overlays these days? Is this still necessary?

flake.nix
```diff
@@ -24,211 +25,299 @@
...
+              default = flake.packages."cardano-db-sync:exe:cardano-db-sync";
+            } // lib.optionalAttrs (system == "x86_64-darwin") {
+              inherit cardano-db-sync-macos;
+            };
+
+            devShells = with inputs.devxShell.devShells."${system}"; {
+              default = ghc8107-iog;
+              ghc8107 = flake.devShells.default; # Restore previous default
+              devx = ghc8107-iog;
+              devx-ghc962 = ghc962-iog;
```
angerman, line 314:
```suggestion
              devx-ghc96 = ghc96-iog;
```
I think this will work, there are aliases now for the latest minor version to each major version.

angerman (commented):
I've given this as good as I could a review. This is great work! just a few nits!

(Also probably does need a rebase 😓 now; sorry for the delay)

sgillespie:
> @sgillespie what's up with the schema-rollback change in the cabal file and the related haskell code?

Great question! The schema-rollback tests no longer have any effect. Those tests will succeed whether there are missing rollbacks or not. They attempt to parse TH definitions, but that has been significantly refactored since then.

flake.nix
sgillespie: You're right, this is no longer necessary

flake.nix
sgillespie: OK, I thought this would make it easier, but then we'll also have to keep it up-to-date. I'll remove this for the time being and we can always add it back if we change our minds

angerman:
@sgillespie we should note this in the description of this PR. Right now it talks only about nix changes, making the Haskell changes look out of place. Even better might be to split that out into a separate PR though.

sgillespie:
> @sgillespie what's up with the schema-rollback change in the cabal file and the related haskell code?

For future reviewers' benefit, I took this out (including the history), so you won't see this. I'll address this in another PR

kderme (dismissed):
LGTM!

Cmdv (approved):

nix/docker.nix
```diff
@@ -1,159 +1,179 @@
...
-      jq # JSON processor
-      iana-etc # IANA protocol and port number assignments
-      iproute # Utilities for controlling TCP/IP networking
-      iputils # Useful utilities for Linux networking
-      libidn # Library for internationalized domain names
-      libpqxx # A C++ library to access PostgreSQL databases
-      postgresql # A powerful, open source object-relational database system
-      socat # Utility for bidirectional data transfer
-      utillinux # System utilities for Linux
-      cardano-cli
```
angerman, line 85: I actually like these comments here quite a lot. Was there a reason fro dropping them @sgillespie?

angerman (approved):
LGTM. Where do the 1.1k extra lines come from? flake.lock?

nix/docker.nix
sgillespie: I thought the majority of them were superfluous. eg, I think it's obvious that `gnutar` provides GNU tar. I don't have a problem adding them back

sgillespie:
> LGTM. Where do the 1.1k extra lines come from? flake.lock?

Yes, the majority came from flake.lock:
```
 .github/workflows/release-binaries.yml |   25 +++
 flake.lock                             | 2451 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------------------------------------------------------
 flake.nix                              |  450 ++++++++++++++++++++++++++------------------
 nix/binary-release.nix                 |   39 ----
 nix/custom-config.nix                  |    4 -
 nix/docker.nix                         |  295 +++++++++++++++--------------
 nix/flake-compat.nix                   |   20 --
 nix/haskell.nix                        |  195 -------------------
 nix/pkgs.nix                           |   51 -----
 nix/scripts.nix                        |   31 ---
 nix/util.nix                           |    9 -
 11 files changed, 2326 insertions(+), 1244 deletions(-)
```

angerman (approved):
🎉

nix/docker.nix
angerman: I agree in the case of gnu tar, that that's not _that_ helpful. Same for jq, postgreql, but the other might be a bit trickier to know. Having that documentation isn't bad. If it would say _why_ we need them it would probably be even better, but let's not review this PR to death 🙄

PR #1502, opened 2023-09-08
Update GitHub CI to use actions/devx

sgillespie:
# Description

Why [input-output-hk/actions/devx-action](https://github.com/input-output-hk/actions#devx-action)? The devx action provides all the system dependencies with nix, but haskell dependencies from cabal. This makes builds easily reproducible from both CI and local development.

This means that if anything goes wrong in CI, you can enter the devx shell, then run the same commands CI runs, and the result should be the same.

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -1,143 +1,92 @@
 name: Haskell CI
 
-on: [push]
+on:
+  push:
+  pull_request:
+  schedule:
+    # Everyday at 4:00 AM UTC
+    - cron: "0 4 * * *"
 
 jobs:
   build:
+    name: build
+    defaults:
+      run:
+        shell: devx {0}
     runs-on: ${{ matrix.os }}
 
-
-    env:
-      # Modify this value to "invalidate" the cabal cache.
-      CABAL_CACHE_VERSION: "2023-05-05"
-
     strategy:
-      fail-fast: false
       matrix:
-        cabal: ["3.10.1.0"]
-        ghc: ["8.10.7", "9.2.7", "9.6.2" ]
         os: [ubuntu-latest, macos-latest]
+        compiler-nix-name: [ghc8107]
 
     steps:
-    - name: Install Haskell
-      uses: input-output-hk/actions/haskell@latest
-      id: setup-haskell
-      with:
-        ghc-version: ${{ matrix.ghc }}
-        cabal-version: ${{ matrix.cabal }}
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install GHC and Cabal
+        uses: input-output-hk/actions/devx@latest
+        with:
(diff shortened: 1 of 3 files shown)


Cmdv (commented):
LGTM 🎉 
one little question will I still be ok to run ghc 9.6.2 locally?

Cmdv (approved):

sgillespie:
> LGTM 🎉 one little question will I still be ok to run ghc 9.6.2 locally?

Good point, I need to add this back to the matrix

kderme (approved):

.github/workflows/haskell.yml
```diff
@@ -1,143 +1,82 @@
 name: Haskell CI
 
-on: [push]
+on:
+  push:
+  pull_request:
+  schedule:
+    # Everyday at 4:00 AM
```
angerman, line 7: Do we know the TZ? is this UTC?

.github/workflows/haskell.yml
```diff
@@ -1,143 +1,82 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install GHC and Cabal
+        uses: input-output-hk/actions/devx@latest
+        with:
+          platform: ${{ matrix.os == 'ubuntu-latest' && 'x86_64-linux' || 'x86_64-darwin' }}
+          target-platform: ""
+          compiler-nix-name: ${{ matrix.compiler-nix-name }}
+          minimal: false
+          iog: true
```
angerman, line 34:
```suggestion
          # enable IOG flavour to bring in all the crypto libraries we need.
          iog: true
```

angerman (approved):
Just a few ideas. Should also `shellcheck` `scripts/postgresql-test.sh` in ci?

.github/workflows/haskell.yml
sgillespie: Great question! It appears to be [UTC](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule). Is there a better time?

sgillespie:
> Just a few ideas. Should also `shellcheck` `scripts/postgresql-test.sh` in ci?

There are a lot of violations in all of our other shell scripts, so I'll address that in another PR. In the meantime, I've fixed all the violations in `postgres-test.sh`

.github/workflows/haskell.yml
angerman: Let's just document that it's 4am utc :-) 4 am is such an ambiguous value without a TZ.

sgillespie: Done

sgillespie:
> Should also `shellcheck` `scripts/postgresql-test.sh` in ci?

https://github.com/input-output-hk/cardano-db-sync/pull/1508

Cmdv (approved):

angerman (approved):
Thank you! This looks good :D So much deleted code 🥳

Issue #1512, opened 2023-09-21
blst not installing on macOs


Cmdv:
macOs is having issues installing `blts` when not on nix, docs need updating with fix.

PR #1513, opened 2023-09-21
Install blst on macos doc update

Cmdv:
# Description

This fixes https://github.com/input-output-hk/cardano-db-sync/issues/1512

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -16,6 +16,10 @@ In addition, Cardano DB Sync requires the following software (instructions below
  * [libsodium-vrf](https://github.com/input-output-hk/libsodium)
  * [secp256k1](https://github.com/bitcoin-core/secp256k1)
  * [blst](https://github.com/supranational/blst)
+ 
+ macOS:
+ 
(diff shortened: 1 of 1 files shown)


doc/installing.md
```diff
@@ -158,6 +158,26 @@ Build it:
...
+./src/vect.h:402:11: fatal error: 'stdlib.h' file not found
+# include <stdlib.h>
+```
+
+check you have x-code command line tools installed or install them by running:
+```bash
+xcode-select --install
+```
+
+if the error still arrises copy the content that x-code installs into `/usr/local/include/` making sure the folder is also on you `$PATH`:
```
sgillespie, line 176: It _might_ make it easier to follow if there was a subsection for macOS here.

sgillespie (changes requested):
Looks good! I added a few suggestions. In particular, I'm asking you to update prerequisites

sgillespie (approved):

PR #1518, opened 2023-09-25
Fix typo in rewards type in schema.md

gildhoff:
The reward type is 'refund' and not 'refunds'. It's just a simple type fix.


Cmdv (approved):
Good spot, thanks 😃

PR #1519, opened 2023-09-25
Speed up prune migrations

### LINKED ISSUE #1511: Speed up the prune migration
Let's say we run the `prune` migration on a db which is close to being fully synced. The prune migration initially runs the consumed migrations, which marks all consumed `tx_out` and then it deletes all the old enough `tx_out`. It could become faster by directly deleting the old enough consumed `tx_out`, instead of doint it in 2 steps.
So instead of running in `runExtraMigrations`
```
migrateTxOut
deleteConsumedTxOut
```

These can be run as
```
deleteConsumedTxOut' -- different from the above
migrateTxOut
```

Cmdv:
# Description

This fixes #1511 

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
+++ b/cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
@@ -174,6 +174,10 @@ updateTxOutConsumedByTxInIdUnique :: MonadIO m => TxId -> Word64 -> TxInId -> Re
 updateTxOutConsumedByTxInIdUnique txOutId index txInId =
   updateWhere [TxOutTxId ==. txOutId, TxOutIndex ==. index] [TxOutConsumedByTxInId =. Just txInId]
 
+deleteTxOutConsumed :: MonadIO m => TxId -> Word64 -> ReaderT SqlBackend m ()
+deleteTxOutConsumed txOutId index =
+  deleteWhere [TxOutTxId ==. txOutId, TxOutIndex ==. index]
+
 getInputPage :: MonadIO m => Word64 -> Word64 -> ReaderT SqlBackend m [(TxInId, TxId, Word64)]
 getInputPage offs pgSize = do
   res <- select $ do
@@ -202,40 +206,67 @@ countConsumed = do
     pure countRows
   pure $ maybe 0 unValue (listToMaybe res)
 
+deleteAndUpdateConsumedTxOut :: MonadIO m => Trace IO Text -> Word64 -> ReaderT SqlBackend m ()
+deleteAndUpdateConsumedTxOut trce blockNoDiff = do
+  maxTxInId <- findMaxTxInId blockNoDiff
+  case maxTxInId of
+    Left errMsg -> liftIO $ logInfo trce $ "No tx_out was deleted: " <> errMsg
+    Right mxtxid -> do
+      migrateNextPage mxtxid 0
+  where
+    migrateNextPage :: MonadIO m => TxInId -> Word64 -> ReaderT SqlBackend m ()
+    migrateNextPage mxTxInId offst = do
+      page <- getInputPage offst pageSize
+      -- before page txinId delete after create extra collumn and update
+      mapM_ (handlePage mxTxInId) page
+      when (fromIntegral (length page) == pageSize) $
+        migrateNextPage mxTxInId $!
+          offst
+            + pageSize
+
+handlePage :: MonadIO m => TxInId -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m ()
+handlePage maxTxInId (txInId, txId, index)
+  | txInId <= maxTxInId = deleteTxOutConsumed txId index
+  | otherwise = do
+      createConsumedTxOut
+      updateTxOutConsumedByTxInIdUnique txId index txInId
+
 deleteConsumedTxOut :: forall m. MonadIO m => Trace IO Text -> Word64 -> ReaderT SqlBackend m ()
 deleteConsumedTxOut trce blockNoDiff = do
-  mBlockHeight <- queryBlockHeight
-  maybe (logNoDelete "No blocks found") deleteConsumed mBlockHeight
-  where
-    logNoDelete txt = liftIO $ logInfo trce $ "No tx_out was deleted: " <> txt
+  maxTxInId <- findMaxTxInId blockNoDiff
+  case maxTxInId of
+    Left errMsg -> liftIO $ logInfo trce $ "No tx_out was deleted: " <> errMsg
+    Right mxtid -> deleteConsumedBeforeTxIn trce mxtid
 
-    deleteConsumed :: Word64 -> ReaderT SqlBackend m ()
+findMaxTxInId :: forall m. MonadIO m => Word64 -> ReaderT SqlBackend m (Either Text TxInId)
+findMaxTxInId blockNoDiff = do
+  mBlockHeight <- queryBlockHeight
+  maybe (pure $ Left $ "No blocks found") deleteConsumed mBlockHeight
+  where
+    deleteConsumed :: Word64 -> ReaderT SqlBackend m (Either Text TxInId)
     deleteConsumed tipBlockNo = do
       if tipBlockNo <= blockNoDiff
-        then logNoDelete $ "Tip blockNo is " <> textShow tipBlockNo
+        then pure $ Left $ "Tip blockNo is " <> textShow tipBlockNo
         else do
           mBlockId <- queryBlockNo $ tipBlockNo - blockNoDiff
           maybe
-            (liftIO $ logError trce $ "BlockNo hole found at " <> textShow (tipBlockNo - blockNoDiff))
+            (pure $ Left $ "BlockNo hole found at " <> textShow (tipBlockNo - blockNoDiff))
             deleteConsumedBeforeBlock
             mBlockId
 
-    deleteConsumedBeforeBlock :: BlockId -> ReaderT SqlBackend m ()
+    deleteConsumedBeforeBlock :: BlockId -> ReaderT SqlBackend m (Either Text TxInId)
     deleteConsumedBeforeBlock blockId = do
       mTxId <- queryMaxRefId TxBlockId blockId False
       case mTxId of
-        Nothing -> logNoDelete $ "No txs found before " <> textShow blockId
+        Nothing -> pure $ Left $ "No txs found before " <> textShow blockId
         Just txId -> do
           mTxInId <- queryMaxRefId TxInTxInId txId True
-          maybe
-            (logNoDelete $ "No tx_in found before or at " <> textShow txId)
-            deleteConsumedBeforeTxIn
-            mTxInId
+          pure $ maybe (Left $ "No tx_in found before or at " <> textShow txId) Right mTxInId
 
-    deleteConsumedBeforeTxIn :: TxInId -> ReaderT SqlBackend m ()
-    deleteConsumedBeforeTxIn txInId = do
-      countDeleted <- deleteWhereCount [TxOutConsumedByTxInId <=. Just txInId]
-      liftIO $ logInfo trce $ "Deleted " <> textShow countDeleted <> " tx_out"
+deleteConsumedBeforeTxIn :: MonadIO m => Trace IO Text -> TxInId -> ReaderT SqlBackend m ()
+deleteConsumedBeforeTxIn trce txInId = do
+  countDeleted <- deleteWhereCount [TxOutConsumedByTxInId <=. Just txInId]
+  liftIO $ logInfo trce $ "Deleted " <> textShow countDeleted <> " tx_out"
 
 queryBlockNo :: MonadIO m => Word64 -> ReaderT SqlBackend m (Maybe BlockId)
 queryBlockNo blkNo = do
--- a/cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
+++ b/cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
@@ -8,11 +8,13 @@
 module Cardano.Db.Migration.Extra.CosnumedTxOut.Queries where
 
 import Cardano.BM.Trace (Trace, logError, logInfo, logWarning)
+import Cardano.Db.Error (LookupFail (..))
 import Cardano.Db.Insert (insertMany', insertUnchecked)
 import Cardano.Db.Migration.Extra.CosnumedTxOut.Schema
 import Cardano.Db.Query (isJust, listToMaybe, queryBlockHeight, queryMaxRefId)
 import Cardano.Db.Text
-import Control.Monad.Extra (foldM, unless, when, whenJust)
+import Control.Exception.Lifted (handle, throwIO)
+import Control.Monad.Extra (unless, when, whenJust)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Trans.Control (MonadBaseControl)
 import Control.Monad.Trans.Reader (ReaderT)
@@ -24,6 +26,7 @@ import qualified Database.Esqueleto.Experimental as Experimental
 import Database.Persist ((<=.), (=.), (==.))
 import Database.Persist.Class (update)
 import Database.Persist.Sql (deleteWhereCount)
+import Database.PostgreSQL.Simple (SqlError)
 
 insertTxOutExtra :: (MonadBaseControl IO m, MonadIO m) => TxOut -> ReaderT SqlBackend m TxOutId
 insertTxOutExtra = insertUnchecked "TxOutExtra"
@@ -70,9 +73,14 @@ setNullTxOutConsumedAfterTxInId :: MonadIO m => TxOutId -> ReaderT SqlBackend m 
 setNullTxOutConsumedAfterTxInId txOutId = do
   update txOutId [TxOutConsumedByTxInId =. Nothing]
 
-migrateTxOut :: MonadIO m => Maybe (Trace IO Text) -> ReaderT SqlBackend m ()
+migrateTxOut ::
+  ( MonadBaseControl IO m
+  , MonadIO m
+  ) =>
+  Maybe (Trace IO Text) ->
+  ReaderT SqlBackend m ()
 migrateTxOut mTrace = do
-  createConsumedTxOut
+  _ <- createConsumedTxOut
   migrateNextPage 0
   where
     migrateNextPage :: MonadIO m => Word64 -> ReaderT SqlBackend m ()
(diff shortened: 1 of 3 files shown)


cardano-db/src/Cardano/Db/Multiplex.hs
```diff
@@ -105,14 +105,15 @@ runExtraMigrations trce blockNoDiff PruneConsumeMigration {..} = do
...
+          ExtraCons.deleteAndUpdateConsumedTxOut trce blockNoDiff
+
           liftIO $ logInfo trce "Running extra migrations consumed_tx_out and prune tx_out"
           ExtraCons.migrateTxOut (Just trce)
-          liftIO $ logInfo trce "Now Running extra migration prune tx_out"
-          ExtraCons.deleteConsumedTxOut trce blockNoDiff
         (True, _, True) -> do
           shouldInsertToMigrationTable
           liftIO $ logInfo trce "Running extra migration prune tx_out"
-          ExtraCons.deleteConsumedTxOut trce blockNoDiff
```
kderme, line 115: I think this shouldn't change.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,67 @@ countConsumed = do
...
+      when (fromIntegral (length page) == pageSize) $
+        migrateNextPage mxTxInId $!
+          offst
+            + pageSize
+
+handlePage :: MonadIO m => TxInId -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m ()
+handlePage maxTxInId (txInId, txId, index)
+  | txInId <= maxTxInId = deleteTxOutConsumed txId index
+  | otherwise = do
+      createConsumedTxOut
```
kderme, line 231: I think this can be called more than once, ie for consecutive pages after `maxTxInId`, possibly causing issues.

cardano-db/src/Cardano/Db/Multiplex.hs
```diff
@@ -105,10 +105,11 @@ runExtraMigrations trce blockNoDiff PruneConsumeMigration {..} = do
           ExtraCons.migrateTxOut (Just trce)
         (False, _, True) -> do
           shouldInsertToMigrationTable
+          liftIO $ logInfo trce "Now Running extra migration prune tx_out"
+          ExtraCons.deleteAndUpdateConsumedTxOut trce blockNoDiff
+
           liftIO $ logInfo trce "Running extra migrations consumed_tx_out and prune tx_out"
           ExtraCons.migrateTxOut (Just trce)
```
kderme, line 112: Since `deleteAndUpdateConsumedTxOut` also does the migration, we may not need calling `migrateTxOut`

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,67 @@ countConsumed = do
...
+    migrateNextPage mxTxInId offst = do
+      page <- getInputPage offst pageSize
+      -- before page txinId delete after create extra collumn and update
+      mapM_ (handlePage mxTxInId) page
+      when (fromIntegral (length page) == pageSize) $
+        migrateNextPage mxTxInId $!
+          offst
+            + pageSize
+
+handlePage :: MonadIO m => TxInId -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m ()
```
kderme, line 227: This handles a single entry of the page, so `handlePage` can be confusing.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
Cmdv: ah ok, we have access to `wPruneTxOutPreviouslySet` so can pass that through as a check?

kderme: I don't think `wPruneTxOutPreviouslySet` can help here. It must be False, otherwise this code is not executed at all. 
`handlePage` being an iterative function can call `createConsumedTxOut` many times. Probably a new local flag can help.

Cmdv: sorry I edited it, we have `hasConsumedField` which should do the trick

cardano-db/src/Cardano/Db/Multiplex.hs
```diff
@@ -105,10 +105,8 @@ runExtraMigrations trce blockNoDiff PruneConsumeMigration {..} = do
...
         (False, _, True) -> do
           shouldInsertToMigrationTable
-          liftIO $ logInfo trce "Running extra migrations consumed_tx_out and prune tx_out"
-          ExtraCons.migrateTxOut (Just trce)
           liftIO $ logInfo trce "Now Running extra migration prune tx_out"
-          ExtraCons.deleteConsumedTxOut trce blockNoDiff
+          ExtraCons.deleteAndUpdateConsumedTxOut trce hasConsumedField blockNoDiff
```
kderme, line 109: `hasConsumedField` is always False here, so why it gets passed to `deleteAndUpdateConsumedTxOut`?

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,66 @@ countConsumed = do
...
+      when (fromIntegral (length page) == pageSize) $
+        migrateNextPage mxTxInId $!
+          offst
+            + pageSize
+
+    handlePageEntry :: MonadIO m => TxInId -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m ()
+    handlePageEntry maxTxInId (txInId, txId, index)
+      | txInId <= maxTxInId = deleteTxOutConsumed txId index
+      | otherwise = do
+          when hasConsumedField createConsumedTxOut
```
kderme, line 230: I'm a bit confused. `createConsumedTxOut` should be called if the field doesn't exist. Also `handlePageEntry` being a iterative function can still call `createConsumedTxOut` more than once.

kderme (changes requested):

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
Cmdv: I was looking for a way to know if it should `createConsumedTxOut` from `syncEnv` then it would never create more than once.

cardano-db/src/Cardano/Db/Multiplex.hs
Cmdv: having a `TVar` would do the trick but I don't have access to `SyncEnv` here in `/cardano-db/`. This logic should all probably live in `/cardano-db-sync/` 🤔

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,70 @@ countConsumed = do
...
+            + pageSize
+
+    handlePageEntry :: MonadIO m => TxInId -> Bool -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m Bool
+    handlePageEntry maxTxInId ranCreateConsumed (txInId, txId, index)
+      | txInId <= maxTxInId = do
+          deleteTxOutConsumed txId index
+          pure False
+      | otherwise = do
+          -- we want to call createConsumedTxOut just once
+          unless ranCreateConsumed createConsumedTxOut
```
Cmdv, line 233: @kderme A problem might be this per "page" so each time a new page is called it will reset `ranCreateConsumed` to `False` might have to change `migrateNextPage` to be a `FoldM` too and return `(ReaderT SqlBackend m (Int, Bool)` 🤔

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,70 @@ countConsumed = do
     pure countRows
   pure $ maybe 0 unValue (listToMaybe res)
 
+deleteAndUpdateConsumedTxOut :: MonadIO m => Trace IO Text -> Word64 -> ReaderT SqlBackend m ()
+deleteAndUpdateConsumedTxOut trce blockNoDiff = do
+  maxTxInId <- findMaxTxInId blockNoDiff
+  case maxTxInId of
+    Left errMsg -> liftIO $ logInfo trce $ "No tx_out were deleted: " <> errMsg
```
kderme, line 213: I think this should be handled differently. If we don't find a `maxTxInId` this means there are no values to delete. However, there are still values to update and the extra column should still be created.

kderme (changes requested):

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +206,70 @@ countConsumed = do
...
+      res <- foldM (handlePageEntry mxTxInId) ranCreateConsumed page
+      when (fromIntegral (length page) == pageSize) $
+        migrateNextPage mxTxInId res $!
+          offst
+            + pageSize
+
+    handlePageEntry :: MonadIO m => TxInId -> Bool -> (TxInId, TxId, Word64) -> ReaderT SqlBackend m Bool
+    handlePageEntry maxTxInId rcc (txInId, txId, index)
+      | txInId <= maxTxInId = do
+          deleteTxOutConsumed txId index
```
kderme, line 229: It may be worth examining if a whole page can be deleted once, with a single db query, instead of doing it entry by entry. This optimization can be done on a different pr.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
Cmdv: yeah think you are right, not sure if it's related but the existing prune/consume tests are failing because they always expect `migrateTxOut` to be ran.
I'm actually not quite grasping it tbh as I don't fully understand the functionality and order of how we handle UTXOs from reading the code :)

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +222,114 @@ countConsumed = do
...
+    -- the list has both bellow and above maxTxInId
+    (xs, ys) -> do
+      deleteEntries xs
+      shouldCreateConsumedTxOut ranCreateConsumedTxOut
+      updateEntries ys
+      pure True
+  where
+    -- the delete + update can happen as one query
+    deleteEntries = mapM_ (\(_, txId, index) -> deleteTxOutConsumed txId index)
+    updateEntries = mapM_ (\(txInId, txId, index) -> updateTxOutConsumedByTxInIdUnique txId index txInId)
```
Cmdv, line 281: @kderme I've isolated the place we could bulk delete/update but I'm not 100% how to do that in persistent when it comes to `deleteWhere`

kderme: I guess it could work like 
```
deleteWhere [FilterOr <$> mkFilter xs]
```
or by using `IN` clause.
But that's great already, it makes it easy to try different queries during benchmarking.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,40 +222,114 @@ countConsumed = do
...
+  (MonadIO m, MonadBaseControl IO m) =>
+  Trace IO Text ->
+  Word64 ->
+  ReaderT SqlBackend m ()
+deleteAndUpdateConsumedTxOut trce blockNoDiff = do
+  maxTxInId <- findMaxTxInId blockNoDiff
+  case maxTxInId of
+    Left _ -> do
+      createConsumedTxOut
+      liftIO $ logInfo trce "Created ConsumedTxOut as there were no blocks present"
```
kderme, line 236: The comment here is a bit misleading. The returned Left message, describes in a better way why no entries need to be deleted.

Other than that, this case should still call `splitAndProcessPageEntries` and update for all `tx_in` entries.

kderme (changes requested):
https://github.com/input-output-hk/cardano-db-sync/pull/1519/files#r1340198363

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -202,45 +224,139 @@ countConsumed = do
...
+    where_
+      ( foldl1
+          (||.)
+          ( map
+              ( \(_, txId, index) ->
+                  txOut E.^. TxOutTxId E.==. val txId E.&&. txOut E.^. TxOutIndex E.==. val index
+              )
+              transactionEntries
+          )
+      )
```
Cmdv, line 308: @kderme seems to work but would need double checking and benchmarking with original I'd say!

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
Cmdv: it can't as `migrateNextPage` needs `mxtxid` might need a code example of what you are thinking.

kderme: It could take a `Maybe` and only update if it's `Nothing`

Cmdv: it can't as `migrateNextPage` needs `mxtxid` might need a code example of what you are thinking.
Also the call site used to be:
```
(False, _, True) -> do
          shouldInsertToMigrationTable
          liftIO $ logInfo trce "Running extra migrations consumed_tx_out and prune tx_out"
          ExtraCons.migrateTxOut (Just trce)
          liftIO $ logInfo trce "Now Running extra migration prune tx_out"
          ExtraCons.deleteConsumedTxOut trce blockNoDiff
         ...
```
`deleteConsumedTxOut` just logs for our left:
```
deleteConsumedTxOut trce blockNoDiff = do
  maxTxInId <- findMaxTxInId blockNoDiff
  case maxTxInId of
    Left errMsg -> liftIO $ logInfo trce $ "No tx_out was deleted: " <> errMsg
```
so to be the same as before `deleteAndUpdateConsumedTxOut` needs to `migrateTxOut` + `logInfo`, unless I'm missing something?

kderme: Yes so the current implementation doesn't log the Left result of `findMaxTxInId` which I think it should.

Cmdv: ah ok I didn't understand this part:
> this case should still call `migrateNextPage` and update for all `tx_in` entries
I'll change the log wording 👍

kderme: These are two separate issues actually.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
kderme: I tried this on mainnet, but got an out of memory sql error. Same when I used a 10_000 page size.
I also tried this approach 
```
  deleteWhere [FilterOr $ mkFilter <$> transactionEntries]
  where
    mkFilter (_, txOutId, index) = FilterAnd [TxOutTxId ==. txOutId, TxOutIndex ==. index]
```
but eventually I got faster results by simply deleting the entries one by one.

cardano-db/src/Cardano/Db/Migration/Extra/CosnumedTxOut/Queries.hs
```diff
@@ -170,10 +146,220 @@ _validateMigration trce = do
...
+  (MonadIO m, MonadBaseControl IO m) =>
+  Trace IO Text ->
+  Word64 ->
+  ReaderT SqlBackend m ()
+deleteAndUpdateConsumedTxOut trce blockNoDiff = do
+  maxTxInId <- findMaxTxInId blockNoDiff
+  case maxTxInId of
+    Left errMsg -> do
+      liftIO $ logInfo trce $ "No tx_out was deleted: " <> errMsg
+      migrateNextPage Nothing False 0
```
kderme, line 218:
```suggestion
      migrateTxOut (Just trce)
```
This may also work. I saw you initially did it [here](https://github.com/input-output-hk/cardano-db-sync/compare/6836212277ea853ca00f29116ec141e5a99c8380..d034f391b42423c4426912590530597d6fb94b8a#diff-f2320ab911286735a4cc1c93a700fd7fe3f14fe3d2400201823b93fa916190c0L237), but then you changed it back

Cmdv: I made those changes following your previous comments https://github.com/input-output-hk/cardano-db-sync/pull/1519#discussion_r1340198363

kderme: The comment was one day before this commit.

Cmdv: One of your comments mentioned:
> It could take a `Maybe` and only update if it's `Nothing`

> this case should still call `migrateNextPage` and update for all `tx_in` entries.

So I made the changes as advised but you're saying put it back to just doing `migrateTxOut (Just trce)` rather than pass the `Maybe`?

kderme: A misunderstanding then. I commended when the code looked like 
```suggestion
      createConsumedTxOut
      liftIO $ logInfo trce "Created ConsumedTxOut as there were no blocks present"
```
and hadn't seen https://github.com/input-output-hk/cardano-db-sync/pull/1519/commits/d649e18ec0f37e20c1f027eb68bfe9c8c3c1ff75, which actually looks like a good idea, unless I'm missing something.

kderme: So either way, the current or the above should be fine.

kderme (approved):

PR #1529, opened 2023-10-03
Add Conway Tests

sgillespie:
# Description

Addresses a subset of #1526. In particular, I've added the following test groups:

 * Simple
 * Command Line Arguments

I've also snuck in a few minor improvements:

 * Omit postgres logs from nix test runner output
 * Eliminated some duplication between the era-specific transaction forging

# Checklist

- [x] Self-reviewed the diff

# Migrations


kderme:
I haven't checked but MigrateConsumedPruneTxOut tests may not pass.

sgillespie:
> I haven't checked but MigrateConsumedPruneTxOut tests may not pass.

You're right:
```
prune with full tx rollback:                                             FAIL (0.33s)
          test/Test/Cardano/Db/Mock/Validate.hs:126:
          Unexpected tx unspent count between tx-in & tx-out
          expected: 12
           but got: 14
          Use -p '/consumed-tx-out + prune-tx-out.prune with full tx rollback/' to rerun this test only.
```

kderme:
These tests have been adjusted at https://github.com/input-output-hk/cardano-db-sync/pull/1539 and require integration

kderme (approved):

Issue #1532, opened 2023-10-11
Rename Offline to OffChain


Cmdv:
As part of https://github.com/input-output-hk/cardano-db-sync/issues/1527 we are going to be fetching more "off chain" data this time with "anchor voting".

There will be parts of the code that can be shared as most of the functionality will be very similar but the naming can be rather confusing as it uses "Offline" which insinuates that something isn't active but in reality it's about getting data that is not on the chain, thus "OffChain".

PR #1533, opened 2023-10-11
Rename offline to offchain

Cmdv:
# Description

this fixes https://github.com/input-output-hk/cardano-db-sync/issues/1532

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/OffChain.hs
```diff
@@ -0,0 +1,154 @@
+{-# LANGUAGE FlexibleContexts #-}
```
kderme, line 1: Could this use 
```
git mv cardano-db-sync/src/Cardano/DbSync/Era/Shelley/OffChain.hs
```
to retain file history?

kderme (commented):
A new .sql file is probaby needed since the schema has changed.
Also could you fill the autogenerated check-list of the pr, especially the `Migration` paragraph which is relevant here?

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/OffChain.hs
sgillespie: I'm not sure this would have the desired effect. 

> Git has a rename command git mv, but that is just for convenience. The effect is indistinguishable from removing the file and adding another with different name and the same content. 

More info here: https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/GitFaq.html#Why_does_Git_not_.22track.22_renames.3F

cardano-db-sync/app/http-get-json-metadata.hs
```diff
@@ -3,8 +3,8 @@
 import Cardano.Db (PoolMetaHash (..), PoolUrl (..))
 import Cardano.DbSync (
   FetchError (..),
-  SimplifiedPoolOfflineData (..),
-  httpGetPoolOfflineData,
+  SimplifiedPoolChainlineData (..),
```
sgillespie, line 6: Is this a typo?

Cmdv: 😂 yeah that's a good one

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -46,10 +46,10 @@ import Cardano.DbSync.Config.Types (
 import Cardano.DbSync.Database
 import Cardano.DbSync.DbAction
 import Cardano.DbSync.Era
-import Cardano.DbSync.Era.Shelley.Offline.Http (
+(import Cardano.DbSync.Era.Shelley.OffChain.Http (
```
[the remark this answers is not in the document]
Cmdv: you never seen `(import ` before? 🤦 I just put this PR up not really checking anything!!

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/OffChain.hs
```diff
@@ -0,0 +1,154 @@
...
+    httpGetOffChainPoolData manager request (pfrPoolUrl pfr) (pfrPoolMDHash pfr)
+  where
+    convert :: Either FetchError SimplifiedOffChainPoolData -> OffChainPoolResult
+    convert eres =
+      case eres of
+        Right smd ->
+          OffChainPoolResultMetadata $
+            ResultMetadata $
+              DB.OffChainPoolData
+                { DB.OffChainPoolDataId = pfrPoolHashId pfr
```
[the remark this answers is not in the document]
Cmdv: no you're right it should be `OffChainPoolDataPoolId`, hls gave me zero errors "apparently"

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/OffChain.hs
kderme: Indeed, but in addition a separate "move files" commit and then an "edit file" commit would help git track the change better.

Cmdv: It didn't work if you check the git history, it shows in different commits the file was reverted -> renamed -> edited but github seems to pick it up as a brand new file 🤷

schema/migration-2-0029-20231012.sql
```diff
@@ -0,0 +1,24 @@
...
+    EXECUTE 'CREATe TABLE "off_chain_pool_data"("id" SERIAL8  PRIMARY KEY UNIQUE,"pool_id" INT8 NOT NULL,"ticker_name" VARCHAR NOT NULL,"hash" hash32type NOT NULL,"json" jsonb NOT NULL,"bytes" bytea NOT NULL,"pmr_id" INT8 NOT NULL)' ;
+    EXECUTE 'ALTER TABLE "off_chain_pool_data" ADD CONSTRAINT "unique_off_chain_pool_data" UNIQUE("pool_id","hash")' ;
+    EXECUTE 'CREATe TABLE "off_chain_pool_fetch_error"("id" SERIAL8  PRIMARY KEY UNIQUE,"pool_id" INT8 NOT NULL,"fetch_time" timestamp NOT NULL,"pmr_id" INT8 NOT NULL,"fetch_error" VARCHAR NOT NULL,"retry_count" word31type NOT NULL)' ;
+    EXECUTE 'ALTER TABLE "off_chain_pool_fetch_error" ADD CONSTRAINT "unique_off_chain_pool_fetch_error" UNIQUE("pool_id","fetch_time","retry_count")' ;
+    EXECUTE 'CREATe TABLE "off_chain_anchor_data"("id" SERIAL8  PRIMARY KEY UNIQUE,"voting_anchor_id" INT8 NOT NULL,"hash" BYTEA NOT NULL,"json" jsonb NOT NULL,"bytes" bytea NOT NULL)' ;
+    EXECUTE 'CREATe TABLE "off_chain_anchor_fetch_error"("id" SERIAL8  PRIMARY KEY UNIQUE,"voting_anchor_id" INT8 NOT NULL,"fetch_error" VARCHAR NOT NULL,"retry_count" word31type NOT NULL)' ;
+    -- Hand written SQL statements can be added here.
```
kderme, line 15: This migration doesn't seem to be correct or what we want. It creates new tables without deleting the previous ones. But even this wouldn't be correct. Ideally we want a table rename, so that migrations with existing data are not lost.

Cmdv: would we transfer the data to the new table then delete old tables?

sgillespie:
There are still 2 failing tests:
```
Migration is idempotent:                                 2023-10-17 12:58:24.330 UTC [143] ERROR:  relation "pool_offline_fetch_error" does not exist
2023-10-17 12:58:24.330 UTC [143] STATEMENT:  CREATE INDEX IF NOT EXISTS idx_pool_offline_fetch_error_pmr_id ON pool_offline_fetch_error (pmr_id);
ExitFailure 3

Errors in file: /tmp/migrate-2023-10-17T12:58:24.118194028Z.log

FAIL
      Exception: ExitFailure 1
      Use -p '/Migration is idempotent/' to rerun this test only.

    Insert foreign key missing:                              2023-10-17 12:58:24.350 UTC [147] ERROR:  constraint "unique_off_chain_pool_fetch_error" for table "off_chain_pool_fetch_error" does not exist
2023-10-17 12:58:24.350 UTC [147] STATEMENT:  INSERT INTO off_chain_pool_fetch_error ("pool_id", "fetch_time", "pmr_id", "fetch_error", "retry_count") VALUES (1, '2023-10-17 12:58:24.348073899Z', 1, 'too good', 5) ON CONFLICT ON CONSTRAINT unique_off_chain_pool_fetch_error DO UPDATE SET "pool_id" = EXCLUDED."pool_id" RETURNING id ;
FAIL
      Exception: DbInsertException "OffChainPoolFetchError" (SqlError {sqlState = "42704", sqlExecStatus = FatalError, sqlErrorMsg = "constraint \"unique_off_chain_pool_fetch_error\" for table \"off_chain_pool_fetch_error\" does not exist", sqlErrorDetail = "", sqlErrorHint = ""})
      Use -p '/Insert foreign key missing/' to rerun this test only.
```

schema/migration-2-0029-20231017.sql
```diff
@@ -0,0 +1,29 @@
...
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 29 THEN
+    EXECUTE 'ALTER TABLE IF EXISTS "pool_offline_data" RENAME TO "off_chain_pool_data"' ;
+    EXECUTE 'ALTER TABLE IF EXISTS "off_chain_pool_data" RENAME CONSTRAINT "unique_pool_offline_data" TO "unique_off_chain_pool_data"' ;
+
+    EXECUTE 'ALTER TABLE IF EXISTS "pool_offline_fetch_error" RENAME TO "off_chain_pool_fetch_error"' ;
+    EXECUTE 'ALTER TABLE IF EXISTS "pool_offline_fetch_error" RENAME CONSTRAINT "unique_pool_offline_fetch_error" TO "unique_off_chain_pool_fetch_error"' ;
```
kderme, line 13: I think this will create the constraint if the old one exists.However when starting from a fresh db, the constraint won't ever be created, due to the `IF EXISTS`

schema/migration-2-0029-20231017.sql
```diff
@@ -0,0 +1,53 @@
...
+      EXECUTE 'ALTER TABLE "pool_offline_fetch_error" RENAME TO "off_chain_pool_fetch_error"' ;
+      EXECUTE 'ALTER TABLE "off_chain_pool_fetch_error" RENAME CONSTRAINT "unique_pool_offline_fetch_error" TO "unique_off_chain_pool_fetch_error"' ;
+      EXECUTE 'ALTER TABLE "off_chain_pool_fetch_error" RENAME CONSTRAINT "pool_offline_fetch_error_pkey" TO "off_chain_pool_fetch_error_pkey"' ;
+    ELSE
+      EXECUTE 'CREATE TABLE "off_chain_pool_fetch_error"("id" SERIAL8 PRIMARY KEY UNIQUE,"pool_id" INT8 NOT NULL,"fetch_time" timestamp NOT NULL,"pmr_id" INT8 NOT NULL,"fetch_error" VARCHAR NOT NULL,"retry_count" word31type NOT NULL)' ;
+      EXECUTE 'ALTER TABLE "off_chain_pool_fetch_error" ADD CONSTRAINT "unique_off_chain_pool_fetch_error" UNIQUE("pool_id","fetch_time","retry_count")' ;
+    END IF ;
+
+    -- if the table "pool_offline_fetch_error" all ready exists just rename everything.
+    IF EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'anchor_offline_data') THEN
```
kderme, line 31: I haven't checked the full logic, but for the anchor voting related tables, probably you can simply edit the existing files, as they have not been released yet.

sgillespie (dismissed):
LGTM!

schema/migration-2-0029-20231017.sql
```diff
@@ -0,0 +1,40 @@
...
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 29 THEN
+    -- if the table "pool_offline_data" all ready exists just rename everything
+    IF EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'pool_offline_data') THEN
+      EXECUTE 'ALTER TABLE "pool_offline_data" RENAME TO "off_chain_pool_data"' ;
+      EXECUTE 'ALTER TABLE "off_chain_pool_data" RENAME CONSTRAINT "unique_pool_offline_data" TO "unique_off_chain_pool_data"' ;
+      EXECUTE 'ALTER TABLE "off_chain_pool_data" RENAME CONSTRAINT "pool_offline_data_pkey" TO "off_chain_pool_data_pkey"' ;
+    -- otherwise we create the newly named table and it's constraint "off_chain_pool_data"
+    ELSE
```
kderme, line 15: I think only this `else` case will ever hit, both for migrations from existing dbs and from empty. So postgres will attempt to recreate an existing table.

kderme (changes requested):

schema/migration-3-0002-20200521.sql
```diff
@@ -9,5 +9,10 @@ CREATE INDEX IF NOT EXISTS idx_tx_block_id ON tx(block_id);
...
 CREATE INDEX IF NOT EXISTS idx_pool_metadata_ref_pool_id ON pool_metadata_ref(pool_id);
-CREATE INDEX IF NOT EXISTS idx_pool_offline_fetch_error_pmr_id ON pool_offline_fetch_error (pmr_id);
-CREATE INDEX IF NOT EXISTS idx_pool_offline_data_pmr_id ON pool_offline_data (pmr_id);
+
+-- as of 13.2.0.0 offline was renamed to off_chain so we rename previous indexes if they exists
+ALTER INDEX [IF EXISTS] idx_pool_offline_data_pmr_id TO idx_off_chain_pool_data_pmr_id;
+ALTER INDEX [IF EXISTS] idx_pool_offline_fetch_error_pmr_id TO idx_off_chain_pool_fetch_error_pmr_id;
+
+CREATE INDEX IF NOT EXISTS idx_off_chain_pool_fetch_error_pmr_id ON off_chain_pool_fetch_error (pmr_id);
+CREATE INDEX IF NOT EXISTS idx_off_chain_pool_data_pmr_id ON off_chain_pool_data (pmr_id);
```
Cmdv, line 18: @kderme I've amended it so that if the index already exists then it will be renamed otherwise create the index `IF NOT EXISTS` when a new db.

Let me know if this is the right file to do this rename. as I'm unsure as to when this file is ran. (before/after `migrate-2`, when running migration script, on new startup 🤷)

kderme: I think this is correct now!
Stage-3 runs after stage-2. Stage-3 files are always executed and their Idempotence is guaranteed by the `IF (NOT) EXISTS` clause. While stage-2 idepotence is guaranteed by the `IF next_version =` clause

cardano-db-sync/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 * When `--disable-ledger` is used, db-sync won't populate the `tx.deposit` field as it's too expensive
 * Support for Conway (see schema docs)
 
-## 13.1.1.3
```
kderme, line 10: Was this accidentally changed? 13.2.0.0 exists above

kderme (dismissed):

kderme (approved):

sgillespie (approved):

Issue #1535, opened 2023-10-16
Improve documentation migrations


Cmdv:
When trying to generate a new migration file, existing documentation didn't actually work.
Just needs updating to reflect some missing procedures when trying to generate new migration file.

PR #1536, opened 2023-10-16
Update migration documentation

Cmdv:
# Description

This fixes https://github.com/input-output-hk/cardano-db-sync/issues/1535

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/doc/schema-management.md
+++ b/doc/schema-management.md
@@ -36,13 +36,28 @@ order them in the correct order for applying to the database.
 
 ## Creating a Migration
 
-Whenever the Haskell schema definition in `Cardano.Db.Schema` is updated, a schema migration can
-be generated using the command:
+Whenever the Haskell schema definition in `Cardano.Db.Schema` is updated, a schema migration will need to be migrated.
+When migrating, `db-sync` caches some of the schema files when it is built so this can be point of confusion.
+
+Firstly you need to run current existing migrations by:
+```
+export PGPASSFILE=config/pgpass-mainnet
+cabal run cardano-db-tool -- run-migrations --mdir schema/
+```
+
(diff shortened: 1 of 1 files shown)


doc/schema-management.md
```diff
@@ -36,13 +36,28 @@ order them in the correct order for applying to the database.
...
+Whenever the Haskell schema definition in `Cardano.Db.Schema` is updated, a schema migration will need to be migrated.
+When migrating, `db-sync` caches some of the schema files when it is built so this can be point of confusion.
+
+Firstly you need to run current existing migrations by:
+```
+export PGPASSFILE=config/pgpass-mainnet
+cabal run cardano-db-tool -- run-migrations --mdir schema/
+```
+
+Once this has completed it's good practice to rebuild `cardano-db-sync` due to how it caches schema files when built, this can be done using the following documentation [Build and Install](./installing.md#build-and-install)
```
kderme, line 48: I think this should be moved after the `create-migration`, because this is what invalidates the schema cache.

doc/schema-management.md
```diff
@@ -36,13 +36,28 @@ order them in the correct order for applying to the database.
...
+
+**Note:**  It is usually best to run the test suite which tests the migrations then generate the migration, this is achieved by doing:
+```
+cd ./cardano-chain-gen
+cabal run test:cardano-chain-gen
```
kderme, line 53: This existed before, but I'm not sure I understand the miotivation.

Cmdv: Think it was so it ran the migration into the test db, but think it might be best to just run it once at the end as a check that it all worked?

kderme (approved):

PR #1547, opened 2023-10-24
Add more Conway tests

sgillespie:
# Description

Addresses a subset of https://github.com/input-output-hk/cardano-db-sync/issues/1526. In particular, I've added the following test groups:

 * rollbacks
 * different configs
 * blocks with txs
 * stake addresses
 * stake distribution
 * rewards

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/ForceIndex.hs
```diff
@@ -0,0 +1,34 @@
...
+import Test.Tasty.HUnit (Assertion ())
+import Prelude ()
+
+checkForceIndexesArg :: IOManager -> [(Text, Text)] -> Assertion
+checkForceIndexesArg =
+  withCustomConfig cliArgs conwayConfigDir testLabel $ \_ _ dbSync -> do
+    startDBSync dbSync
+
+    -- Verify number of DB indexes
+    assertEqQuery dbSync DB.queryPgIndexesCount 159 "unexpected number of indexes"
```
kderme, line 20: Fyi after a rebase this value may need to change to 161

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/ForceIndex.hs
```diff
@@ -0,0 +1,34 @@
...
+    cliArgs = initCommandLineArgs {claForceIndexes = True}
+    testLabel = "conwayCLACheckForceIndexesArg"
+
+checkNoForceIndexesArg :: IOManager -> [(Text, Text)] -> Assertion
+checkNoForceIndexesArg =
+  withCustomConfigAndDropDB cliArgs conwayConfigDir testLabel $ \_ _ dbSync -> do
+    startDBSync dbSync
+
+    -- Verify number of DB indexes
+    assertEqQuery dbSync DB.queryPgIndexesCount 94 "unexpected number of indexes"
```
kderme, line 31: Same here. To 96

cardano-chain-gen/src/Cardano/Mock/Forging/Tx/Conway.hs
```diff
@@ -124,6 +184,81 @@ mkSimpleTx isValid' txBody =
...
+  Delegatee StandardCrypto ->
+  StakeCredential StandardCrypto ->
+  ConwayTxCert StandardConway
+mkDelegTxCert delegatee = mkTxDelegCert (`ConwayDelegCert` delegatee)
+
+mkTxDelegCert ::
+  (StakeCredential StandardCrypto -> ConwayDelegCert StandardCrypto) ->
+  StakeCredential StandardCrypto ->
+  ConwayTxCert StandardConway
+mkTxDelegCert = (ConwayTxCertDeleg .)
```
Cmdv, line 260: minor change if you agree, just wondering if this might be a little too point free?

Cmdv (approved):
Really good, appreciate the comments in the tests cases too. Makes it much clearer what is happening 🎉 👌 :shipit:

cardano-chain-gen/src/Cardano/Mock/Forging/Tx/Conway.hs
sgillespie: I went back and forth a few times, I don't mind making it a little more readable

Cmdv: no big deal we can leave it 👍

Cmdv (approved):
🎉

kderme (approved):
LGTM! Is there a list of tests that were not tranfered, like `rewardsDeregistration` or had to be altered for Conway?

Cmdv:
@kderme I think he's still working through them but is going to break up the PR as it's getting rather large 😄

sgillespie:
> LGTM! Is there a list of tests that were not tranfered, like rewardsDeregistration or had to be altered for Conway?

As far is I can remember, only tests from the rewards suite were not transferred:

 * rewardsDeregistration
 * rewardsReregistration
 * mirReward
 * mirRewardDereg
 * mirReward

Furthermore, any test that used deleg certificates had to be slightly altered, since Conway changed this. You'll see this in the rollback and stake suites, but there should be no change in overall behavior.

> I think he's still working through them but is going to break up the PR as it's getting rather large

This is also correct.  I still have the following test suites left:

 * plutus send scripts
 * plutus cert scripts
 * MultiAssets plutus scripts
 * pools and smash
 * Babbage inline and reference
 * Hard Fork

PR #1561, opened 2023-11-16
Add more flags

kderme:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -24,7 +24,7 @@ main = do
     CmdVersion -> runVersionCommand
     CmdRun params -> do
       let maybeLedgerStateDir = enpMaybeLedgerStateDir params
-      case (maybeLedgerStateDir, enpShouldUseLedger params) of
+      case (maybeLedgerStateDir, enpHasLedger params) of
         (Just _, True) -> run params
         (Nothing, False) -> run params
         (Just _, False) -> run params
@@ -78,7 +78,8 @@ pRunDbSyncNode =
     <*> pPGPassSource
     <*> pEpochDisabled
     <*> pHasCache
-    <*> pUseLedger
+    <*> pHasLedger
+    <*> pShouldUseLedger
     <*> pSkipFix
     <*> pOnlyFix
     <*> pForceIndexes
@@ -188,13 +189,22 @@ pHasCache =
         <> Opt.help "Disables the db-sync caches. Reduces memory usage but it takes longer to sync."
     )
 
-pUseLedger :: Parser Bool
-pUseLedger =
+pHasLedger :: Parser Bool
+pHasLedger =
   Opt.flag
     True
     False
     ( Opt.long "disable-ledger"
         <> Opt.help "Disables the leger state. Drastically reduces memory usage and it syncs faster, but some data are missing."
+    )
+
+pShouldUseLedger :: Parser Bool
+pShouldUseLedger =
+  Opt.flag
+    True
+    False
+    ( Opt.long "dont-use-ledger"
+        <> Opt.help "Maintains the ledger state but doesn't use it, except to load UTxO. To be used with --bootstrap-tx-out"
     )
 
 pSocketPath :: Parser SocketPath
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -93,6 +93,7 @@ pRunDbSyncNode =
     <*> pForceTxIn
     <*> pDisableAllMode
     <*> pFullMode
+    <*> pOnlyUTxO
     <*> pOnlyGov
     <*> pMigrateConsumed
     <*> pPruneTxOut
@@ -315,6 +316,15 @@ pFullMode =
     True
     ( Opt.long "full"
         <> Opt.help "Makes db-sync run with all possible functionalities."
+    )
+
+pOnlyUTxO :: Parser Bool
+pOnlyUTxO =
+  Opt.flag
+    False
+    True
+    ( Opt.long "only-utxo"
+        <> Opt.help "A mode which combines --bootstrap-tx-out with some disable flags"
     )
 
 pOnlyGov :: Parser Bool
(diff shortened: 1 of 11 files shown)


cardano-db-sync/app/cardano-db-sync.hs
```diff
@@ -188,15 +190,24 @@ pHasCache =
         <> Opt.help "Disables the db-sync caches. Reduces memory usage but it takes longer to sync."
     )
 
-pUseLedger :: Parser Bool
-pUseLedger =
+pHasLedger :: Parser Bool
```
Cmdv, line 193: Could this be renamed to `ShouldUseLedgerState`? 

Just it's a little confusing when there is both `HasLedger` + `ShouldUseLedger` 🤔

Cmdv: I'm just saying `HasLedger` makes it confusing in it's naming, putting the word `State` somewhere with it makes it clearer which flag it came from.

Another confusion is that if `HasLedger = False` and `ShouldUseLedger = True`. How can one "use the ledger" if it doesn't "have one".

kderme: > I'm just saying HasLedger makes it confusing in it's naming, putting the word State somewhere with it makes it clearer which flag it came from.

It's quite common to use `Ledger` in code meaning `LedgerState`. Also the flags don't have the word state, ie `--disable-ledger`, `--dont-use-ledger`

> Another confusion is that if HasLedger = False and ShouldUseLedger = True. How can one "use the ledger" if it doesn't "have one".

I think this is not possible, since 
```
useLedger = enpHasLedger snp && enpShouldUseLedger snp && not (enpOnlyUTxO snp)
```
useLedger is True only if hasLedger is True

Cmdv: ok cool no worries 👍

doc/configuration.md
```diff
@@ -105,14 +105,27 @@ Disables fetching pool offchain metadata.
...
+### --dont-use-ledger : Experimental
+
+Maintains the ledger state, but doesn't use any of its data, except to load UTxO. To be used with `--bootstrap-tx-out`
+
+### --only-utxo : Experimental
+
+This is the equivalent of using `--dont-use-ledger`
+`--disable-shelley`, `--disable-plutus-extra`, `--disable-offchain-pool-data`, `--disable-gov`, `--bootstrap-tx-out`.
+This mode initially populates only a few tables, like `block` and `tx`. It maintains a ledger state but doesn't use any of its data. When syncing is completed, it loads the whole UTxO set from the ledger to the `tx_out` and `ma_tx_out` tables.
+After that db-sync can be restarted with `--disable-ledger-state` to continue
```
rdlrt, line 121: Can this be somehow simplified, it is a little hacky for operations/automations.
As against that, maybe the flag combinations could be adapted to extract UTxOs from ledger-state once upon booting initially, and then not rely on it beyond. This could potentially even be controlled against a control table (eg: config) within database. At the moment, the number of command line options and permutations seem to be getting a bit out-of-hand, especially from testing point of view

kderme: > Can this be somehow simplified, it is a little hacky for operations/automations.

Indeed, that's possible, it requires a somewhat big refactoring, to make the LedgerEnv a mutable var, but it's something we will eventually do.

> At the moment, the number of command line options and permutations seem to be getting a bit out-of-hand, especially from testing point of view.

Testing/QA is indeed a challenge with so many different ways to run db-sync. The plan is to have `modes`, which are common combinations of these flags, based on user input. We can then focus our testing efforts on these modes. `--only-utxo` which is introduced in this pr is an example of such mode which meets your requirement (except the ledger state automatic dropping which is not done in this pr).

cardano-db-sync/src/Cardano/DbSync/Ledger/Event.hs
```diff
@@ -95,8 +95,8 @@ toOrdering ev = case ev of
   LedgerStartAtEpoch {} -> 8
   LedgerNewEpoch {} -> 9
 
-convertAuxLedgerEvent :: OneEraLedgerEvent (CardanoEras StandardCrypto) -> Maybe LedgerEvent
-convertAuxLedgerEvent = toLedgerEvent . wrappedAuxLedgerEvent
+convertAuxLedgerEvent :: Bool -> OneEraLedgerEvent (CardanoEras StandardCrypto) -> Maybe LedgerEvent
+convertAuxLedgerEvent hasRew = toLedgerEvent hasRew . wrappedAuxLedgerEvent
```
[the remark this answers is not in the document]
kderme: Yes, I'll change it to `hasReward` to make it clearer.

doc/configuration.md
```diff
@@ -105,14 +105,27 @@ Disables fetching pool offchain metadata.
...
 Disables all data related to governance
 
-### --disable-all
+### --disable-all : Experimental
 
 Disables almost all data except `block`, `tx` and data related to the ledger state
 
+### --dont-use-ledger : Experimental
+
+Maintains the ledger state, but doesn't use any of its data, except to load UTxO. To be used with `--bootstrap-tx-out`
```
rdlrt, line 114: Should there be a flag combination that would skip maintaining ledger-state until reaching tip? Currently, it still takes ledger-state dump periodically during the sync. I'd have thought `--bootstrap-tx-out` would only maintain ledger-state upon block/tx table has synched to tip?

kderme: > Should there be a flag combination that would skip maintaining ledger-state until reaching tip (and then execute once)?

This can be done, but I'm not sure if it's worth it. The only benefit I can think of is that ledger state and the memory spike would be maintained for a smaller duration. After db-sync reaches tip without the ledger state it would have to restart the chainsync protocol from genesis to get again all the blocks and apply the ledger rules. Also i don't understand the `execute once` part.

> Currently, it still takes ledger-state dump periodically during the sync

This is an orthogonal improvement we can make, which applies to every long syncing scenario: we shouldn't take that frequent or any snapshots.

> Add entry in a config table to denote migration was done (to ignore bootstrap-tx-out even if specified)

This one exists in table `extra_migrations`

rdlrt: > After db-sync reaches tip without the ledger state it would have to restart the chainsync protocol from genesis to get again all the blocks and apply the ledger rules.

Oh - I misunderstood then, I thought similar to CLI/Ogmios, DBSync could fetch 'current' ledger-state dump from node without 'having' to validate entire history

kderme: I don't think these state queries are very reliable and with UTxO-HD coming, the available queries may change. Ideally the node would provide a way to stream UTxO from its current state, which would result in no memory spike on the DBSync side.

We will sync with the consensus team on this manner.

PR #1562, opened 2023-11-16
1527 conway offchain voting

### LINKED ISSUE #1527: Get offchain Voting Anchor
This is similar to the off chain pool metadata that we already have.
A new thread needs to be added. The db schema is already in place, but may need adjustments.

Cmdv:
# Description

Add OffChain Voting Anchor utilising existing offchain code for pools
this fixes #1527 

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -196,11 +197,13 @@ insertShelleyBlock syncEnv shouldLog withinTwoMins withinHalfHour blk details is
 
     insertStakeSlice syncEnv $ apStakeSlice applyResult
 
+    unless (ioGov iopts)
```
kderme, line 200:
```suggestion
    when (ioGov iopts)
```

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,283 @@
...
+    trce = getTrace syncEnv
+    iopts = getInsertOptions syncEnv
+
+    queuePoolInsert :: OffChainPoolResult -> IO ()
+    queuePoolInsert = atomically . writeTBQueue (envOffChainPoolResultQueue syncEnv)
+
+runFetchOffChainVoteThread :: SyncEnv -> IO ()
+runFetchOffChainVoteThread syncEnv = do
+  -- if dissable gov is active then don't run voting anchor thread
+  unless (ioGov iopts) $ do
```
kderme, line 192:
```suggestion
  when (ioGov iopts) $ do
```
`ioGov` means we have governance data

kderme (changes requested):

cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -313,6 +315,12 @@ insertCheckOffChainPoolFetchError pofe = do
   foundMeta <- existsPoolMetadataRefId (offChainPoolFetchErrorPmrId pofe)
   when (foundPool && foundMeta) . void $ insertCheckUnique "OffChainPoolFetchError" pofe
 
+insertOffChainVoteData :: (MonadBaseControl IO m, MonadIO m) => OffChainVoteData -> ReaderT SqlBackend m ()
+insertOffChainVoteData pod = void $ insertCheckUnique "OffChainVoteData" pod
```
kderme, line 319: There is a race condition here.
Normally when we insert in a table A that references another table B we do the following
```
bid <- queryTableB hash
insertA $ A {abid = bid, ...}
```
This provides atomicity, the query and the insert happens in the same db transaction. So it's impossible to have a rollback between the queryTableB and insertA

However here we have
```
thread 1:
bid <- queryTableB hash
pushQueue $ A { abid = bid }

thread 2:
a <- flushQueue
insertA a
```

So atomicity is lost, since insert and query don't happen in the same db transaction. This means that a rollback between the 2 operations will cause invalid data to be inserted in the metadata table. The solution is to redo the query inside insertA to validate that all referenced data are still there. You can check how `insertCheckOffChainPoolData` and `insertCheckOffChainPoolFetchError` solve the issue.

There is possibly an alternative solution, to clean the Queue when there is a rollback, but this would need some thinking.

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -19,10 +19,12 @@ module Cardano.DbSync (
   runDbSyncNode,
   runDbSync,
   -- For testing and debugging
-  FetchError (..),
+  OffChainFetchError (..),
   SimplifiedOffChainPoolData (..),
   httpGetOffChainPoolData,
-  parsePoolUrl,
+  parseOffChainPoolUrl,
```
kderme, line 25: These are most probably imported and reexporeted for tools like http-get-json-metadata.hs. But can't these tools directly import the underlying modules?

cardano-db/src/Cardano/Db/Types.hs
```diff
@@ -237,6 +238,11 @@ newtype VoteUrl = VoteUrl {unVoteUrl :: Text}
   deriving (Eq, Ord, Generic)
   deriving (Show) via (Quiet VoteUrl)
 
+-- | The raw binary hash of a stake vote metadata.
```
kderme, line 241:
```suggestion
-- | The raw binary hash of a vote metadata.
```
stake applies to pools

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,283 @@
...
+
+loadOffChainWorkQueue ::
+  (MonadBaseControl IO m, MonadIO m) =>
+  Trace IO Text ->
+  LoadOffChainWorkQueue a m ->
+  ReaderT SqlBackend m ()
+loadOffChainWorkQueue _trce offChainWorkQueue = do
+  whenM (liftIO $ atomically (isEmptyTBQueue (lQueue offChainWorkQueue))) $ do
+    now <- liftIO Time.getPOSIXTime
+    runnablePools <- filter (isRunnable now offChainWorkQueue) <$> lGetData offChainWorkQueue now 100
```
kderme, line 87:
```suggestion
    runnableOffChainData <- filter (isRunnable now offChainWorkQueue) <$> lGetData offChainWorkQueue now 100
```

since this is generic

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,280 @@
...
+
+loadOffChainWorkQueue ::
+  (MonadBaseControl IO m, MonadIO m) =>
+  Trace IO Text ->
+  LoadOffChainWorkQueue a m ->
+  ReaderT SqlBackend m ()
+loadOffChainWorkQueue _trce offChainWorkQueue = do
+  whenM (liftIO $ atomically (isEmptyTBQueue (lQueue offChainWorkQueue))) $ do
+    now <- liftIO Time.getPOSIXTime
+    runnableOffChainPools <- filter (isRunnable now offChainWorkQueue) <$> lGetData offChainWorkQueue now 100
```
kderme, line 87:
```suggestion
    runnableOffChainData <- filter (isRunnable now offChainWorkQueue) <$> lGetData offChainWorkQueue now 100
```

since its used for both pools and vote metadata.

cardano-db-sync/src/Cardano/DbSync/OffChain/Query.hs
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+queryOffChainVoteWorkQueue :: MonadIO m => UTCTime -> ReaderT SqlBackend m [OffChainVoteWorkQueue]
+queryOffChainVoteWorkQueue _now = do
+  res <- select $ do
+    (va :& ocpfe) <-
+      from
+        $ table @VotingAnchor
+          `innerJoin` table @OffChainVoteFetchError
+        `on` (\(va :& ocpfe) -> ocpfe ^. OffChainVoteFetchErrorVotingAnchorId ==. va ^. VotingAnchorId)
```
kderme, line 103: Similarly to `queryOffChainPoolWorkQueue`, we need to check here that there is no entry for this at `OffChainVoteData`.

kderme (changes requested):
Some remaining issues. Also I tested this on sanchonet and even though there are entries in `voting_anchor`, there are no entries in `off_chain_vote_data` or `off_chain_vote_fetch_error`.

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,280 @@
...
+runFetchOffChainVoteThread :: SyncEnv -> IO ()
+runFetchOffChainVoteThread syncEnv = do
+  -- if dissable gov is active then don't run voting anchor thread
+  when (ioGov iopts) $ do
+    logInfo trce "Running Offchain Vote Anchor fetch thread"
+    forever $ do
+      tDelay
+      -- load the offChain pool work queue using the db
+      _ <- runReaderT (loadOffChainVoteWorkQueue trce (envOffChainVoteWorkQueue syncEnv)) (envBackend syncEnv)
+      voteq <- blockingFlushTBQueue (envOffChainVoteWorkQueue syncEnv)
```
kderme, line 195: I think this is what causes the inexistence of data in sanchonet. It loads then it flushes and blocks. If at any point there are no data to load then `blockingFlushTBQueue` will block indefinitely. This causes a dead end, the Queue (currently) is not filled by a separate thread, so there is no way to make progress. One solution is to use the non blocking variant of `blockingFlushTBQueue`

kderme:
A separate issue on sanchonet when off_chain_vote_data was inserted
```
cardano-db-sync: DbInsertException "OffChainVoteData" (SqlError {sqlState = "22P02", sqlExecStatus = FatalError, 
sqlErrorMsg = "invalid input syntax for type json", sqlErrorDetail = "The input string ended unexpectedly.", sqlErrorHint = ""})
```

cardano-db-sync/src/Cardano/DbSync/OffChain/Http.hs
```diff
@@ -0,0 +1,232 @@
...
+                , spodContentType = mContentType
+                }
+        OffChainVoteUrl _ ->
+          pure $
+            SimplifiedOffChainVoteDataType
+              SimplifiedOffChainVoteData
+                { sovaHash = metadataHash
+                , sovaBytes = respBS
+                , -- TODO: no json format decided for vote metadata yet so we are leaving it blank for now
+                  sovaJson = "TODO"
```
kderme, line 175:
```suggestion
                  sovaJson = "{}"
```
since this needs to be a valid json format

kderme:
Could you add a Changelog entry under `sancho-next`

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -196,11 +196,13 @@ insertShelleyBlock syncEnv shouldLog withinTwoMins withinHalfHour blk details is
 
     insertStakeSlice syncEnv $ apStakeSlice applyResult
 
+    when (ioGov iopts)
+      . lift
+      $ insertOffChainVoteResults tracer (envOffChainVoteResultQueue syncEnv)
+
     when (ioOffChainPoolData iopts && unBlockNo (Generic.blkBlockNo blk) `mod` offChainModBase == 0)
```
kderme, line 203: Since this no longer does queries, we could call it for every block, so the `offChainModBase` check in unecessary.

kderme (changes requested):

kderme:
Could you add a new Changelog entry under `sancho-next`?

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,269 @@
...
+
+runFetchOffChainVoteThread :: SyncEnv -> IO ()
+runFetchOffChainVoteThread syncEnv = do
+  -- if dissable gov is active then don't run voting anchor thread
+  when (ioGov iopts) $ do
+    logInfo trce "Running Offchain Vote Anchor fetch thread"
+    forever $ do
+      tDelay
+      -- load the offChain pool work queue using the db
+      _ <- runReaderT (loadOffChainVoteWorkQueue trce (envOffChainVoteWorkQueue syncEnv)) (envBackend syncEnv)
```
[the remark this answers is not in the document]
Cmdv: do you mean we need to extend `SyncEnv` to be something like:
```
data SyncEnv = SyncEnv
  { envVoteBackend :: !SqlBackend
  , envPoolBackend :: !SqlBackend
  ...
  }
  ```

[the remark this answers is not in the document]
Comment (kderme) [reply cardano-db-sync/src/Cardano/DbSync/OffChain.hs:193]: It doesn't need to be in the `SyncEnv`. It can be initiated in the `runFetchOffChainPoolThread` and vote thread and remain there.

File: cardano-db-sync/src/Cardano/DbSync/OffChain.hs [hunk 13]
```diff
@@ -0,0 +1,269 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE NoImplicitPrelude #-}
```
sgillespie, line 7: I notice we have a lot of modules using both Prelude and Cardano.Prelude. Should we use this pattern everywhere (`NoImplicitPrelude`)?

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -0,0 +1,269 @@
...
+    queuePoolInsert = atomically . writeTBQueue (envOffChainPoolResultQueue syncEnv)
+
+runFetchOffChainVoteThread :: SyncEnv -> IO ()
+runFetchOffChainVoteThread syncEnv = do
+  -- if dissable gov is active then don't run voting anchor thread
+  when (ioGov iopts) $ do
+    logInfo trce "Running Offchain Vote Anchor fetch thread"
+    forever $ do
+      tDelay
+      -- load the offChain pool work queue using the db
```
sgillespie, line 192: I think this should be "offchain vote work queue"

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
Cmdv: could set it globally in the cabal file? unless we need `Prelude` for some reason

kderme (approved):

schema/migration-2-0034-20231220.sql
```diff
@@ -0,0 +1,24 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 34 THEN
+    EXECUTE 'ALTER TABLE "committee_de_registration" ADD COLUMN "cold_key" bytea NOT NULL' ;
+    EXECUTE 'ALTER TABLE "committee_de_registration" DROP COLUMN "hot_key"' ;
```
kderme, line 10: I think changes related to `committee_de_registration` should be deleted.

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -650,16 +650,21 @@ share
...
-    bytes                   ByteString          sqltype=bytea
+  OffChainVoteData
+    votingAnchorId      VotingAnchorId      noreference
+    hash                ByteString
+    json                Text                sqltype=jsonb
+    bytes               ByteString          sqltype=bytea
+    UniqueOffChainVoteData votingAnchorId hash
+    deriving Show
 
-  OffChainAnchorFetchError
```
kderme, line 659: These 2 tables that are deleted, should be deleted also from the sql files.

kderme (changes requested):

cardano-db/src/Cardano/Db/Schema.hs
Cmdv: having a bit of trouble with the rebasing!! just reran a migration but it's doing weird things it seems

kderme (approved):
LGTM!

PR #1567, opened 2023-11-30
Fixes for Sancho 2-3-0 migrations

sorki:
The `cold_key` to `hot_key` otherwise fails on populated table.

Second one is just a sequence renaming to match table renaming.

### DIFF
--- a/schema/migration-2-0032-20231107.sql
+++ b/schema/migration-2-0032-20231107.sql
@@ -6,8 +6,7 @@ DECLARE
 BEGIN
   SELECT stage_two + 1 INTO next_version FROM schema_version ;
   IF next_version = 32 THEN
-    EXECUTE 'ALTER TABLE "committee_de_registration" ADD COLUMN "cold_key" bytea NOT NULL' ;
-    EXECUTE 'ALTER TABLE "committee_de_registration" DROP COLUMN "hot_key"' ;
+    EXECUTE 'ALTER TABLE "committee_de_registration" RENAME COLUMN "cold_key" TO "hot_key"' ;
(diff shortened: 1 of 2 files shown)


kderme:
Thank you for both fixes.
For the first one, we don't really attempt to have backwards compatibility in these sancho pre-releases, but still good to have.

sorki:
> Thank you for both fixes. For the first one, we don't really attempt to have backwards compatibility in these sancho pre-releases, but still good to have.

Yw! No problem, it's quick to resync but for us it means some additional manual steps due to replication and custom stuff on top, so sometimes it is faster to just fix these by hand.

Cmdv:
@sorki 
> One more
> 
> ```
> ALTER INDEX unique_pool_offline_data RENAME TO unique_off_chain_pool_data;
> ALTER INDEX unique_pool_offline_fetch_error RENAME TO unique_off_chain_pool_fetch_error;
> ```

They already exist in [this](https://github.com/input-output-hk/cardano-db-sync/blob/master/schema/migration-2-0030-20231017.sql#L10) migration file 👍

kderme (approved):

schema/migration-2-0032-20231107.sql
```diff
@@ -6,8 +6,7 @@ DECLARE
 BEGIN
   SELECT stage_two + 1 INTO next_version FROM schema_version ;
   IF next_version = 32 THEN
-    EXECUTE 'ALTER TABLE "committee_de_registration" ADD COLUMN "cold_key" bytea NOT NULL' ;
-    EXECUTE 'ALTER TABLE "committee_de_registration" DROP COLUMN "hot_key"' ;
+    EXECUTE 'ALTER TABLE "committee_de_registration" RENAME COLUMN "cold_key" TO "hot_key"' ;
```
kderme, line 9:
```suggestion
    EXECUTE 'ALTER TABLE "committee_de_registration" RENAME COLUMN "hot_key" TO "cold_key"' ;
```

PR #1585, opened 2023-12-22
Make sure docker references are lower case

### LINKED ISSUE #1583: Intersect Migration: Docker publish is failing
After the Intersect GitHub migration, the "upload to ghcr.io" action is [failing](https://github.com/IntersectMBO/cardano-db-sync/actions/runs/7278004328/job/19831261116):

```
Error parsing reference: "ghcr.io/IntersectMBO/cardano-db-sync:sancho-3-0-0" is not a valid repository/tag: invalid reference format: repository name must be lowercase
```

In the immediate term, we'll manually push the failing tag. For future tags, we'll permanently fix the workflow

sgillespie:
# Description

Fixes #1583. Before uploading an image, downcase its repository name (ie IntersectMBO).

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/release-ghcr.yml
+++ b/.github/workflows/release-ghcr.yml
@@ -47,12 +47,23 @@ jobs:
 
       - name: Upload to ghcr.io
         run: |
+          # Downcase the package repository, because docker reference
+          # are required to be lower case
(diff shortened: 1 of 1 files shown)


Cmdv (approved):
🎉

PR #1590, opened 2024-01-03
Flag to keep specified Tx Metadata by key names

### LINKED ISSUE #1574: Filter tx metadata by key
tx metadata on mainnet are a lot in size and take a significant part of syncing time. For this reason we have a `--disable-metadata` flag which completely disables them. However some application may need only a specific set of metadata, based on its metadata key. For this reason we want to add a new option, in the form of `--tx-metadata <keys>`. <keys> could be a simple list of keys. When the metadata are prepared, only the specified keys will be taken into consideration. The `InsertOptions` field `ioMetadata` needs to be adjusted for this.

One possible application for this is [1694-labeled](https://github.com/Quantumplation/CIPs/blob/badf723b7d0de401b4f1aa954faa4adcce7f1d06/CIP-0100/README.md#hosting) metadata, but there are others

Cmdv:
# Description

This fixes #1574 

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -229,10 +229,11 @@ pSlotNo =
           <> Opt.metavar "WORD"
       )
 
-pKeepTxMetadata :: Parser [Text]
+pKeepTxMetadata :: Parser [Word64]
 pKeepTxMetadata =
   Opt.many
-    ( Opt.strOption
+    ( Opt.option
+        Opt.auto
         ( Opt.long "keep-tx-metadata"
             <> Opt.help "Insert a specific set of tx metadata, based on the tx metadata key names"
         )
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -7,6 +7,7 @@ import Cardano.DbSync.Config
 import Cardano.DbSync.Metrics (withMetricSetters)
 import Cardano.Prelude
 import Cardano.Slotting.Slot (SlotNo (..))
+import Data.List.Split (splitOn)
 import Data.String (String)
 import qualified Data.Text as Text
 import Data.Version (showVersion)
@@ -231,13 +232,17 @@ pSlotNo =
 
 pKeepTxMetadata :: Parser [Word64]
 pKeepTxMetadata =
-  Opt.many
-    ( Opt.option
-        Opt.auto
-        ( Opt.long "keep-tx-metadata"
-            <> Opt.help "Insert a specific set of tx metadata, based on the tx metadata key names"
-        )
+  Opt.option
+    (parseCommaSeparated <$> Opt.str)
+    ( Opt.long "keep-tx-metadata"
+        <> Opt.help "Insert a specific set of tx metadata, based on the tx metadata key names"
     )
+  where
+    parseCommaSeparated :: String -> [Word64]
+    parseCommaSeparated str =
+      case traverse readMaybe (splitOn "," str) of
+        Just values -> values
+        Nothing -> error "Failed to parse comma-separated values"
 
 pHasInOut :: Parser Bool
 pHasInOut =
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -87,6 +87,7 @@ pRunDbSyncNode =
     <*> pHasShelley
     <*> pHasMultiAssets
     <*> pHasMetadata
+    <*> pKeepTxMetadata
     <*> pHasPlutusExtra
     <*> pHasGov
     <*> pHasOffChainPoolData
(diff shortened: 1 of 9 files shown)


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -1205,28 +1208,38 @@ prepareTxMetadata ::
...
-      mjson <- safeDecodeToJson tracer "prepareTxMetadata" jsonbs
-      pure
-        DB.TxMetadata
-          { DB.txMetadataKey = DbWord64 key
-          , DB.txMetadataJson = mjson
-          , DB.txMetadataBytes = singleKeyCBORMetadata
-          , DB.txMetadataTxId = txId
-          }
+      let metadataNames = ioKeepMetadataNames inOpts
+          isMatchingKey = key `elem` map (read . Text.unpack) metadataNames
```
kderme, line 1225: The read can fail at this point. It's better to parse it once when dbsync starts as Word64.

cardano-db-sync/app/cardano-db-sync.hs
```diff
@@ -228,6 +229,15 @@ pSlotNo =
           <> Opt.metavar "WORD"
       )
 
+pKeepTxMetadata :: Parser [Text]
```
kderme, line 232: A simple list may not be enough to represent this. For example does the emply list mean to keep all or to keep nothing?

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -235,7 +235,7 @@ extractSyncOptions snp aop =
     iopts
       | enpOnlyGov snp = onlyGovInsertOptions useLedger
       | enpOnlyUTxO snp = onlyUTxOInsertOptions
-      | enpFullMode snp = fullInsertOptions useLedger
+      | enpFullMode snp = fullInsertOptions useLedger (enpKeepMetadataNames snp)
```
kderme, line 238: This defaults to `[]`, so by default no metadata will be kept. This changes the dewfault behaviour, which should be to keep all.

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -199,8 +199,8 @@ getPrunes :: SyncEnv -> Bool
 getPrunes = do
   DB.pcmPruneTxOut . getPruneConsume
 
-fullInsertOptions :: Bool -> InsertOptions
-fullInsertOptions useLedger = InsertOptions True useLedger True True True True True True True
+fullInsertOptions :: Bool -> [Text] -> InsertOptions
+fullInsertOptions useLedger keepTxMDNames = InsertOptions True useLedger True True True True keepTxMDNames True True True
```
sgillespie, line 203: Personally, I would prefer to see the record form of the constructor here, it's a lot easier to follow when there are many arguments of the same type.

cardano-db-sync/src/Cardano/DbSync/Api/Types.hs
```diff
@@ -78,6 +78,7 @@ data InsertOptions = InsertOptions
   , ioRewards :: !Bool
   , ioMultiAssets :: !Bool
   , ioMetadata :: !Bool
+  , ioKeepMetadataNames :: ![Text]
```
sgillespie, line 81: I think `![a]` is a bit of a misnomer, because (if I understand correctly) it will only evaluate the head of the list

sgillespie (approved):

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -199,8 +199,8 @@ getPrunes :: SyncEnv -> Bool
 getPrunes = do
   DB.pcmPruneTxOut . getPruneConsume
 
-fullInsertOptions :: Bool -> InsertOptions
-fullInsertOptions useLedger = InsertOptions True useLedger True True True True True True True
+fullInsertOptions :: Bool -> Strict.Maybe [Word64] -> InsertOptions
+fullInsertOptions useLedger maybeKeepMNames = InsertOptions True useLedger True True True True maybeKeepMNames True True True
```
kderme, line 203: Since it's `full` there is no need to pass `maybeKeepMNames` here. It can be hardcoded to keep all metadata.

Cmdv: yeah sorry I miss read @sgillespie review comment 😅

sgillespie (dismissed):

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -232,6 +233,11 @@ extractSyncOptions snp aop =
     , snapshotEveryLagging = enpSnEveryLagging snp
     }
   where
+    maybeKeepMNames =
+      if null (enpKeepMetadataNames snp)
+        then Strict.Just (enpKeepMetadataNames snp)
+        else Strict.Nothing
```
kderme, line 239: Shouldn't these two cases be reversed?

kderme (approved):
LGTM! Could you squash the fix commits and then we can merge.

PR #1595, opened 2024-01-04
Insert in Bulk

kderme:
# Description

This is an experimental impmenetation of https://github.com/orgs/IntersectMBO/projects/8/views/3?pane=issue&itemId=48804685. The purpose is to benchmark it and evaluate its complexity and correctness.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Util.hs
```diff
@@ -185,15 +189,30 @@ whenMaybe :: Monad m => Maybe a -> (a -> m b) -> m (Maybe b)
...
+      (b, cs') <- f cs a
+      pure (cs <> cs', b : bs)
+
 third :: (a, b, c) -> c
 third (_, _, c) = c
 
-thrd3 :: (a, b, c, d) -> c
-thrd3 (_, _, c, _) = c
+first3 :: (a, b, c) -> a
+first3 (a, _, _) = a
```
sgillespie, line 204: It looks like this is already provided by [extra](https://hackage.haskell.org/package/extra-1.7.14/docs/Data-Tuple-Extra.html#v:first3), which we already [use](https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/cardano-db-sync.cabal#L179)

cardano-db-sync/src/Cardano/DbSync/Util.hs
```diff
@@ -185,15 +189,30 @@ whenMaybe :: Monad m => Maybe a -> (a -> m b) -> m (Maybe b)
...
 
 forth4 :: (a, b, c, d) -> d
 forth4 (_, _, _, d) = d
 
+{-# NOINLINE [1] zipWith3 #-}
+zipWith3 :: (a -> b -> c -> d) -> [a]-> [b] -> [c] -> [d]
+zipWith3 z = go
```
sgillespie, line 211: This also looks to be part of Prelude

PR #1596, opened 2024-01-05
Replace Database Jsonb types with Text

### LINKED ISSUE #1588: Replace jsonb by text
jsonb Postres type provides the ability to query keys of the json. However it's slower to insert. We could use text by default and document ways for users to migrate to json if they do such queries. A command like

```
alter table tx_metadata alter column json type jsonb USING json::jsonb;
```

could be used

Cmdv:
# Description

This fixes #1588

Replacing jsonb types in the database with text. This is to improve insertion speeds/performance.
As part of regression there is a new flag added to allow to put the type constraints back if the user requires it.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


schema/migration-2-0035-20240104.sql
```diff
@@ -0,0 +1,25 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 35 THEN
+    EXECUTE 'ALTER TABLE "tx_metadata" ALTER COLUMN "json" TYPE VARCHAR' ;
```
sgillespie, line 9: Would this make sense to be `text` (no limit)?

cardano-db/src/Cardano/Db/Error.hs
```diff
@@ -34,6 +34,7 @@ data LookupFail
   | DBMultipleGenesis
   | DBExtraMigration !String
   | DBPruneConsumed !String
+  | DBResetJsonb !String
```
sgillespie, line 37: I know you didn't create this, but why not Text here?

schema/migration-2-0035-20240104.sql
Cmdv: guess persistent converts the type `Text` to psql type `VARCHAR`?

cardano-db/src/Cardano/Db/Error.hs
Cmdv: it's because `SqlError` has a `Show` instance and we can just concat the `SqlError` and `LookupFail` without converting `Text -> String`.

schema/migration-2-0035-20240104.sql
kderme: varchar and text are the same under the hood https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying

Cmdv:
Closing this PR as it's been superseded by https://github.com/IntersectMBO/cardano-db-sync/pull/1700

PR #1598, opened 2024-01-08
Test: Hard Fork (Babbage -> Conway) with parameter update proposal

### LINKED ISSUE #1575: Test HF to Conway
There is currently a HF test, but it uses the hardcoded TestForkAtEpoch config value. This has 2 disadvantages: this option will be eventually dropped and it's not that realistic as it's not the real way that HF will happen on mainnet. We could add a new test that forks using the parameter update proposal and in particular the protocol major field. After an adequate amount of these proposals (I think all genesis bootstrap keys need to provide one), there is a hard fork on the next epochy boundary.

In db, this proposals are depicted in `param_proposal.protocol_major` and after the fork in `epoch_param.protocol_major`

sgillespie:
# Description

Fixes #1575. Adds a new conway test the forces a hard fork with a parameter update proposal. Note that, because the configs changed, I had to regenerate all the fingerprints in Conway and Babbage

# Checklist

- [x] Self-reviewed the diff

# Migrations


Cmdv (approved):
LGTM :shipit: 
One question how come most of the fingerprint files changed, is it because of the changes in the config files?

kderme (approved):
Looks good! 
One thing to notice is that the test start from major version 7 (7 and 8 are both Babbage versions) and jumps directly to 9. I didn't know it was possible, but it's not an issue.
```
testing=# select epoch_no, protocol_major from epoch_param;
 epoch_no | protocol_major 
----------+----------------
        1 |              7
        2 |              9
(2 rows)
```

sgillespie:
> LGTM :shipit: One question how come most of the fingerprint files changed, is it because of the changes in the config files?

Correct!

PR #1603, opened 2024-01-12
Fix (CI): Reduce macOS GHA jobs

sgillespie:
# Description

macOS runners in GitHub are scarce (they are limited per-organizatiions), so we need to limit our jobs to one

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -18,8 +18,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
         compiler-nix-name: [ghc810, ghc92, ghc96]
+        include:
+          # We want a single job, because macOS runners are scarce.
+          - os: macos-latest
+            compiler-nix-name: ghc810
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -18,8 +18,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
         compiler-nix-name: [ghc810, ghc92, ghc96]
+        include:
+          # We want a single job, because macOS runners are scarce.
+          - os: macos-latest
+            compiler-nix-name: ghc810
```
angerman, line 26: Don't we want to default to 9.6 here? We want to decommission 8.10 asap. 8.10 is only around for backup in case 9.6 somehow terribly doesnt' work out.

angerman (changes requested):

.github/workflows/haskell.yml
sgillespie: OK, I updated to ghc810 for macOS, and left 810 and 96 for linux. Hopefully I can drop 810 soon

angerman (approved):
LGTM

PR #1607, opened 2024-01-18
optimizations

kderme:
# Description

Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1589, https://github.com/IntersectMBO/cardano-db-sync/issues/1582, https://github.com/IntersectMBO/cardano-db-sync/issues/1584, https://github.com/IntersectMBO/cardano-db-sync/issues/1606

# Checklist

- [X] Commit sequence broadly makes sense
- [X] Commits have useful messages
- [ ] Self-reviewed the diff

# Migrations

- [X] The pr causes a [breaking change](https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/migrations.md) of type a,b or c

The reward migration works on upgrade
The tx_out migation is WIP


schema/migration-2-0035-20240108.sql
```diff
@@ -0,0 +1,29 @@
+-- Persistent generated migration.
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 35 THEN
+    EXECUTE 'alter domain addr29type drop constraint addr29type_check' ;
```
kderme, line 9: When we create these domains
```
CREATE DOMAIN addr29type AS bytea CHECK (octet_length (VALUE) = 29);
```
we don't give any name to the constraint. `addr29type_check` is the name that postgres gives by default.

schema/migration-2-0037-20240117.sql
```diff
@@ -0,0 +1,23 @@
...
+
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 37 THEN
+    EXECUTE 'CREATe TABLE instant_reward AS TABLE reward WITH NO DATA';
+    EXECUTE 'INSERT INTO instant_reward (SELECT * FROM reward WHERE pool_id IS NULL)';
+--    EXECUTE 'ALTER TABLE "instant_reward"("id" SERIAL8  PRIMARY KEY UNIQUE,"addr_id" INT8 NOT NULL,"type" rewardtype NOT NULL,"amount" lovelace NOT NULL,"earned_epoch" INT8 NOT NULL GENERATED ALWAYS AS ((CASE WHEN spendable_epoch >= 1 then spendable_epoch-1 else 0 end)) STORED,"spendable_epoch" INT8 NOT NULL)' ;
```
kderme, line 11: TODO: remove

cardano-chain-gen/test/Test/Cardano/Db/Mock/Validate.hs
```diff
@@ -254,39 +260,46 @@ assertRewardCounts env st filterAddr mEpoch expected = do
...
 
     updateAddrCounters ::
-      Reward ->
+      RewardSource ->
       Maybe (Word64, Word64, Word64, Word64, Word64) ->
       (Word64, Word64, Word64, Word64, Word64)
-    updateAddrCounters reward Nothing = updateCounters reward (0, 0, 0, 0, 0)
-    updateAddrCounters reward (Just cs) = updateCounters reward cs
+    updateAddrCounters tp Nothing = updateCounters tp (0, 0, 0, 0, 0)
+    updateAddrCounters tp (Just cs) = updateCounters tp cs
```
Cmdv, line 267: very minor but I couldn't understand how `tp` was deduced from `RewardSource`

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -304,14 +300,23 @@ share
     addrId              StakeAddressId      noreference
     type                RewardSource        sqltype=rewardtype
     amount              DbLovelace          sqltype=lovelace
-    earnedEpoch         Word64
+    earnedEpoch         Word64 generated="((CASE WHEN (type='refund') then spendable_epoch else (CASE WHEN spendable_epoch >= 2 then spendable_epoch-2 else 0 end) end) STORED)"
```
Cmdv, line 303: is there no way to do this functionality (and bellow) in the codebase? Just incase this functionality gets lost here, failing that a comment as to why it's here and what it does would suffice.

schema/migration-2-0037-20240117.sql
Cmdv: need to do the TODO 😄

Cmdv (commented):
LGTM minor question and suggestions

cardano-db-sync/src/Cardano/DbSync/Era/Util.hs
```diff
@@ -45,8 +45,8 @@ safeDecodeToJson tracer tracePrefix x = do
       pure Nothing
     Right json ->
       -- See https://github.com/IntersectMBO/cardano-db-sync/issues/297
-      if containsUnicodeNul json
+      if False && containsUnicodeNul json
```
sgillespie, line 48: What's this?

kderme: Good catch, I was testing if this check is redundant or not. It's not

sgillespie (approved):

sgillespie:
The hlint check seems to be failing

cardano-db/src/Cardano/Db/Schema.hs
kderme: This functionality exists in Haskell code as well, but is ignored, due to the generated field.

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -376,10 +376,10 @@ mkSyncEnv trce backend connectionString syncOptions protoInfo nw nwMagic systemS
...
   bootstrapVar <- newTVarIO bts
   -- Offline Pool + Anchor queues
-  opwq <- newTBQueueIO 100
-  oprq <- newTBQueueIO 100
-  oawq <- newTBQueueIO 100
-  oarq <- newTBQueueIO 100
+  opwq <- newTBQueueIO 1000
+  oprq <- newTBQueueIO 1000
+  oawq <- newTBQueueIO 1000
+  oarq <- newTBQueueIO 1000
```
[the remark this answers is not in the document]
kderme: Yes, small queues means that fetching thread can block, expecially since this pr extracts from the queue very rarely, so data are expected to pile up.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert.hs
```diff
@@ -197,11 +197,11 @@ insertShelleyBlock syncEnv shouldLog withinTwoMins withinHalfHour blk details is
 
     insertStakeSlice syncEnv $ apStakeSlice applyResult
 
-    when (ioGov iopts)
+    when (ioGov iopts && (withinHalfHour || unBlockNo (Generic.blkBlockNo blk) `mod` 10000 == 0))
```
Cmdv, line 200: `(withinHalfHour || unBlockNo (Generic.blkBlockNo blk) `mod` 10000 == 0)` could probably sit in the where clause

Cmdv (approved):
LGTM minor recommendation

PR #1609, opened 2024-01-19
Move permanent flags to config file

### LINKED ISSUE #1576: Move flags to config file
We have recently added a big number of options https://github.com/input-output-hk/cardano-db-sync/blob/master/doc/configuration.md. These mostly cannot change between separate executions, which means the possibly they need to be part of the config file.

sgillespie:
# Description

Fixes #1576. Move configuration flags that can't be changed (disable-ledger, prune-tx-out, disable-shelley, and so on) and add them to the main configuration file. 

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-chain-gen/test/testfiles/config-conway/test-db-sync-config-bootstrap.json
```diff
@@ -0,0 +1,117 @@
...
+      "scFormat": "ScText",
+      "scKind": "StdoutSK",
+      "scName": "stdout",
+      "scRotation": null
+    }
+  ],
+  "insert_options": {
+    "tx_out": "bootstrap"
+  }
+}
```
Cmdv, line 117: unsure if you are already doing this but could these not be a single file with updated value, here is a simple example:
https://github.com/IntersectMBO/cardano-db-sync/blob/045330da2352504f67f68a0b870cc5e08951810b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/MigrateConsumedPruneTxOut.hs#L60-L65

but for this scenarios it would be: 
```haskell
  where
    mkSynNodeConfig :: IO SyncNodeConfig
    mkSynNodeConfig = do
      -- have 1 init file that is read and parsed
      initConfigFile <- mkSyncNodeConfig conwayConfigDir
      -- update the results with required values for the specific test in above tests 
      -- it's "tx_out": "bootstrap" so that is reflected as so
      pure $ 
        initConfigFile 
          { dncInsertConfig = initInsertConfig { spcTxOut = TxOutBootstrap } }

-- have an init for specific set of tests or all tests if required 
initInsertConfig :: SyncInsertConfig
initInsertConfig = 
  SyncInsertConfig
    { spcTxOut = TxOutEnable
    , spcLedger = LedgerEnable
    , spcShelley = ShelleyDisable
    ...
  }
```

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -199,52 +200,61 @@ getPrunes :: SyncEnv -> Bool
...
+    SyncInsertConfig opts -> opts
+
+fullInsertOptions :: SyncInsertOptions
+fullInsertOptions =
+  SyncInsertOptions
+    { sioTxOut = TxOutEnable
+    , sioLedger = LedgerEnable
+    , sioShelley = ShelleyEnable
+    , sioMultiAsset = MultiAssetEnable
+    , sioMetadata = MetadataEnable
```
Cmdv, line 219: is the consensus that in the presets a user won't be able to add whitelists?

doc/configuration.md
```diff
@@ -6,157 +6,144 @@ features have been added, including historic rewards with Shelley, scripts with
...
+**Full**
 
-### --disable-offchain-pool-data : Experimental
-_(previousle called --disable-offline-data)__
+Enables all other options.
 
-Disables fetching pool offchain metadata.
+**Only UTxO**
+
+This is the equivalent of using `--dont-use-ledger` `--disable-shelley`,
```
Cmdv, line 83: might want to mention that this refers to command line flags we previously had.

> The equates to the following deprecated flags:  `--dont-use-ledger` `--disable-shelley`...

sgillespie: I believe the docs should basically be finished now.

cardano-db-sync/src/Cardano/DbSync/Api.hs
sgillespie: As we discussed earlier, a preset will overwrite _all_ other settings, for the time being

cardano-chain-gen/test/testfiles/config-conway/test-db-sync-config-bootstrap.json
sgillespie: I've written a couple new tests in this style, but the majority I've kept unchanged

kderme:
I had this error, which I think used to be a warning
```
cardano-db-sync -- --config ../cardano-configs/sanchonet/db-sync-config-new.json --socket-path ../db-paths/sancho.socket --state-dir ../ledger-state/sancho-4-0-0-config --schema-dir schema 
cardano-db-sync: Error: If not using --state-dir then make sure to have ledger disabled. For more details view https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/syncing-and-rollbacks.md#ledger-state
```

This seems to hapen with "ledger" : "disable" or with some preset values.

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -199,52 +200,64 @@ getPrunes :: SyncEnv -> Bool
...
+onlyGovInsertOptions =
+  disableAllInsertOptions
+    { sioGovernance = GovernanceConfig True
+    }
+
+disableAllInsertOptions :: SyncInsertOptions
+disableAllInsertOptions =
+  SyncInsertOptions
+    { sioTxOut = TxOutDisable
+    , sioLedger = LedgerDisable
```
kderme, line 252: For `onlyGovInsertOptions` the ledger should be enabled by default.
Ideally the ledger should remain an option even with preset insert configuration. For this pr it would be enough to change the only gov to have `LedgerEnable` by default.

cardano-db-sync/src/Cardano/DbSync/Config/Types.hs
```diff
@@ -200,9 +356,250 @@ parseGenSyncNodeConfig o =
...
+    preset <- obj .:? "preset"
+    case preset :: Maybe Text of
+      Nothing -> SyncInsertConfig <$> parseJSON (Aeson.Object obj)
+      Just "full" -> pure FullInsertOptions
+      Just "only_utxo" -> pure OnlyUTxOInsertOptions
+      Just "only_gov" -> pure OnlyGovInsertOptions
+      Just "disable_all" -> pure DisableAllInsertOptions
+      Just other -> fail $ "unexpected preset: " <> show other
+
+instance ToJSON SyncInsertConfig where
```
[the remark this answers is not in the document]
sgillespie: These `ToJSON` are used for the Hedgehog roundtrip tests

kderme (approved):
Great work!

Issue #1600, opened 2024-01-10
Extend modularity to specified stake address, minting policy, script hash


kderme:
We want to filter only the requested keys

profd2004:
Closing this as part of a triage pass over the older open issues.

Modularity filters for stake addresses, minting policies and script hashes shipped in the config system in 13.3.0.0.

Happy to reopen if a case isn't covered.

PR #1610, opened 2024-01-22
MA Policies Whitelist

Cmdv:
# Description

This is the first part of #1600 adding a whitelist for MA Policies and renaming: 
`KeepMetadataNames` -> `WhitelistMetadataNames`

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/app/cardano-db-sync.hs
```diff
@@ -230,20 +231,28 @@ pSlotNo =
...
-    parseCommaSeparated str =
-      case traverse readMaybe (splitOn "," str) of
-        Just values -> values
-        Nothing -> error "Failed to parse comma-separated values"
+
+pWhiteListMAPolicies :: Parser [Word64]
+pWhiteListMAPolicies =
+  Opt.option
+    (parseCommaSeparated <$> Opt.str)
+    ( Opt.long "whitelist-multi-asset-policy"
```
sgillespie, line 247: I would also put an `Opt.value` here to prevent this option from being mandatory

cardano-db-sync/app/cardano-db-sync.hs
```diff
@@ -230,20 +231,28 @@ pSlotNo =
...
+        <> Opt.help "Only insert a specific sellected list of multi-assets, based on the multi-asset's policy name"
+    )
+
+parseCommaSeparated :: String -> [Word64]
+parseCommaSeparated str =
+  case traverse readMaybe (splitOn "," str) of
+    Just values -> values
+    Nothing -> error "Failed to parse comma-separated values"
```
sgillespie, line 255: Would an explicit exception context make more sense here?

doc/configuration.md
```diff
@@ -155,8 +155,7 @@ Some field are left empty when using this flag, like
 
 Until the ledger state migration happens any restart requires reusing the `--bootstrap-tx-out` flag. After it's completed the flag can be omitted on restarts.
 
-### --keep-tx-metadata
+### --whitelist-tx-metadata
```
sgillespie, line 158: :100:

sgillespie (dismissed):

cardano-db-sync/app/cardano-db-sync.hs
Cmdv: I've deleted these because they are being replaced by your config file PR 😄

cardano-db-sync/src/Cardano/DbSync/Config.hs
```diff
@@ -91,3 +97,61 @@ coalesceConfig pcfg ncfg adjustGenesisPath = do
 
 mkAdjustPath :: SyncPreConfig -> (FilePath -> FilePath)
 mkAdjustPath cfg fp = takeDirectory (pcNodeConfigFilePath cfg) </> fp
+
+-- check both whitelist but also checking plutus Maybes first
+-- TODO: cmdv: unsure if this is correct because if plutusMaybeCheck fails then no multiasset whitelist is not checked
+plutusMultiAssetWhitelistCheck :: SyncEnv -> [Generic.TxOut] -> Bool
+plutusMultiAssetWhitelistCheck syncEnv txOuts =
+  plutusMaybeCheck txOuts && (plutusWhitelistCheck syncEnv txOuts || multiAssetWhitelistCheck syncEnv txOuts)
```
Cmdv, line 105: @kderme is this what you had in mind?

cardano-db-sync/src/Cardano/DbSync/Config.hs
```diff
@@ -88,3 +94,61 @@ coalesceConfig pcfg ncfg adjustGenesisPath = do
...
+    plutuswhitelistCheck whitelist = do
+      any
+        ( isJust
+            . ( \txOut -> do
+                  case (Generic.txOutScript txOut, Generic.maybePaymentCred $ Generic.txOutAddress txOut) of
+                    (Just script, _) ->
+                      if Generic.txScriptHash script `elem` whitelist
+                        then Just txOut
+                        else Nothing
+                    (_, Just address) ->
```
kderme, line 126: Instead of taking cases this should work as an `or` between two check functions

cardano-db-sync/src/Cardano/DbSync/Config.hs
```diff
@@ -88,3 +94,61 @@ coalesceConfig pcfg ncfg adjustGenesisPath = do
...
+plutusMaybeCheck :: [Generic.TxOut] -> Bool
+plutusMaybeCheck =
+  any (\txOut -> isJust (Generic.txOutScript txOut) || isJust (Generic.maybePaymentCred $ Generic.txOutAddress txOut))
+
+plutusWhitelistCheck :: SyncEnv -> [Generic.TxOut] -> Bool
+plutusWhitelistCheck syncEnv txOuts = do
+  -- first check the config option
+  case ioPlutusExtra iopts of
+    PlutusEnable -> True
+    PlutusDisable -> False
```
kderme, line 113: As we discussed Disabled should return `True` at this point, conceptually meaning that the tx is in fact whitelisted. Later, while inserting the plutus data, it will be checked again.

cardano-db-sync/src/Cardano/DbSync/Config.hs
```diff
@@ -88,3 +94,61 @@ coalesceConfig pcfg ncfg adjustGenesisPath = do
...
+              )
+        )
+        txOuts
+
+multiAssetWhitelistCheck :: SyncEnv -> [Generic.TxOut] -> Bool
+multiAssetWhitelistCheck syncEnv txOuts = do
+  let iopts = soptInsertOptions $ envOptions syncEnv
+  case ioMultiAssets iopts of
+    MultiAssetEnable -> True
+    MultiAssetDisable -> False
```
kderme, line 140: Same, `MultiAssetDisable` should return True.

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Config.hs
```diff
@@ -96,10 +96,9 @@ mkAdjustPath :: SyncPreConfig -> (FilePath -> FilePath)
 mkAdjustPath cfg fp = takeDirectory (pcNodeConfigFilePath cfg) </> fp
 
 -- check both whitelist but also checking plutus Maybes first
--- TODO: cmdv: unsure if this is correct because if plutusMaybeCheck fails then no multiasset whitelist is not checked
 plutusMultiAssetWhitelistCheck :: SyncEnv -> [Generic.TxOut] -> Bool
 plutusMultiAssetWhitelistCheck syncEnv txOuts =
-  plutusMaybeCheck txOuts && (plutusWhitelistCheck syncEnv txOuts || multiAssetWhitelistCheck syncEnv txOuts)
+  plutusMaybeCheck txOuts || (plutusWhitelistCheck syncEnv txOuts || multiAssetWhitelistCheck syncEnv txOuts)
```
kderme, line 101: `plutusWhitelistCheck` is called in `plutusMaybeCheck`, so why is it also called here?

Cmdv: I can't see `plutusWhitelistCheck` inside `plutusMaybeCheck`, am I understanding you right?

kderme: Oh sorry I got confused by identation. 

Actually I don't see what's the purpose of `plutusMaybeCheck`. It may return `False` even with `PlutusEnable`, which should never happen. `plutusWhitelistCheck` already does the necessary work.

Cmdv: The was my original comment
> -- TODO: cmdv: unsure if this is correct because if plutusMaybeCheck fails then no multiasset whitelist is not checked

I was trying to make what we spoke about and push the checks as far up the chain and I may have miss understood what you meant and that's why I left this comment there about the `plutusMaybeCheck`. It's there as a check so you don't do any whitelist checks if there isn't any `txOutScript` or `txOutAddress` and that's what it was an `&&` not an `||` so this is:

```
if plutusMaybeCheck txOuts
    then (plutusWhitelistCheck syncEnv txOuts || multiAssetWhitelistCheck syncEnv txOuts)
    else False
```

which is essentially bellow as it will return false straight away if `plutusMaybeCheck txOuts` if false:

```
plutusMaybeCheck txOuts && (plutusWhitelistCheck syncEnv txOuts || multiAssetWhitelistCheck syncEnv txOuts)
```

kderme: I still can't fully understand what does `plutusMaybeCheck` check, since all necessary checks are done in `plutusWhitelistCheck`. Using `or` instead of `and` here will simply give `True` in more cases than we want to.

Cmdv: it was just to save doing big `elem` checks if they're not needed. But it's all good we can just do it inside the whitelist checks.

Cmdv:
closing this PR as it's been superseded by https://github.com/IntersectMBO/cardano-db-sync/pull/1644

PR #1619, opened 2024-02-05
Test to allow custom Config values

### LINKED ISSUE #1618: Test allow custom Config values
With the upcoming change in https://github.com/IntersectMBO/cardano-db-sync/issues/1576 , we are moving command line flags into the config file.

What would be useful is to be able to pass in custom config options via the test.

Cmdv:
# Description

This fixes #1618 
Also multiple reading of the config file is mitigated to just once when db-sync starts.

We can now read a config json file once in the test class that as an init and update it with any custom values required for testing. A single example of it working is:

https://github.com/IntersectMBO/cardano-db-sync/blob/eecc3ee307d425ca29d507b81b4d692615bf85f2/cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/MigrateConsumedPruneTxOut.hs#L38-L65

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
+++ b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
@@ -229,8 +229,8 @@ mkConfig staticDir mutableDir cmdLineArgs config = do
   pure $ Config (Consensus.pInfoConfig pInfoDbSync) pInfoDbSync pInfoForger forging' syncPars
 
 mkSyncNodeConfig :: FilePath -> IO SyncNodeConfig
-mkSyncNodeConfig staticDir =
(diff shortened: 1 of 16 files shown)


sgillespie:
LGTM, but I worry that the list of arguments (runDbSync and withFullConfig and friends) is getting a bit too long

sgillespie (approved):

Cmdv:
@sgillespie I tried to pop in `SyncNodeParams` but that has a `Show` instance which is required by `Logging.Configuration`, so I made a placeholder instance. But then ran into the issue that `SyncNodeParams` is populated via `Parser` monad in `pRunDbSyncNode` so to also add the reading of the json file in that step I couldn't work out how to do.

PR #1635, opened 2024-02-29
Refactor Shelley Insert.hs

Cmdv:
# Description

Refactoring `/Cardano/DbSync/Era/Shelley/Insert.hs` into smaller modules.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/Tx.hs
```diff
@@ -0,0 +1,433 @@
...
+  insertTxOut,
+) where
+
+import Cardano.BM.Trace (Trace)
+import Cardano.Db (DbLovelace (..), DbWord64 (..))
+import qualified Cardano.Db as DB
+import Cardano.DbSync.Api
+import Cardano.DbSync.Api.Types (InsertOptions (..), SyncEnv (..))
+import Cardano.DbSync.Cache.Types (Cache (..))
+
```
sgillespie, line 22: Looks like a superfluous blank line here

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/GovAction.hs
```diff
@@ -0,0 +1,352 @@
...
+  insertCostModel,
+  insertCredDrepHash,
+  insertDrep,
+  insertDrepDistr,
+  insertGovActionProposal,
+  insertParamProposal,
+  insertVotingProcedures,
+  insertVotingAnchor,
+  resolveGovActionProposal,
+  updateEnacted,
```
sgillespie, line 22: Most of this stuff is Conway specific, so I wonder if it would make sense to create an `Era/Conway`

sgillespie (dismissed):

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/Tx.hs
Cmdv: I was splitting between cardano related libraries and standard libraries. Can remove not to worry 👍

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Insert/GovAction.hs
Cmdv: The problem with `insertShelleyBlock` is that upstream it's used for all eras https://github.com/IntersectMBO/cardano-db-sync/blob/15656e42a09d7b9813729e8e284d622c77795714/cardano-db-sync/src/Cardano/DbSync/Default.hs#L155-L182

So technically we should move specific era inserts into the correct era directories and each could be in a single file, rather than the sub modules? eg:
```
Era/Byron/Insert.hs
Era/Conway/Shelley.hs
Era/Conway/Insert.hs
```
Then have `insertShelleyBlock` moved and renamed to a module say `Cardano.DbSync.Insert`

sgillespie (approved):
LGTM

PR #1636, opened 2024-03-03
Make it build with ghc-9.8

erikd:
# Description

Make it build with ghc-9.8

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        compiler-nix-name: [ghc810, ghc96]
+        compiler-nix-name: [ghc810, ghc96, ghc98]
         include:
           # We want a single job, because macOS runners are scarce.
           - os: macos-latest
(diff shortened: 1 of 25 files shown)


sgillespie (approved):

Cmdv (approved):

Cmdv (changes requested):
looks like it's now failing tests in the CI 😢 
This will need investigation as it's hard to know why `-Wno-x-partial` would cause incorrect functionality.

erikd:
> This will need investigation as it's hard to know what `-Wno-x-partial` would cause incorrect functionality.

I can't see how  `-Wno-x-partial` could possibly cause changes in behavior. All it does is turn off a warning.

Cmdv:
> I can't see how -Wno-x-partial could possibly cause changes in behaviour. All it does is turn off a warning.
unless the ghc98 CI is doing something different, I'm just trying to re run the tests now. It's compiling fine just the tests fail in various ways only on ghc98 but passes fine in ghc96 ghc810 (I'm not looking at the ci/hydra ones they always seem to fail). Hence something need to be investigated makes no sense the tests would all pass in other ghc versions 🤷

Cmdv (approved):

sgillespie (dismissed):

sgillespie (approved):

Cmdv:
🎉

PR #1648, opened 2024-03-08
Prepare sancho-4-1-0

kderme:
# Description

Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1571, https://github.com/IntersectMBO/cardano-db-sync/issues/1633, https://github.com/IntersectMBO/cardano-db-sync/issues/1637, https://github.com/IntersectMBO/cardano-db-sync/issues/1646

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/GovAction.hs
```diff
@@ -124,17 +128,39 @@ insertGovActionProposal cache blkId txId govExpiresAt (index, pp) = do
...
+    insertNewCommitteeInfo gapId q =
+      void . DB.insertNewCommitteeInfo $
+        DB.NewCommitteeInfo
+          { DB.newCommitteeInfoGovActionProposalId = gapId
+          , DB.newCommitteeInfoQuorumNumerator = fromIntegral $ numerator r
+          , DB.newCommitteeInfoQuorumDenominator = fromIntegral $ denominator r
           }
       where
         r = unboundRational q -- TODO work directly with Ratio Word64. This is not currently supported in ledger
+    insertMembers gapId removed added q = do
```
Cmdv, line 150: not sure if this is a diff view error but doesn't look like these following 2 functions are indented correctly?

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/GovAction.hs
```diff
@@ -124,17 +128,39 @@ insertGovActionProposal cache blkId txId govExpiresAt (index, pp) = do
...
+      whenJust mmCommittee $ \mCommittee -> do
+        -- Nothing means we're not in Conway so it can't happen.
+        let committee = updatedCommittee removed added q mCommittee
+        mapM_ (insertNewMember gapId) (Map.toList $ committeeMembers committee)
+
+    insertNewMember gapId (cred, e) = do
+      chId <- insertCommitteeHash cred
```
Cmdv, line 157: Would there ever be a situation where inserting the Committee hash would fail and we need to handle it?

cardano-db-sync/src/Cardano/DbSync/Fix/PlutusScripts.hs
```diff
@@ -109,18 +109,18 @@ findWrongPlutusScripts tracer =
...
+      bytes <- maybeToEither (Just "No bytes found for plutus script") id $ DB_V_13_0.scriptBytes dbScript
+      case DB_V_13_0.scriptType dbScript of
+        PlutusV1 -> do
+          -- The bug only affected Alonzo script
+          let script :: AlonzoScript StandardAlonzo = PlutusScript (AlonzoPlutusV1 (Plutus $ PlutusBinary $ SBS.toShort bytes))
+          let hsh :: Ledger.ScriptHash StandardCrypto = Ledger.hashScript @StandardAlonzo script
+          Right $ Generic.unScriptHash hsh
+        PlutusV2 -> Left Nothing
+        PlutusV3 -> Left Nothing
+        _ -> Left $ Just "Non plutus script found where it shouldn't."
```
Cmdv, line 123: not 100% clear on the message, does `A non-Plutus script was discovered in an unexpected location.` sound better?

Cmdv (commented):
LGTM just a couple of questions

sgillespie (approved):

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/GovAction.hs
kderme: It looks good to me. `insertNewCommittee`, `insertNewCommitteeInfo` and `insertMembers` are defined in the same level. The first calls the other 2.

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/GovAction.hs
kderme: Internally it uses `insertCheckUnique`, so if the entry already exists, it will return the existing id.

Cmdv: perfect thanks

Cmdv (approved):

cardano-db-sync/src/Cardano/DbSync/Fix/PlutusScripts.hs
kderme: Leaving this as is for now. This comment existed before, but was commended out here https://github.com/IntersectMBO/cardano-db-sync/pull/1648/files/5f2881e04587c2a7cd25f695dfd5c4a3b453cc3a#diff-23041e6f296ec26493d184b4167eb2d3070f4483fcbfe1d5048bef429b209f3bL123

PR #1687, opened 2024-04-29
Save the DRep Pulser State after computing it

### LINKED ISSUE #1599: Replace the evaluated drep pulser in ledger state
In Conway era, at the first block of each epoch, DBSync uses `finishDRepPulser` to force the computation of the drep snapshots. It extractthe information it needs and stores it in the db. To avoid recomputing the drep pulser we can adjust the ledger state by replacing the `newEpochStateDRepPulsingStateL` with the computed version. ledger provides `forceDRepPulsingState` which can be used for this.

sgillespie:
# Description

Fixes #1599. The bulk of this work is done in the `HasNewEpochState` typeclass, which allows us to update the ledger state in `ExtLedgerState`. 

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Ledger/Types.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Ledger/Types.hs
@@ -218,9 +218,45 @@ instance HasNewEpochState StandardShelley where
     hApplyExtLedgerState $
       fn (applyNewEpochState' st) :* fn id :* fn id :* fn id :* fn id :* fn id :* Nil
 
+instance HasNewEpochState StandardAllegra where
+  getNewEpochState st = case ledgerState st of
+    LedgerStateAllegra allegra -> Just (shelleyLedgerState allegra)
+    _ -> Nothing
+
+  applyNewEpochState st =
+    hApplyExtLedgerState $
+      fn id :* fn (applyNewEpochState' st) :* fn id :* fn id :* fn id :* fn id :* Nil
+
+instance HasNewEpochState StandardMary where
+  getNewEpochState st = case ledgerState st of
+    LedgerStateMary mary -> Just (shelleyLedgerState mary)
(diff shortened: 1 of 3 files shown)


cardano-db-sync/src/Cardano/DbSync/Ledger/Types.hs
```diff
@@ -193,3 +198,97 @@ instance Anchorable (WithOrigin SlotNo) CardanoLedgerState CardanoLedgerState wh
   getAnchorMeasure _ = getTipSlot . clsState
 
 data SnapshotPoint = OnDisk LedgerStateFile | InMemory CardanoPoint
+
+-- | Per-era pure getters and setters on @NewEpochState@. Note this is a bit of an abuse
+-- of the cardano-ledger/ouroboros-consensus public APIs, because ledger state is not
+-- designed to be updated this way. We are only replaying the chain, so this should be
+-- safe.
+class HasNewEpochState era where
```
kderme, line 206: That's a very useful class to have, which can be reused for https://github.com/orgs/IntersectMBO/projects/8/views/3?pane=issue&itemId=60533809

kderme (dismissed):
Great work!
Good to merge when it's tested on preprod or preview.

sgillespie:
> Great work! Good to merge when it's tested on preprod or preview.

It can't be tested in preprod or preview right?

Also, I forgot to update the comment, I have done some basic manual testing in sanchonet

kderme:
> It can't be tested in preprod or preview right?

Right, it would only work on sanchonet after rebasing to master

sgillespie:
> Right, it would only work on sanchonet after rebasing to master

Yes, this is how I was able to test it on sanchonet. I applied my changes on top of `sancho-4-2-1`

Cmdv (approved):

PR #1704, opened 2024-05-16
1685 - expired epoch fix

### LINKED ISSUE #1685: retired epoch of gov_action_proposal is wrong
It's one epoch before it's actually retired. This should be fixed and the rollback should be adjusted.

Cmdv:
# Description

this fixes #1685 

The function adjusted is inserted on new epoch so will also work for rollbacks.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


kderme:
@Cmdv does this work fine on sanchonet?

Cmdv:
> @Cmdv does this work fine on sanchonet?

@kderme Does this look right?
SCR-20240516-jnt

kderme:
What query produced the first one? In general I find text easier to read than screenshots. I would assume that dropped and expired epochs should match.

Cmdv:
@kderme right this should have done the trick:

```
select id, dropped_epoch, expired_epoch, expiration from gov_action_proposal;

+----+---------------+---------------+------------+
| id | dropped_epoch | expired_epoch | expiration |
|----+---------------+---------------+------------|
| 5  | 311           | <null>        | 314        |
| 23 | 316           | 315           | 315        |
| 12 | 316           | 315           | 315        |
| 6  | 315           | 314           | 314        |
| 7  | 315           | 314           | 314        |
| 8  | 315           | 314           | 314        |
| 9  | 315           | 314           | 314        |
| 10 | 315           | 314           | 314        |
```

kderme (approved):

PR #1705, opened 2024-05-16
update node 8.11

kderme:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db/Insert.hs
```diff
@@ -214,7 +214,7 @@ insertManyRewardRests ::
   (MonadBaseControl IO m, MonadIO m) =>
   [RewardRest] ->
   ReaderT SqlBackend m ()
-insertManyRewardRests = insertManyCheckUnique "Many Rewards Rest"
+insertManyRewardRests = void . insertMany' "Many Rewards Rest"
```
Cmdv, line 217: Is it worth catching any failures?

cardano-db-sync/src/Cardano/DbSync/Config/Conway.hs
```diff
@@ -36,11 +41,27 @@ readConwayGenesisConfig ::
...
+        , ucppDRepVotingThresholds = def
+        , ucppCommitteeMinSize = 0
+        , ucppCommitteeMaxTermLength = EpochInterval 0
+        , ucppGovActionLifetime = EpochInterval 0
+        , ucppGovActionDeposit = Coin 0
+        , ucppDRepDeposit = Coin 0
+        , ucppDRepActivity = EpochInterval 0
+        , ucppMinFeeRefScriptCostPerByte = minBound
+        , ucppPlutusV3CostModel = case mkCostModel PlutusV3 (replicate 233 0) of
+            Left err -> error $ "Failed to create default Plutus V3 CostModel " <> show err
```
Cmdv, line 62: `logAndThrowIO` inside of `Error.hs` allows for better error handling and messaging.

cardano-db/src/Cardano/Db/Insert.hs
kderme: It cannot fail due to a unique key, since it was removed. So I thjink it's fine

cardano-db-sync/src/Cardano/DbSync/Config/Conway.hs
kderme: The problem is `logAndThrowIO` works in IO. This will anyway fail very early if it's wrong and it's just when Conway Genesis is ignored.

Cmdv: cool that makes sense 👍

Cmdv (approved):

PR #1718, opened 2024-05-30
Add EpochState table

kderme:
# Description

Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1706, https://github.com/IntersectMBO/cardano-db-sync/issues/1711, https://github.com/IntersectMBO/cardano-db-sync/issues/1715

# Checklist

- [ ] Self-reviewed the diff

# Migrations

It has a migration and it only affects Conway schema

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
@@ -228,7 +228,7 @@ applyBlock env blk = do
     let (ledgerEvents, deposits) = splitDeposits ledgerEventsFull
     let !newLedgerState = finaliseDrepDistr (lrResult result)
     !details <- getSlotDetails env (ledgerState newLedgerState) time (cardanoBlockSlotNo blk)
-    !newEpoch <- fromEitherSTM $ mkNewEpoch (clsState oldState) newLedgerState (findAdaPots ledgerEvents)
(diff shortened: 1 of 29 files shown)


cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
```diff
@@ -289,10 +292,11 @@ applyBlock env blk = do
     finaliseDrepDistr ledger =
       ledger & newEpochStateT %~ forceDRepPulsingState @StandardConway
 
-getEnacted :: ExtLedgerState CardanoBlock -> Maybe (GovRelation StrictMaybe StandardConway)
+getEnacted :: ExtLedgerState CardanoBlock -> Maybe (ConwayGovState StandardConway)
 getEnacted ls = case ledgerState ls of
   LedgerStateConway cls ->
-    Just $ govStatePrevGovActionIds $ Consensus.shelleyLedgerState cls ^. Shelley.newEpochStateGovStateL
+    Just $ Consensus.shelleyLedgerState cls ^. Shelley.newEpochStateGovStateL
+  --    Just $ govStatePrevGovActionIds $ Consensus.shelleyLedgerState cls ^. Shelley.newEpochStateGovStateL
```
sgillespie, line 299: This looks like it can be removed

Cmdv (approved):
LGTM  though it looks like the forced index tests are failing

PR #1719, opened 2024-06-02
Convert Handling SyncEnv via an MTL

Cmdv:
# Description

Just had a go at converting handling our `SyncEnv` via an MTL.
Been having trouble when trying to log something whilst debugging and it always involves passing along `SyncEnv` through all the functions. Also it can be quite a pain to pass `SyncEnv` + `Trace IO Text` all the time:

It's quite nice to use, here's an example:
```
fnc :: App (Bool)
fnc = do
  syncEnv <- ask
  epochState <- asks envEpochState
  pure True
```

Got a bunch of helper functions that can be used when in the `App` monad:

```
-- | Get `Trace` from the `SyncEnv`.
askTrace :: App (Trace IO Text)

-- | Get `Ledger.Network` from the `SyncEnv`.
askNetwork :: App Ledger.Network

-- | Get `InsertOptions` from the `SyncEnv`.
askInsertOptions :: App InsertOptions
```
There's one error I can't work out how it happened but in `/Config/Cardano.hs` saying that there is an instance missing when trying to use `protocolInfoCardano`:

```
mkProtocolInfoCardano ::
  GenesisConfig ->
  [Consensus.ShelleyLeaderCredentials StandardCrypto] -> -- this is not empty only in tests
  (ProtocolInfo CardanoBlock, IO [BlockForging IO CardanoBlock])
mkProtocolInfoCardano genesisConfig shelleyCred =
  protocolInfoCardano
    CardanoProtocolParams
      { paramsByron =

  ```
  error:
  
  ```
  No instance for (Ouroboros.Consensus.Ledger.SupportsProtocol.LedgerSupportsProtocol
                         (Ouroboros.Consensus.Shelley.Ledger.Block.ShelleyBlock
                            (Ouroboros.Consensus.Protocol.TPraos.TPraos StandardCrypto)
                            (cardano-ledger-shelley-1.11.0.0:Cardano.Ledger.Shelley.Era.ShelleyEra
                               StandardCrypto)))
        arising from a use of ‘protocolInfoCardano’
...

### CODE REVIEW

File: cardano-db-sync/src/Cardano/DbSync/AppT.hs [hunk 1]
```diff
@@ -0,0 +1,294 @@
...
+
+import Cardano.BM.Trace (Trace, logError)
+import qualified Cardano.Db as DB
+import Cardano.DbSync.Cache.Types (CacheState)
+import Cardano.DbSync.Config.Types (LedgerStateDir, SyncNodeConfig)
+import Cardano.DbSync.Error.Types (SyncNodeError)
+import Cardano.DbSync.Ledger.Types (CardanoLedgerState, LedgerDB)
+import Cardano.DbSync.Types (CardanoBlock, CardanoInterpreter, OffChainPoolResult, OffChainPoolWorkQueue, OffChainVoteResult, OffChainVoteWorkQueue)
+import qualified Cardano.Ledger.Shelley.API as Ledger
+import Cardano.Prelude (
```
sgillespie, line 49: If we're using `NoImplicitPrelude`, we might as well just import all of `Cardano.Prelude`. Better yet, we might consider adding it our custom prelude implicitly with [mixins](https://kowainik.github.io/projects/relude#mixins)

cardano-db-sync/src/Cardano/DbSync/AppT.hs
```diff
@@ -0,0 +1,294 @@
...
+import Prelude (Show (..), error, (++), (.))
+
+-- | The application monad parameterized over `IO`.
+type App = AppT IO
+
+-- | The application monad transformer.
+newtype AppT m a = AppT
+  { runAppT :: ReaderT SyncEnv (ExceptT SyncNodeError m) a
+  }
+  deriving newtype (Functor, Applicative, Monad, MonadIO, MonadReader SyncEnv, MonadError SyncNodeError)
```
sgillespie, line 99: :ok_hand:

cardano-db-sync/src/Cardano/DbSync/AppT.hs
```diff
@@ -0,0 +1,294 @@
...
+  case result of
+    Left err -> error $ "Error occurred: " ++ show err
+    Right val -> pure val
+
+-- | Lift an `App` action to fit the required type for `runDbIohkNoLogging`
+runAppWithNoLogging ::
+  MonadUnliftIO m =>
+  SyncEnv ->
+  App a ->
+  ReaderT SqlBackend (NoLoggingT m) a
```
sgillespie, line 232: Another option here is to use the `SqlPersistT` [alias](https://hackage.haskell.org/package/persistent-2.14.6.1/docs/Database-Persist-Sql.html#t:SqlPersistT)

cardano-db-sync/src/Cardano/DbSync/AppT.hs
Cmdv: yeah I agree, there are lots of places where it uses prelude and others `Cardano.Prelude` I'm happy with the later and having `NoImplicitPrelude` globally.

cardano-db-sync/src/Cardano/DbSync/AppT.hs
Cmdv: Good catch, didn't even look for it

Cmdv: could we do this as part of another PR just that all of the db queries are like this:

```
insertAdaPots :: (MonadBaseControl IO m, MonadIO m) => AdaPots -> ReaderT SqlBackend m AdaPotsId
insertAdaPots = insertUnchecked "AdaPots"

insertBlock :: (MonadBaseControl IO m, MonadIO m) => Block -> ReaderT SqlBackend m BlockId
insertBlock = insertUnchecked "Block"
```

sgillespie: Yes, of course! Let's keep this PR as small as is reasonable

Issue #1193, opened 2022-07-12
Transaction witness information


johnnuy:
I would like to be able to query information about the witnesses that signed a given transaction.  This information is apparently available in the raw CBOR but is not exposed through db-sync as of now.

chbfiv:
On Tue, Jul 12, 2022 at 2:56 PM johnnuy ***@***.***>
wrote:

> I would like to be able to query information about the witnesses um by ct
> be nq  that signed a given transaction.jib  in k to 😯 uvy  This
> information is apparently available in the raw CBOR but is not exposed
> through db-sync I as qmN ijvuvbiiu in t
>
> Message ID: ***@***.***>
>
n thanks ourf  Bbj--
Bourke Floyd

bgdsh:
+1 on this issue

profd2004:
@johnnuy could you narrow down which witness data you actually need: vkey and signature pairs, native script witnesses, Plutus script witnesses, bootstrap witnesses, or all of it? Two things exist now that did not in 2022, as `extra_key_witness` covers required signers and the `tx_cbor` insert option stores the raw transaction bytes so witnesses can be decoded client-side. Knowing whether decoding from `tx_cbor` would be good enough, or whether you need indexed queryable columns, is what decides the shape of this one.

profd2004:
@johnnuy closing this for lack of a reply. We asked on 7 August which witness data you actually need. Two things exist now that did not in 2022: `extra_key_witness` covers required signers, and the `tx_cbor` insert option stores the raw transaction bytes so witnesses can be decoded client-side. If decoding from `tx_cbor` is not enough and you need indexed queryable columns, comment with the specific witness types and the queries you want to run, and we will reopen.

PR #1723, opened 2024-06-04
Store CBOR serialized transactions in `tx_cbor` table

### LINKED ISSUE #1701: Option to enable raw CBOR collection
It would be very helpful for cardano-db-sync to enable us to collect raw cbor into a table. This would need a new option, something like `--enable-cbor-collection` to now increase the size of the database as I imagine most of the users won't use this feature.

There are other tools that provide this, but it is helpful to have it within db-sync so that we are present consistent data to end users.

sorki:
# Description

Optional feature that can be enabled via insert options like so:
```
 "insert_options": {
    "tx_cbor": "enable",
```

Testing on preview network, with minimal insert options
```
  "insert_options": {
    "tx_cbor": "enable",
    "tx_out": { "value": "disable" },
    "ledger": "disable",
    "shelley": { "enable": false },
    "multi_asset": { "enable": false },
    "metadata": { "enable": false },
    "plutus": { "enable": false },
    "governance": "disable",
    "offchain_pool_data": "disable",
    "json_type": "disable"
  }
```

results in about `6.2GB` of data
```
csyncdb=> SELECT pg_size_pretty(pg_total_relation_size('"public"."tx_cbor"'));
 pg_size_pretty
----------------
 6272 MB
(1 row)
```

Closes #1701

# Checklist

- [x] Self-reviewed the diff

# Migrations


schema/migration-4-0003-20210116.sql
```diff
@@ -26,6 +26,7 @@ CREATE INDEX IF NOT EXISTS idx_tx_in_tx_in_id ON tx_in(tx_in_id) ;
 CREATE INDEX IF NOT EXISTS idx_tx_in_tx_out_id ON tx_in(tx_out_id) ;
 CREATE INDEX IF NOT EXISTS idx_tx_metadata_tx_id ON tx_metadata(tx_id) ;
 CREATE INDEX IF NOT EXISTS idx_tx_out_tx_id ON tx_out(tx_id) ;
+CREATE INDEX IF NOT EXISTS idx_tx_cbor_tx_id ON tx_cbor(tx_id) ;
```
Cmdv, line 29: It might be that you have to generate a new migration file using [these instructions](https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/schema-management.md#creating-a-migration)

Cmdv (commented):
Thanks this is great, I left a little comment about the migration file. Also seems it's having a bit of a fit on mac related tests. On next push I'll take a look see if it persists. 👍

schema/migration-4-0003-20210116.sql
sorki: Added as separate file and updated counts in `cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/ForceIndex.hs` as well.

cardano-db-sync/src/Cardano/DbSync/Era/Byron/Insert.hs
```diff
@@ -237,24 +237,37 @@ insertByronTx syncEnv blkId tx blockIndex = do
...
+            , DB.txScriptSize = 0
+            }
+
+      when (ioTxCBOR iopts) $ do
+        void
+          . lift
+          . DB.insertTxCBOR
+          $ DB.TxCbor
+            { DB.txCborTxId = txId
+            , DB.txCborBytes = serialize' tx
```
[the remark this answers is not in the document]
sorki: Don't think so as there's `instance ToCBOR TxAux` but I've switched this to do `serialize' $ Byron.taTx tx` similar to TX size calculation above. Looks like this will omit witnesses (and annotations) but Byron TXs are not that important anyway and there's [another issue for storing TX wits]( #1193) which I've might try to tackle at some point.

sorki:
Added a rollback deletion, schema documentation, changelog entry and hopefully resolved review issues / tests.

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Types.hs
```diff
@@ -49,6 +49,7 @@ import Ouroboros.Consensus.Cardano.Block (StandardAlonzo, StandardBabbage, Stand
 data Tx = Tx
   { txHash :: !ByteString
   , txBlockIndex :: !Word64
+  , txCBOR :: !ByteString
```
kderme, line 52: Let's leave this lazy
```suggestion
  , txCBOR :: ByteString
```

kderme (dismissed):
One small comment to address and good to go

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Generic/Tx/Types.hs
sorki: I agree, fixed!

Cmdv (approved):
Very good, thanks for the contribution 😄

Cmdv (approved):

kderme (approved):

kderme:
Merged at https://github.com/IntersectMBO/cardano-db-sync/pull/1749. Thank you

PR #1731, opened 2024-06-12
1728 - stake address cache

### LINKED ISSUE #1728: Improve stake address Cache
The analysis at https://github.com/IntersectMBO/cardano-db-sync/issues/1721 shows that there are many misses in the cache about stake addresses, for addresses that are used but never registered. One solution is to try and cache stake addresses always, by changing the `DontCacheNew` to `CacheNew`. This can have an impact in memory usage. A banchmark which shows the eventual rss memory used and the [DbBadger](https://github.com/darold/pgbadger) results for the current master and the modified branch would be very useful

Cmdv:
# Description

This fix introduces a reduction in Select queries to the stake address database and instead do them locally in an LRU cache with a size of 1,500,000 entries. This turned out to be a good size to keep RAM usage low yet have a fairly high hit rate. Whilst there I also increased the MultiAsset LRU cache capacity to `250000` which has resulted in.

epoch taking around 40 mins to complete syncing down to 10-15mins!!!

| number of queries | Before | After |
| :---         | :---     |  :---    |
| stake address  | 2,748,224 | 66,834 |
| multi asset | 2,180,423 | 1,131,696 |

download

### New Statistics
```
Cache Statistics:
  Stake Addresses: cache size: 1500000, hit rate: 98%, hits: 3135062, misses: 61260
  Pools: cache size: 3128, hit rate: 99%, hits: 2485150, misses: 3128
  Datums: cache capacity: 250000, cache size: 75820, hit rate: 40%, hits: 87835, misses: 129899
  Multi Assets: cache capacity: 250000, cache size: 250000, hit rate: 73%, hits: 3042122, misses: 1099061
  Previous Block: hit rate: 49%, hits: 21001, misses: 21003
```

The previous cache used which is a `Map` has a `cache size: 1,273,020` so there is potential of increasing size of new cache.

When db-sync is restarted the cache will be populated with the last inserted stake address into the db (using the chosen capacity of the cache).

All newly inserted stake address are inserted into the cache

This fixes #1728


Cmdv:
@kderme With Multi Asset cache all I did was to up the cache capacity from `50000` to `250000`. I've not actually investigating what is happening. 

In regards to misses, I have a feeling it's not quite working as expected will double check they are all being marked correctly. As you say the misses and the number of queries the db report should be the same. `5,574` seem to be not marked from the  db-sync side 🤔

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Block.hs
```diff
@@ -98,9 +98,9 @@ insertBlockUniversal syncEnv shouldLog withinTwoMins withinHalfHour blk details
     blockGroupedData <- foldM (\gp (idx, tx) -> txInserter idx tx gp) mempty zippedTx
     minIds <- insertBlockGroupedData syncEnv blockGroupedData
 
-    -- now that we've inserted the Block and all it's txs lets cache what we'll need
+    -- now that we've inserted the Block and all it's txs lets cacheStatus what we'll need
     -- when we later update the epoch values.
-    -- if have --dissable-epoch && --dissable-cache then no need to cache data.
+    -- if have --dissable-epoch && --dissable-cacheStatus then no need to cacheStatus data.
```
kderme, line 103:
```suggestion
    -- if have --disable-epoch && --dissable-cache then no need to cacheStatus data.
```

sgillespie (approved):
LGTM, thanks for the extra comments!

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -382,7 +383,16 @@ mkSyncEnv ::
   IO SyncEnv
 mkSyncEnv trce backend connectionString syncOptions protoInfo nw nwMagic systemStart syncNodeConfigFromFile syncNP ranMigrations runMigrationFnc = do
   dbCNamesVar <- newTVarIO =<< dbConstraintNamesExists backend
-  cache <- if soptCache syncOptions then newEmptyCache 250000 50000 else pure useNoCache
+  cache <-
+    if soptCache syncOptions
+      then
+        newEmptyCache
+          LRUCacheCapacity
+            { lirCapacityStakeHashRaw = 1500000
```
kderme, line 391: From a recent full sync to mainnet
```
[db-sync-node:Info:79] [2024-06-19 15:03:48.27 UTC] 
Cache Statistics:
  Stake Addresses: cache size: 1664030, hit rate: 89%, hits: 656453661, misses: 73037617
  Pools: cache size: 5878, hit rate: 99%, hits: 536552899, misses: 472490
  Datums: cache capacity: 50000, cache size: 50000, hit rate: 42%, hits: 19769700, misses: 26260235
  Multi Assets: cache capacity: 250000, cache size: 250000, hit rate: 75%, hits: 715052104, misses: 234336378
  Previous Block: hit rate: 49%, hits: 9554635, misses: 9554637
[db-sync-node:Info:79] [2024-06-19 15:03:48.27 UTC] Starting epoch 492
```
I would say we should increase it to 2000000

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -83,115 +83,133 @@ getCacheStatistics cs =
...
+              pure $ Right addrId
+            _other -> do
+              liftIO $ atomically $ writeTVar (cStakeRawHashes ci) lruCache
+              pure $ Right addrId
+        Nothing -> do
+          queryRes <- mapLeft (,bs) <$> queryStakeAddress bs
+          liftIO $ missCreds (cStats ci)
+          case queryRes of
+            Left _ -> pure queryRes
+            Right stakeAddrsId -> do
```
kderme, line 201: I think we should update the cache only if `UpdateCache` is set.

sgillespie (dismissed):

cardano-db-sync/src/Cardano/DbSync/Api.hs
Cmdv: I'll check how much this increases our RAM usage as it might be slightly too much.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -123,75 +125,85 @@ insertStakeAddress rewardAddr stakeCredBs =
...
+              when (isCacheActive cache) $
```
kderme, line 192: This code only runs if the cache is Active. We actually want to check that the specific request wants to update the cache

> I think we should update the cache only if UpdateCache is set.

Cmdv: sorry misunderstood 🤦‍♀️

cardano-db-sync/src/Cardano/DbSync/Api.hs
Cmdv: Just been syncing with `mainnet` and epoch it's jumped back up to 40 mins, I think it's because each lookup on the LRU is more expensive due to the increase of the cache capacity

PR #1737, opened 2024-06-17
Fix committee and indexes

kderme:
# Description

Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1735. Also fixes the indexes
It also cherry-picks some forgotten commits from release 13.2.0.2

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Query.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Query.hs
@@ -1,68 +1,29 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 
 module Cardano.DbSync.Era.Shelley.Query (
-  queryPoolHashId,
-  queryStakeAddress,
-  queryStakeRefPtr,
+  resolveStakeAddress,
   resolveInputTxId,
   resolveInputTxOutId,
   resolveInputValue,
   resolveInputTxOutIdValue,
   queryResolveInputCredentials,
-  queryPoolUpdateByBlock,
 ) where
 
 import Cardano.Db
 import qualified Cardano.DbSync.Era.Shelley.Generic as Generic
 import Cardano.DbSync.Util
-import Cardano.Ledger.BaseTypes (CertIx (..), TxIx (..))
-import Cardano.Ledger.Credential (Ptr (..))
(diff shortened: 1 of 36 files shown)


cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Query.hs
```diff
@@ -82,41 +44,3 @@ resolveInputTxOutIdValue txIn =
...
-    -- Need to order by DelegationSlotNo descending for correct behavior when there are two
-    -- or more delegation certificates in a single epoch.
-    orderBy [desc (blk ^. BlockSlotNo)]
-    limit 1
-    pure (sr ^. StakeRegistrationAddrId)
-  pure $ unValue <$> listToMaybe res
-
--- Check if there are other PoolUpdates in the same blocks for the same pool
-queryPoolUpdateByBlock :: MonadIO m => BlockId -> PoolHashId -> ReaderT SqlBackend m Bool
-queryPoolUpdateByBlock blkId poolHashId = do
```
Cmdv, line 109: 👌 I did find it confusing that we had db queries inside of cardano-db-sync but was too scared to move 😅

Cmdv (changes requested):
LGTM just a simple export might need removing

sgillespie:
Looks like you have some tests to update:

```
force-indexes
        check force-index adds indexes:                                          FAIL (51.21s)
          test/Test/Cardano/Db/Mock/Validate.hs:153:
          unexpected number of indexes: got 171 expected 172
          Use -p '/Conway unit tests.Command Line Arguments.force-indexes.check force-index adds indexes/' to rerun this test only.
        check no force-index doesn't add indexes:                                FAIL (51.22s)
          test/Test/Cardano/Db/Mock/Validate.hs:153:
          unexpected number of indexes: got 108 expected 107
          Use -p '/Conway unit tests.Command Line Arguments.force-indexes.check no force-index doesn'\''t add indexes/' to rerun this test only.
```

cardano-db-sync/src/Cardano/DbSync/Ledger/Types.hs
```diff
@@ -143,7 +142,7 @@ data ApplyResult = ApplyResult
   , apSlotDetails :: !SlotDetails
   , apStakeSlice :: !Generic.StakeSliceRes
   , apEvents :: ![LedgerEvent]
-  , apCommittee :: !(Maybe (StrictMaybe (Committee StandardConway)))
+  , apGovActionState :: !(Maybe (ConwayGovState StandardConway))
```
sgillespie, line 145: :ok_hand: This will come in handy for testing, too

Cmdv (commented):
LGTM just a couple of little things. Also do we need documentation for `EnableFutureGenesis` in the docs?

kderme:
Done at https://github.com/IntersectMBO/cardano-db-sync/pull/1744

PR #1743, opened 2024-06-26
Add treasury donation to tx table

### LINKED ISSUE #1732: Treasury donations
With Conway it's possible to send money to the treasury. We should capture this in a new field in `tx` which is by default and for previous entries 0. Also we should take this into account in out tx fee calculation.

sgillespie:
# Description

Fixes #1732 

# Checklist

- [x] Any changes are noted in the [changelog](https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)
- [x] Self-reviewed the diff

# Migrations


Cmdv:
looks like theres failures when building the tests 😞

kderme (approved):
Just one comment, seems good otherwise

schema/migration-2-0044-20240626.sql
```diff
@@ -0,0 +1,19 @@
+-- Persistent generated migration.
```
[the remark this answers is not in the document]
sgillespie: Done!

sgillespie:
> looks like theres failures when building the tests 😞

Fixed!

kderme (approved):
LGTM! I left a comment about some missing functionality. We could merge this and do it on a separate pr though.

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/Tx.hs
```diff
@@ -77,6 +77,7 @@ insertTx syncEnv isMember blkId epochNo slotNo applyResult blockIndex tx grouped
   let !mdeposits = if not (Generic.txValidContract tx) then Just (Coin 0) else lookupDepositsMap txHash (apDepositsMap applyResult)
   let !outSum = fromIntegral $ unCoin $ Generic.txOutSum tx
       !withdrawalSum = fromIntegral $ unCoin $ Generic.txWithdrawalSum tx
+      !treasuryDonation = fromIntegral . unCoin $ Generic.txTreasuryDonation tx
```
kderme, line 80: IIrc we wanted to use this in the deposit calculation for a specifc case.

sgillespie: Thanks for the reminder, I just updated it to be subtracted from the inputs. However, I couldn't find a way to test it, so please carefully re-review

kderme (approved):
LGTM!

PR #1748, opened 2024-07-01
Add better docs on setting up postgresql db

### LINKED ISSUE #1747: Add better documentation on setting up postgresql db
It isn't too clear how to setup the postgresql database as it's nested within the documentation:

```shell
git clone https://github.com/IntersectMBO/cardano-db-sync
cd cardano-db-sync
PGPASSFILE=config/pgpass-mainnet scripts/postgresql-setup.sh --createdb
git checkout <latest-official-tag> -b tag-<latest-official-tag>
```

it's line:
```
PGPASSFILE=config/pgpass-mainnet scripts/postgresql-setup.sh --createdb
```

Cmdv:
# Description

This fixes #1747

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/doc/building-running.md
+++ b/doc/building-running.md
@@ -55,15 +55,40 @@ make install
 
 ```
 
-- with cabal:
+### Clone repository
 
-```
+Start by cloning the `cardano-db-sync` repository from GitHub.
+```sh
 git clone https://github.com/IntersectMBO/cardano-db-sync
-cd cardano-db-sync
-PGPASSFILE=config/pgpass-mainnet scripts/postgresql-setup.sh --createdb
-git checkout <latest-official-tag> -b tag-<latest-official-tag>
+```
 
+Navigate into the cloned repository directory.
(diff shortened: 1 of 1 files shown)


sgillespie:
Good idea!

I'd also like you to add this to one of:

 * https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/running.md
 * https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/installing.md
 * https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/installing-with-nix.md

Further, I'm wondering if we need to keep this document around, since all the information seems to be covered in the documents above

Cmdv:
@sgillespie agreed I feel like we need to get this into a better shape all together because the new stuff is good but some of the historic is slightly outdated and spread across multiple files in slightly non logical ways.

kderme (approved):

PR #1756, opened 2024-07-02
1727 - Improve Tx queries performance

### LINKED ISSUE #1727: Improve the tx queries
The analysis at https://github.com/IntersectMBO/cardano-db-sync/issues/1721 shows that when we query the `tx` table, we unnecesarily get the whole tx instead of simply the id. We could improve this query.

In addition, we could use a small cache with the most recents txs. An analysis shows that 50% of outputs are consumes within the next 1000 bocks. So if we simply cache the latest 1000 blocks, which is maximum of 50000 txs, we can have a 50% hit-ratio.

Cmdv:
# Description

This fixes #1727 

> TxId: cache size: 350000, cache capacity: 350000, hit rate: 67%, hits: 897728, misses: 434637

why the size of the cache:
```
SELECT AVG(tx_count) as average_tx_count
FROM (
     SELECT tx_count
     FROM epoch
     ORDER BY id DESC
     LIMIT 200
) as last_200_epochs;

+---------------------+
| average_tx_count    |
|---------------------|
| 349219.595000000000 |
+---------------------+
````
SCR-20240709-eri

and it used to be around `696,813` queries.

TODO:
- [x] add to CacheStatus
- [x] Check the cache before making the query unless cache not active
- [x] make custom FIFO cache
- [x] Populate the cache right after inserting TX

# Checklist

- [x] Any changes are noted in the [changelog](https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)
- [x] Self-reviewed the diff

# Migrations

### CODE REVIEW

File: cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs [hunk 1]
```diff
@@ -45,6 +46,8 @@ import Ouroboros.Consensus.Cardano.Block (StandardCrypto)
 
 type StakePoolCache = Map PoolKeyHash DB.PoolHashId
 
+type CacheTx = Map ByteString DB.TxId
```
kderme, line 49: Keeping many small ByteString long term in memory is ill-advised, since they can cause memory fragmentation. That's why all ledger types are backed by ShortByteString and we should do the same. Also since ledger has all types in ShortByteString, we avoid the ShortByteString -> ByteString transition when we query something in the cache.

I just realised we use ByteString for `cStakeRawHashes` which should also change.

cardano-db-sync/src/Cardano/DbSync/Api/Ledger.hs
```diff
@@ -54,19 +54,19 @@ bootStrapMaybe ::
...
-    migrateBootstrapUTxO syncEnv emptyTxCache -- TODO: hardcoded to empty
-
-newtype TxCache = TxCache {txIdCache :: Map ByteString DB.TxId}
-
-emptyTxCache :: TxCache
-emptyTxCache = TxCache mempty
+    case envCache syncEnv of
+      NoCache -> migrateBootstrapUTxO syncEnv
+      ActiveCache ci -> do
+        -- TODO: hardcoded to empty
```
kderme, line 60: I think it's a good idea reusing the cache for this separate functionality

kderme:
Any ideas on how the eviction of old data will work?

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
Cmdv: Just to note the existing cache is:
```
newtype TxCache = TxCache {txIdCache :: Map ByteString DB.TxId}
```

So you're saying we need to use `ShortByteString` but try not to convert them back and forth often. But the code I changed uses `ByteString` as it's checking against `TxHash` in `queryTxId`. 

This means I'd have to convert `ByteString -> ShortByteString` every time I add a new value to the `Map`.

I'm unsure how I can use `ShortByteString` in the cache without doing the conversion?

cardano-db-sync/src/Cardano/DbSync/Api/Ledger.hs
Cmdv: I'm not sure I understand what you mean, what is "reusing the cache"? The exiting cache or the new cache which is part of `CacheStatus`. It's all honesty it's not directly clear what `TxCache` was used for and why it was just populated as empty because I couldn't see where it was been populated after.

kderme: I liked the idea of using the `TxCache` for the UTxO bootsrap, which is separate from https://github.com/IntersectMBO/cardano-db-sync/issues/1727. We get two for the cost of one.

Btw the previous definition of `TxCache` was an unimplemented optimization, so you can ignore.

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
kderme: The initial type for the hash of a tx is `TxId StandardCrypto` defined at
https://github.com/IntersectMBO/cardano-ledger/blob/f4119c9cd27e69943dcf91d22d0f9905be32fa75/libs/cardano-ledger-core/src/Cardano/Ledger/TxIn.hs#L56

Note that we use functions like `safeHashToByteString` and `unTxHash` to turn it into a ByteString before using  `queryTxId`. This transition may happen even before the creation of the Generic.Tx type currently.

I propose changing the Cache type to
```
newtype TxCache = TxCache {txIdCache :: Map (Ledger.TxId StandardCrypto) DB.TxId}
```

(I initially mentioned `ShortByteString` thinking that it's used to define `SafeHash`, but we don't need to go that deep. `TxId` is good enough)

kderme: Another note to avoid confusion. This definition
```
unTxHash :: TxId era -> ByteString
```

is a bit misleading. It should be
```
unTxHash :: TxId c -> ByteString
```

cardano-db-sync/src/Cardano/DbSync/Api/Ledger.hs
Cmdv: ahhhh that's why nothing was happening to it. Wasn't sure if it was voodoo magic or something 😅

Cmdv:
@kderme I'm making it an LRU so they'll just get pushed out if they haven't gotten used. Will be easier than trying to work out when a TxId in the cache should be deleted.

kderme:
We should somehow handle rollbacks, since `tx` table is deleted on rollbacks. I think we can follow a very simple approach, simply delete the whole TxCache when there is a rollback, as part of `rollbackCache`.

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
```diff
@@ -0,0 +1,144 @@
...
+
+6. Thread safety:
+   - Note: This implementation is not thread-safe. Concurrent access should be
+     managed externally if required.
+-}
+
+data FifoCache k v = FifoCache
+  { maxCapacity :: {-# UNPACK #-} !Natural
+  , currentSize :: {-# UNPACK #-} !Int
+  , queue :: !(Maybe (NE.NonEmpty k))
```
kderme, line 68: Is there any reason to use Maybe NonEmpty instead of List?

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -361,6 +369,72 @@ queryPrevBlockWithCache msg cache hsh =
...
+          liftIO $ missTxIds (cStats ci)
+          liftIO $ atomically $ modifyTVar (cTxIds ci) $ FIFO.insertTxIdCache ledgerTxId resTxId
+          pure resTxId
+
+resolveInputTxId ::
+  MonadIO m =>
+  Generic.TxIn ->
+  CacheStatus ->
+  ReaderT SqlBackend m (Either DB.LookupFail DB.TxId)
+resolveInputTxId txIn cache = do
```
kderme, line 399: There is a lot of duplication between this and `queryTxIdWithCache`

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
```diff
@@ -0,0 +1,144 @@
...
+
+insert :: Hashable k => k -> v -> FifoCache k v -> FifoCache k v
+insert key value cache@FifoCache {..}
+  | currentSize >= fromIntegral maxCapacity =
+      case queue of
+        Nothing -> cache -- This should never happen if invariants are maintained
+        Just q ->
+          let (oldestKey, mNewQueue) = NE.uncons q
+              newQueue = Just $ maybe (key NE.:| []) (NE.cons key) mNewQueue
+              newMap = HM.insert key value (HM.delete oldestKey cacheMap)
```
kderme, line 89: You delete from the `queue` but not from the `cacheMap`. This means that `cacheMap` will grow forever?

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
```diff
@@ -0,0 +1,144 @@
...
+6. Thread safety:
+   - Note: This implementation is not thread-safe. Concurrent access should be
+     managed externally if required.
+-}
+
+data FifoCache k v = FifoCache
+  { maxCapacity :: {-# UNPACK #-} !Natural
+  , currentSize :: {-# UNPACK #-} !Int
+  , queue :: !(Maybe (NE.NonEmpty k))
+  , cacheMap :: !(HM.HashMap k v)
```
kderme, line 69: Any specific reason to use HashMap instead of simple Map? Won't this cause hash collisions?

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
kderme: What is the reason for queue?

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
Cmdv: it's because for lookups, `HashMap` is O(1) average case, O(n) worst case yet `Map` is O(log n) which is useful for large dataset. 

`HashMap` will use slightly more memory usage but the performance gains arise when dealing upwards of tens of thousands of elements.

There shouldn't be any hash collisions as from my understanding all `txLedgerTxId` are unique by default.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
Cmdv: i tried to fix this duplication but the problem is the call site happen at different locations when `TxIn` is and isn't accessible. I managed to abstract out the cache logic but the insertion logic required a bunch of  higher-order functions and it made it way more confusing to read imo

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
kderme: `txLedgerTxId` are unique, but the hash of the tx is rehashed by the HashMap implementation. This second hash is not collision resistant and is one of the reasons that HashMap's are never used in the codebase of Cardano.

Existance of collisions is documented at https://hackage.haskell.org/package/hashable-1.5.0.0/docs/Data-Hashable.html#g:1

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
kderme: Doesn't `uncons` return the last element inserted?

Cmdv: no it's the first element in the list which is the oldest in the FIFO queue

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
Cmdv: ahh wondered why there wasn't a Hashable instance of `TxId`.

Well that throws a spanner in the works. Guess I'll just remove `queue :: !(Maybe (NE.NonEmpty k))` and make the cache using `Data.Map.Strict`. I was trying to be as fast as possible 😓

kderme: This shouldn't affect queue. Can't Hashable simply change to Data.Map.Strict?

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
kderme: `NonEmpty` is simply a list which can't be empty. It's not FIFO.

Cmdv: I've removed it

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
Cmdv: yeah that's what I've done now saves all the custom instances and the `Hashable` issue:

```
data FIFOCache k v = FIFOCache
  { maxCapacity :: !Word64
  , cacheMap :: !(Map.Map k v)
  , keyOrder :: !(Seq.Seq k)  -- To track insertion order
  }

empty :: Word64 -> FIFOCache k v
empty capacity = FIFOCache capacity Map.empty Seq.empty

insert :: Ord k => k -> v -> FIFOCache k v -> FIFOCache k v
insert key value cache@FIFOCache{..}
  | Map.size cacheMap >= fromIntegral maxCapacity =
      let oldestKey = Seq.index keyOrder 0
          !newKeyOrder = Seq.drop 1 keyOrder Seq.|> key
          !newMap = Map.insert key value $ Map.delete oldestKey cacheMap
      in cache { cacheMap = newMap, keyOrder = newKeyOrder }
  | otherwise =
      cache { cacheMap = Map.insert key value cacheMap
            , keyOrder = keyOrder Seq.|> key }

lookup :: Ord k => k -> FIFOCache k v -> Maybe v
lookup key FIFOCache{..} = Map.lookup key cacheMap

getSize :: FIFOCache k v -> Int
getSize = Map.size . cacheMap

getCapacity :: FIFOCache k v -> Word64
getCapacity = maxCapacity

cleanupCache :: FIFOCache k v -> FIFOCache k v
cleanupCache cache = cache { cacheMap = Map.empty, keyOrder = Seq.empty }
```

kderme: Seems, good, except I would suggest using the Strict version of Sequence in `cardano-strict-containers` `Data.Sequence.Strict`

Cmdv: cool just added `Data.Sequence.Strict` think I have it right!
```haskell
data FIFOCache k v = FIFOCache
  { maxCapacity :: !Word64
  , cacheMap :: !(Map.Map k v)
  , keyOrder :: !(StrictSeq k)
  }

empty :: Word64 -> FIFOCache k v
empty capacity = FIFOCache capacity Map.empty Seq.empty

insert :: Ord k => k -> v -> FIFOCache k v -> FIFOCache k v
insert key value cache@FIFOCache {..}
  | Map.size cacheMap >= fromIntegral maxCapacity =
      case Seq.lookup 0 keyOrder of
        Nothing -> cache -- This should never happen if invariants are maintained
        Just oldestKey ->
          let !newKeyOrder = key Seq.<| Seq.drop 1 keyOrder
              !newMap = Map.insert key value $ Map.delete oldestKey cacheMap
           in cache {cacheMap = newMap, keyOrder = newKeyOrder}
  | otherwise =
      cache
        { cacheMap = Map.insert key value cacheMap
        , keyOrder = keyOrder Seq.|> key
        }

lookup :: Ord k => k -> FIFOCache k v -> Maybe v
lookup key FIFOCache {..} = Map.lookup key cacheMap

getSize :: FIFOCache k v -> Int
getSize = Map.size . cacheMap

getCapacity :: FIFOCache k v -> Word64
getCapacity = maxCapacity

cleanupCache :: FIFOCache k v -> FIFOCache k v
cleanupCache cache = cache {cacheMap = Map.empty, keyOrder = Seq.empty}
```

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
```diff
@@ -162,25 +178,48 @@ textShowStats (ActiveCache ic) = do
       , textShow (prevBlockHits stats)
       , ", misses: "
       , textShow (prevBlockQueries stats - prevBlockHits stats)
+      , "\n  TxId: "
+      , "cache size: "
+      , textShow (FIFO.getSize txIds)
+      , ", cache capacity: "
+      , textShow (FIFO.getCapacity txIds)
+      , if credsQueries stats == 0
```
kderme, line 186:
```suggestion
      , if txIdsQueries stats == 0
```

Cmdv:
New stats with alternate `Strict.Map` + `Sequence.Strict`
```
epoch 429
TxId: cache size: 350000, cache capacity: 350000, hit rate: 65%, hits: 794125, misses: 424352
```
PGBadger results:
SCR-20240709-nn8

It feels like this could be better, I'm sure I'm adding any Tx that were called via DB with `tryUpdateCacheTx` in all locations which are:

**added**
cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/Tx.hs:125: 
cardano-db-sync/src/Cardano/DbSync/Era/Byron/Genesis.hs:248:

**not added**
cardano-db-sync/src/Cardano/DbSync/Era/Byron/Insert.hs:261: -- no access to Tx
cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Genesis.hs:280: 
cardano-db-sync/src/Cardano/DbSync/Era/Byron/Insert.hs:283:

cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
```diff
@@ -0,0 +1,64 @@
...
+      case Seq.lookup 0 keyOrder of
+        Nothing -> cache -- This should never happen if invariants are maintained
+        Just oldestKey ->
+          let !newKeyOrder = key Seq.<| Seq.drop 1 keyOrder
+              !newMap = Map.insert key value $ Map.delete oldestKey cacheMap
+           in cache {cacheMap = newMap, keyOrder = newKeyOrder}
+  | otherwise =
+      cache
+        { cacheMap = Map.insert key value cacheMap
+        , keyOrder = keyOrder Seq.|> key
```
kderme, line 51:
```suggestion
        , keyOrder = keyOrder Seq.<| key
```

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
```diff
@@ -162,25 +180,55 @@ textShowStats (ActiveCache ic) = do
       , textShow (prevBlockHits stats)
       , ", misses: "
       , textShow (prevBlockQueries stats - prevBlockHits stats)
+      , "\n  TxId: "
+      , "cache size: "
+      , textShow (FIFO.getSize txIds)
+      , ", cache capacity: "
+      , textShow (FIFO.getCapacity txIds)
```
kderme, line 187: May no be needed

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/LedgerEvent.hs
```diff
@@ -75,6 +75,7 @@ insertNewEpochLedgerEvents syncEnv currentEpochNo@(EpochNo curEpoch) =
           liftIO . logInfo tracer $ "Persistant SQL Statement Cache size is " <> textShow persistantCacheSize
           stats <- liftIO $ textShowStats cache
           liftIO . logInfo tracer $ stats
+          liftIO $ resetCacheStatistics cache
```
kderme, line 78: may not be needed

kderme:
Merged through https://github.com/IntersectMBO/cardano-db-sync/pull/1769, thank you!

PR #1762, opened 2024-07-05
Improve stake cache

kderme:
# Description

Use ledger types instead of ByteString and use double cache.
- One cache works as before: it's only populated for addresses with rewards/delegations and entries are only evicted manually on a deregistration. Exptected current size on mainnet os 1_700_000
- A second smaller LRU cache with capacity 100_000.
 
# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Cache.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Cache.hs
@@ -30,7 +30,7 @@ import Cardano.BM.Trace
 import qualified Cardano.Db as DB
 import Cardano.DbSync.Cache.Epoch (rollbackMapEpochInCache)
 import qualified Cardano.DbSync.Cache.LRU as LRU
-import Cardano.DbSync.Cache.Types (CacheAction (..), CacheInternal (..), CacheStatistics (..), CacheStatus (..), StakeCache (..), initCacheStatistics)
+import Cardano.DbSync.Cache.Types (CacheAction (..), CacheInternal (..), CacheStatistics (..), CacheStatus (..), StakeCache (..), initCacheStatistics, shouldCache)
 import qualified Cardano.DbSync.Era.Shelley.Generic.Util as Generic
 import Cardano.DbSync.Era.Shelley.Query
 import Cardano.DbSync.Era.Util
@@ -89,7 +89,7 @@ queryOrInsertRewardAccount ::
   Ledger.RewardAccount StandardCrypto ->
   ReaderT SqlBackend m DB.StakeAddressId
 queryOrInsertRewardAccount trce cache cacheUA rewardAddr = do
-  eiAddrId <- queryRewardAccountWithCacheRetBs trce cache cacheUA rewardAddr
+  eiAddrId <- queryStakeAddrWithCacheRetBs trce cache cacheUA rewardAddr
   case eiAddrId of
     Left (_err, bs) -> insertStakeAddress rewardAddr (Just bs)
     Right addrId -> pure addrId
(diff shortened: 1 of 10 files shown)


Cmdv (changes requested):
Do we have any performance stats from running a couple of epochs or seeing if there is a query improvement via PGBadger?

I'd be interested to know how big `scStableCache` gets before `deleteStakeCache` is called?

edit: I left a load of comments but it seems that nothing came through will add them in the morning 🤦

kderme:
> I'd be interested to know how big scStableCache gets before deleteStakeCache is called?

Entries are deleted only when the user of the api asks for an Eviction, ie when the stake address is deregistered.

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
```diff
@@ -53,12 +60,13 @@ data CacheStatus
 
 data CacheAction
   = UpdateCache
+  | UpdateCacheStrong
```
Cmdv, line 63: I was a little left confused with the naming of `UpdataCacheStong` from my understanding:

`UpdataCacheStong` inserts into `scStableCache`
`scStableCache` inserts into `scLruCache`

not sure if renaming would help, here's a couple:

```haskell
data CacheUpdateStrategy
  = UpdateLRU
  | UpdateStable
  | NoUpdate
  | EvictAndUpdate
  deriving (Eq)
```
```haskell
data CacheUpdateStrategy
  = UpdateFlexible
  | UpdateFixed
  | NoUpdate
  | EvictAndUpdate
  deriving (Eq)
```

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -143,59 +132,60 @@ queryStakeAddrWithCache ::
...
+                    UpdateCache -> stakeCache {scLruCache = LRU.insert cred stakeAddrsId (scLruCache stakeCache)}
+                    UpdateCacheStrong -> stakeCache {scStableCache = Map.insert cred stakeAddrsId (scStableCache stakeCache)}
+                    _ -> stakeCache
+              liftIO $
+                atomically $
+                  writeTVar (cStakeRawHashes ci) stakeCache'
               pure $ Right stakeAddrsId
 
+-- | True if it was found in LRU
+queryStakeCache :: StakeCred -> StakeCache -> Maybe (DB.StakeAddressId, StakeCache, Bool)
```
Cmdv, line 178: It doesn't look like the `Bool` is used at the call site.
Also how come we check `scStableCache` first isn't that more likely to be the bigger cache of the two?

cardano-db-sync/src/Cardano/DbSync/Cache/Types.hs
```diff
@@ -45,6 +47,11 @@ import Ouroboros.Consensus.Cardano.Block (StandardCrypto)
 
 type StakePoolCache = Map PoolKeyHash DB.PoolHashId
 
+data StakeCache = StakeCache
```
Cmdv, line 50: Is it possible to have a little code comment about the reasoning for having 2 caches?

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Epoch.hs
```diff
@@ -214,7 +214,7 @@ insertEpochStake syncEnv nw epochNo stakeChunk = do
       (StakeCred, (Shelley.Coin, PoolKeyHash)) ->
       ExceptT SyncNodeError (ReaderT SqlBackend m) DB.EpochStake
     mkStake cache (saddr, (coin, pool)) = do
-      saId <- lift $ queryOrInsertStakeAddress trce cache UpdateCache nw saddr
+      saId <- lift $ queryOrInsertStakeAddress trce cache UpdateCacheStrong nw saddr
```
Cmdv, line 217: how can one tell which update strategy they should pick? I can see we have two, but can't work out why some go the LRU whilst others go to the Map.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
kderme, line 178:
```suggestion
queryStakeCache :: StakeCred -> StakeCache -> Maybe (DB.StakeAddressId, StakeCache)
```

kderme:
Merged through https://github.com/IntersectMBO/cardano-db-sync/pull/1769

Issue #1751, opened 2024-07-01
Missing Documentation per OSC Governance


Ranchhand87:
The following Required docs are missing and should show on the Main page of Code: CHANGELOG.md
If there are CHANGELOG's please create a top level file and link accordingly.

The following documents/features are suggested by OSC Governance: SECURITY.md

For Reference: https://intersect.gitbook.io/open-source-committee/policies/governance
Also see Documentation and Legal

sgillespie:
We have a [CHANGELOG](https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md). We could either move it or symlink it. 

@kderme is there any reason we want to keep the `cardano-db-sync` CHANGELOG separate from everything else in this repo? If not, I propose we move it to the top level

kderme:
We only maintain a single Changelog and don't split it into packages. I'm fine with moving it to top level.

Cmdv:
closing issue as it has been solved 👍

PR #1763, opened 2024-07-05
doc: Move CHANGELOG.md to top-level

sgillespie:
# Description

Partially addresses #1751 

# Checklist

- [x] Self-reviewed the diff

# Migrations


sgillespie:
Note that we have a bunch of other local package-specific changelogs. I'm not sure if we should do anything with them, so I left them as is

Cmdv (approved):

kderme (dismissed):
LGTM! May need a rebase though

kderme:
Seems like some newer entries have been lost during rebasing

```
* Added an optional feature to store pool stats per epoch transactions in `pool_stat` table
* EnableFutureGenesis is now true by default
```

sgillespie:
> Seems like some newer entries have been lost during rebasing
> 
> ```
> * Added an optional feature to store pool stats per epoch transactions in `pool_stat` table
> * EnableFutureGenesis is now true by default
> ```

I knew it looked to easy!

Cmdv (approved):

PR #1769, opened 2024-07-11
Prepare release 13.3

kderme:
# Description

Cherry-picks https://github.com/IntersectMBO/cardano-db-sync/pull/1756, but changes it to a rolling cache
Cherry-picks https://github.com/IntersectMBO/cardano-db-sync/pull/1762
Introduces EpochState 

TODO:
Add EpochState migration file
Document Schema for EpochState
Document Schema for Conway
Update Changelog
Make EnableFutureGenesis True by default

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 


cardano-db-sync/src/Cardano/DbSync/Cache/FIFO.hs
```diff
@@ -0,0 +1,68 @@
...
+for key-value storage. It keeps 3 rolling Maps and only the first one is inserted.
+When the first one is full, the last Map is removed.
+
+Key operations and their complexities:
+- Insertion: O(log mapCapacity)
+- Lookup: O(log (mapCapacity * n))
+- Removal of oldest elements: O(1)
+- Size query: O(1)
+-}
+data FIFOCache k v = FIFOCache
```
Cmdv, line 29: this isn't technically a FIFO cache anymore, did you find this to be more performant as I don't recall discussing this?

kderme: This is what was initially discussed (I used 3 Maps instead of 2 to smoothen the flactuations of total Cache size)
It is some form of FIFO except it doesn't evicts a single element, but 1/3 of them when needed.

I plan to run a full sync on this pr today to check performance.

kderme: > lookups would 0(log n) * 3

Worst case yes, but we know that the first Map has a way higher probability to hit than the other 2. So for most queries it will simply lookup a smaller Map. Also the previous implementation with the Queue, requires 2 Map operations per insert, one lookup and one delete and on a bigger Map.

Cmdv: Cool, guess the stats will tell all 😅 I'm sure there's actually not much in it, I didn't keep the stats but when I tried a 2 Map approach it wasn't as good as the HashMap and the StrictMap was a little less performant than the HashMap. 🤷

kderme:
@Cmdv for reference, syncing with this one 
```
Cache Statistics:
  Stake Addresses: cache sizes: 980132 and 100000, hit rate: 94%, hits: 118881456, misses: 6791890
  Pools: cache size: 4754, hit rate: 99%, hits: 91357561, misses: 471366
  Datums: cache capacity: 250000, cache size: 93121, hit rate: 35%, hits: 51026, misses: 93121
  Multi Assets: cache capacity: 250000, cache size: 250000, hit rate: 81%, hits: 92183671, misses: 21156109
  Previous Block: hit rate: 49%, hits: 6678620, misses: 6678622
  TxId: cache size: 204752, cache capacity: 300000, hit rate: 75%, hits: 36143578, misses: 11613943
[db-sync-node:Info:77] [2024-07-12 08:24:08.36 UTC] Starting epoch 311
```
Will post more

PR #1775, opened 2024-07-16
1641 - config can override presets

### LINKED ISSUE #1641: Config: Preset ALWAYS overrides Ledger
Currently, if using a preset, you cannot also configure the ledger. This is because the preset overrides ALL other options. This means that, in following config,

```
"insert_options": {
  "preset": "only_utxo",
  "ledger": "enable"
}
```

The ledger setting will have no effect. We would like to allow users to specify the ledger setting along with the prefix, and set a default if none is specified (`"enable"`).

One way to do this would be to add a Ledger field to all of the preset constructors: https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/src/Cardano/DbSync/Config/Types.hs#L149

Cmdv:
# Description

This fixes #1641

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Config/Types.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Config/Types.hs
@@ -55,6 +55,10 @@ module Cardano.DbSync.Config.Types (
   isTxOutConsumed,
   isTxOutPrune,
   forceTxIn,
+  fullInsertOptions,
+  onlyUTxOInsertOptions,
+  onlyGovInsertOptions,
+  disableAllInsertOptions,
 ) where
 
 import qualified Cardano.BM.Configuration as Logging
@@ -68,7 +72,8 @@ import Cardano.Slotting.Slot (SlotNo (..))
 import Control.Monad (fail)
 import Data.Aeson (FromJSON (..), ToJSON (..), Value (..), (.!=), (.:), (.:?), (.=))
 import qualified Data.Aeson as Aeson
-import Data.Aeson.Types (Parser, typeMismatch)
+import Data.Aeson.Key (fromText)
+import Data.Aeson.Types (Pair, Parser, typeMismatch)
 import Data.ByteString.Short (ShortByteString (), fromShort, toShort)
 import Data.Default.Class (Default (..))
 import Ouroboros.Consensus.Cardano.CanHardFork (TriggerHardFork (..))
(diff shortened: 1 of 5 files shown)


cardano-db-sync/src/Cardano/DbSync/Config/Types.hs
```diff
@@ -150,14 +155,20 @@ data SyncPreConfig = SyncPreConfig
...
   deriving (Eq, Show)
 
+-- data SyncInsertConfig
+--   = FullInsertOptions
+--   | OnlyUTxOInsertOptions
+--   | OnlyGovInsertOptions
+--   | DisableAllInsertOptions
+--   | SyncInsertConfig SyncInsertOptions
+--   deriving (Eq, Show)
+
```
sgillespie, line 171: What's this?

sgillespie: I think instead of completely removing this, we can slightly change it:
```
data SyncInsertPreset
   = FullInsertPreset
   | OnlyUTxOInsertPreset
   | OnlyGovInsertPreset
   | DisableAllInsertPreset
  deriving (Eq, Show)
```

and then sicPreset can be `Maybe SyncInsertPreset`. This would avoid the stringly typed tests below

cardano-db-sync/test/Cardano/DbSync/ApiTest.hs
```diff
@@ -22,46 +23,50 @@ prop_extractInsertOptions :: Property
...
 
-  let insertOpts = pcInsertConfig cfg
-  coverInsertCfg insertOpts
+  let insertCfg = pcInsertConfig cfg
+  coverInsertCfg insertCfg
 
-  case insertOpts of
-    FullInsertOptions ->
+  case insertCfg of
+    SyncInsertConfig (Just "full") _ ->
```
sgillespie, line 30: These would be nicer to use as ADTs (rather than strings), as I mentioned in my other comment

cardano-db-sync/test/Cardano/DbSync/Gen.hs
```diff
@@ -106,11 +107,11 @@ syncNodeConfig loggingCfg =
...
-    [ pure FullInsertOptions
-    , pure OnlyUTxOInsertOptions
-    , pure OnlyGovInsertOptions
-    , pure DisableAllInsertOptions
-    , SyncInsertConfig <$> syncInsertOptions
+    [ pure $ SyncInsertConfig (Just "full") fullInsertOptions
+    , pure $ SyncInsertConfig (Just "only_utxo") onlyUTxOInsertOptions
+    , pure $ SyncInsertConfig (Just "only_gov") onlyGovInsertOptions
+    , pure $ SyncInsertConfig (Just "disable_all") disableAllInsertOptions
+    , SyncInsertConfig Nothing <$> syncInsertOptions
```
sgillespie, line 114: We might want to add another choice in here to generate a preset AND overrides

sgillespie (changes requested):
Good job! I had a few concerns on the new type and the property tests. Feel free to discuss if you don't agree

cardano-db-sync/src/Cardano/DbSync/Config/Types.hs
Cmdv: sorted, not a clue what I was thinking there!! 😅

cardano-db-sync/test/Cardano/DbSync/Gen.hs
sgillespie: I think we're still missing some cases, because of `parseOverrides`, but I can't think of a good way to deal with this.

sgillespie (approved):

PR #1778, opened 2024-07-19
test: Add a Conway parameter change governance test

sgillespie:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-chain-gen/test/testfiles/config-conway-bootstrap/test-db-sync-config-no-metadata.json
```diff
@@ -0,0 +1,119 @@
...
+  ],
+  "setupScribes": [
+    {
+      "scFormat": "ScText",
+      "scKind": "StdoutSK",
+      "scName": "stdout",
+      "scRotation": null
+    }
+  ],
+  "insert_options": {
```
Cmdv, line 114: Could all of these `test-db-sync-config-*` could be created in code as the only difference I can see is the `insert_options:`?

sgillespie: This is a good observation. I copied the whole directory which has a bunch of stuff that's not needed. I only needed to update the shelley genesis config, I can pretty much delete everything else

PR #1786, opened 2024-07-24
Fixes for 13.3.0.0

kderme:
# Description

Fixes
- an issue where script data were considered wrong even if they are not
- falling back to CIP-100 if CIP-108 or CIP-119 is not followed
- cardano-db-tool upgrade

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/OffChain/Vote/Types.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/OffChain/Vote/Types.hs
@@ -58,9 +58,19 @@ getRationale = \case
 
 eitherDecodeOffChainVoteData :: LBS.ByteString -> DB.AnchorType -> Either String OffChainVoteData
 eitherDecodeOffChainVoteData lbs = \case
-  DB.GovActionAnchor -> OffChainVoteDataGa <$> eitherDecode' lbs
-  DB.DrepAnchor -> OffChainVoteDataDr <$> eitherDecode' lbs
+  DB.GovActionAnchor ->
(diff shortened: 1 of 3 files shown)


Cmdv:
Any description as to what these fixes are other than the commit messages?

cardano-db-sync/src/Cardano/DbSync/OffChain/Vote/Types.hs
```diff
@@ -58,10 +58,20 @@ getRationale = \case
...
+    eitherDecodeAlternative (OffChainVoteDataDr <$> eitherDecode' lbs) (OffChainVoteDataOther <$> eitherDecode' lbs)
   DB.OtherAnchor -> OffChainVoteDataOther <$> eitherDecode' lbs
 
+eitherDecodeAlternative :: Either String OffChainVoteData -> Either String OffChainVoteData -> Either String OffChainVoteData
+eitherDecodeAlternative one two =
+  case one of
+    Right _ -> one
+    Left err1 -> case two of
+      Right _ -> two
+      Left err2 -> Left $ err1 <> ", CIP-100:" <> err2
```
sgillespie, line 73: I wonder if this could be done with `<|>`?

kderme: `<|>` doesn't provide a way to return a default value in case of a failure. I think that's why Either e doesn't have an Alternative instance, while ie for `Maybe` there is an easy default value `Nothing`

kderme:
> Any description as to what these fixes are other than the commit messages?

Added description

Cmdv (approved):

PR #1797, opened 2024-07-30
Doc: Rework Docker documentation

sgillespie:
# Description

Updates the docker documentation to (hopefully!) be more organized and up-to-date. Here is the [rendered version](https://github.com/sgillespie/cardano-db-sync/blob/doc/docker/doc/docker.md).

I've also created a new `docker-compose.example.yml`. The ideas around this are:

 1. Discourage usage in non-dev workflows
 2. Encourage users to maintain their own version
 3. Avoid tying this file to db-sync releases

To that end, I've left the existing one for now with the idea that we'll remove it in a few months.

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -2,277 +2,294 @@
 
 ## Quickstart
 
-### Clone the repository
+We provide an example
+[docker-compose.yml](https://github.com/IntersectMBO/cardano-db-sync/blob/master/docker-compose.example.yml)
+to quickly get set up with `cardano-db-sync`. Keep in mind that this is only a template,
+and users are encouraged to tweak it to meet their needs.
 
-Regular users should almost never attempt building and running from the `master` branch. Instead,
-they should build and run the latest release tag. The tags can be listed using the `git tag`
-command. Pre-release tags (eg things like `12.0.0-preX`) should also be avoided in most cases.
-```
-git clone git@github.com:IntersectMBO/cardano-db-sync.git
-cd cardano-db-sync
-git checkout <latest-official-tag> -b tag-<latest-official-tag>
-```
-### Start `cardano-node`, `postgresql`, and `cardano-db-sync` services using Docker
+It is not recommended to use this for a production set up, and should only be
+used for local development.
 
-``` console
-docker-compose up -d && docker-compose logs -f
-```
-### tada:tada:
+Create a working directory:
 
-The PostgreSQL database is exposed on localhost port `5432`
-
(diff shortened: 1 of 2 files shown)


doc/docker.md
```diff
@@ -2,49 +2,57 @@
 
 ## Quickstart
 
-### Clone the repository
+We provide an example
+[docker-compose.yml](https://github.com/IntersectMBO/cardano-db-sync/blob/master/docker-compose.yml)
+to quickly get set up with `cardano-db-sync`.  Keep in mind that this is only a template,
```
Cmdv, line 7: double space in front of `  Keep`

doc/docker.md
```diff
@@ -55,224 +63,211 @@ b73d677399de:/# psql -U postgres -c "SHOW ALL" | grep maintenance
...
+Specifies a `cardano-db-sync` snapshot archive to download and restore from. If omitted,
+it will sync from genesis.
+
+### `DB_SYNC_CONFIG` (optional)
+
+Overrides the `db-sync-config.json` provided by the network configuration.
+
+#### `EXTRA_DB_SYNC_ARGS` (optional)
+
+Extra command line arguments to pass to `cardano-db-sync`.
```
Cmdv, line 110: I'm not sure if you agree but would be cool to have little examples of each arg? Also maybe a little link to the command line args

doc/docker.md
sgillespie: Sharp eye!

doc/docker.md
sgillespie: > maybe a link to the command line args docs

On second thought, we apparently don't have CLI args docs anymore :shrug:

sgillespie: Fixed up a little bit, any better?

doc/docker.md
Cmdv: 👓

doc/docker.md
Cmdv: perfect ❤️ 

> On second thought, we apparently don't have CLI args docs anymore 🤷

did we/I accidentally remove them?

Cmdv (approved):
🚢 :shipit:

doc/docker.md
sgillespie: No, we turned it into the config docs

Cmdv: ah yeah of course, I don't mind adding them in another PR so we get this PR through

Issue #1697, opened 2024-05-09
Support different schemas for Address


kderme:
Even though db-sync already supports different insert options, most of them operate on the same db schema. The only case where this is not true is the `consumed-tx-out` flag, which adds a new field to the `tx_out` table. Since we want to optionally add a new table `Address`, which also requires to change the `tx_out.address` field, we have the following challenges:
- An option to add the address feauture should be orthogonal to `consumed-tx-out`
- This forces us to support 4 different combinations, so 4 different schemas
- Persistent doesn't provide much flexibility for this. We may want 4 different Schema files which define at least `TxOut`

I think for this we can follow the same approach as we did for `consumed-tx-out`: Initially run the `.sql` file migrations and then run additional migration from Haskell code instead of `.sql` files. We could keep a flag in `ExtraMigrations` about whether the address table feauture is enabled and adjust the queries accordingly.

PR #1820, opened 2024-08-19
1333 - add new AddressDetail table for TxOut

### LINKED ISSUE #1333: Split out addresses table from tx_out

Cmdv:
# Description

Adding a new `AddressDetail` table for `TxOut` this functionality lives behind a config flag `use_address_table`.

This fixes #1333 + #1697

Also restructured the `Db` directory as it wasn't in the best state
```
.
├── Db
│   ├── Error.hs
│   ├── Git
│   │   ├── RevFromGit.hs
│   │   └── Version.hs
│   ├── Migration
│   │   ├── Haskell.hs
│   │   └── Version.hs
│   ├── Migration.hs
│   ├── Operations
│   │   ├── Core
│   │   │   ├── AlterTable.hs
│   │   │   ├── Delete.hs
│   │   │   ├── Insert.hs
│   │   │   ├── MinId.hs
│   │   │   ├── Query.hs
│   │   │   └── QueryHelper.hs
│   │   └── Variant
│   │       ├── ConsumedTxOut.hs
│   │       ├── JsonbInSchemaQueries.hs
│   │       ├── Multiplex.hs
│   │       └── TxOut.hs
│   ├── PGConfig.hs
│   ├── Run.hs
│   ├── Schema
│   │   ├── Core
│   │   │   ├── Schema.hs
│   │   │   └── TxOut.hs
│   │   ├── Orphans.hs
│   │   ├── Types.hs
│   │   └── Variant
│   │       ├── ConsumedTxOut.hs
│   │       └── TxOut.hs
│   ├── Types.hs
│   └── Version
│       ├── V13_0
│       │   ├── Query.hs
│       │   └── Schema.hs
│       └── V13_0.hs
└── Db.hs
```
# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db/Query.hs
```diff
@@ -1233,34 +1242,36 @@ queryAdaPots blkId = do
 -- | Get the UTxO set after the specified 'BlockId' has been applied to the chain.
 -- Not exported because 'BlockId' to 'BlockHash' relationship may not be the same
 -- across machines.
-queryUtxoAtBlockId :: MonadIO m => BlockId -> ReaderT SqlBackend m [(TxOut, ByteString)]
+queryUtxoAtBlockId :: MonadIO m => BlockId -> ReaderT SqlBackend m [(TxOut, Text, ByteString)]
```
Cmdv, line 1245: I'll need two versions of these queries, 
1: for use with`AdressDetail` and when config flag is true
2: normal use with no flag or use of `AddressDetail` table

cardano-db/src/Cardano/Db/Schema.hs
```diff
@@ -151,14 +151,15 @@ share
...
-    address             Text
+    address             Text Maybe
     addressHasScript    Bool
     paymentCred         ByteString Maybe    sqltype=hash28type
     stakeAddressId      StakeAddressId Maybe noreference
     value               DbLovelace          sqltype=lovelace
     dataHash            ByteString Maybe    sqltype=hash32type
     inlineDatumId       DatumId Maybe       noreference
     referenceScriptId   ScriptId Maybe      noreference
+    addressDetailId     AddressDetailId Maybe noreference
```
kderme, line 162: I think this needs a new Schema.hs file, instead of editting the default tx_out table. It's fine adding a new table, but eventually we don't want to change the existing tables.

Cmdv: It's really only changing `address Text Maybe` by adding a `Maybe` This shouldn't change functionality for how current users query the db? 

Though technically we could store the address in both tables as it's not big in size so this would remain just a `Text`.

This solution means that we don't have to keep multiple schemas updated for what isn't really an excessive feature that can be kept behind a config flag.

Just the more we have schemas the more complex the handling of their functionality within the code will become, especially when more than one schema is involved.

I know currently we run raw sql in the code to handle the extra schema and again the more of these we have the harder it will be to see what is happening from within the codebase.

Ideally we could just make these tables dynamic at runtime but from researching https://stackoverflow.com/questions/78871107/dynamic-tables-at-runtime-in-haskell-persistent there doesn't seem to be a way we can do it with Persistent.

This implementation would be an additional table in the base schema + 1 new `Maybe` field for the pointer to table if needed.

If you think that having another schema is still the best way to go then happy to do it, just wanted to note my apprehensions in doing so 👍

kderme: It's interesting trying to merge the multiple schemas into 1. We could make a pros-cons list of this approach:
**pros:**
- It makes it easier to introduce and **maintain** 1 schema instead of multiple. More and faster feautures means happier users.
- Users get **exposed** to different feautures directly which may motivate them migrate to a better schema solution for them.

**cons**
- We don't take the most benefit of each implementation **performance** wise. By merging the 2 schemas into 1, we force multiple Null or reduntant fields in a table that is hugely populated, so it takes space.
- Many unused null fields can **confuse** users. So can many fields can are annotated as nullable due to a different config. eg the address of a tx_out should never be null normally for a users that only cares about default behaviour.

I'm afraid I don't have a clear opinion right now. I guess an important question is, is the main objective of this feauture to reduce space (in which case cons outweight) or to improve tx_out queries?

Cmdv: @kderme Ok I've managed to do something else. We have 2 new schema files and a "base" schema
[BasicAddress](https://github.com/IntersectMBO/cardano-db-sync/blob/1aae0342269828f1bdd2744be94f62db2ecd9fcb/cardano-db/src/Cardano/Db/Schema/Address/BasicAddress.hs)
[ExtendedAddress](https://github.com/IntersectMBO/cardano-db-sync/blob/1aae0342269828f1bdd2744be94f62db2ecd9fcb/cardano-db/src/Cardano/Db/Schema/Address/ExtendedAddress.hs)

The config uses the type `data AddressType = AddressBasic | AddressExtended` and this flag is then used when we migrate the schemas (this part I need help with as I'm not 100% how our migration is working but technically it should work) something in the lines of:
```Haskell
runMigration migrateBaseCardanoDb
case addressType of
  AddressBasic -> runMigration migrateBasicAddressCardanoDb
  AddressExtended -> runMigration migrateExtendedAddressCardanoDb
```
From there we have to handle all the places that uses `DB.TxOut` to make sure we are using the right types for the right table being used. I converted a couple of queries in slightly different styles in [Cardano.Db.Query.TxOut](https://github.com/IntersectMBO/cardano-db-sync/blob/1aae0342269828f1bdd2744be94f62db2ecd9fcb/cardano-db/src/Cardano/Db/Query/TxOut.hs)

The draw backs being that I'm not sure if you could switch back and forth with changing the config value, handling cases whenever using `DB.TxOut` and dependent on style chosen in `Cardano.Db.Query.TxOut` we could have two functions for each db query related to effected tables. When generating migration files it's probably best to always take priority on using base/basic schemas (should probably rename `BasicAddress` -> `BaseAddress`). Then have the affected changes documented in our docs.

Come to think of something similar could be implemented with `Cardano.Db.Migration.Extra.CosnumedTxOut.Schema` but how we do the migration I'd need pointing as to what happen because it's not clear to me.

The benefit would be existing table schema would be exactly the same and we'd have a "Base" schema with the ability to have custom schemas to handle specific scenarios. We'd get the types for both variations so we'd keep the type safety.

kderme: > The draw backs being that I'm not sure if you could switch back and forth with changing the config value

cardano-db/src/Cardano/Db/Schema/Address/ExtendedAddress.hs
```diff
@@ -0,0 +1,94 @@
...
+----------------------------------------------
+-- Extended Address TxOut
+----------------------------------------------
+  TxOut
+    addressId           AddressId Maybe noreference
+    dataHash            ByteString Maybe    sqltype=hash32type
+    index               Word64              sqltype=txindex
+    inlineDatumId       DatumId Maybe       noreference
+    referenceScriptId   ScriptId Maybe      noreference
+    stakeAddressId      StakeAddressId Maybe noreference
```
kderme, line 46: I think this can be removed, since it now exists in `Address`

cardano-db/src/Cardano/Db/Query/TxOut.hs
```diff
@@ -0,0 +1,181 @@
...
+  (^.),
+  type (:&) ((:&)),
+ )
+import Database.Persist.Sql (PersistField)
+
+data AddressType = AddressBasic | AddressExtended
+
+class (PersistEntity (TxOutTable a), PersistField (TxOutIdType a)) => TxOutFields (a :: AddressType) where
+  type TxOutTable a :: Type
+  type TxOutIdType a :: Type
```
sgillespie, line 44: :fire:

kderme:
The `FOREIGN KEY ma_tx_out_tx_out_id_fkey` in function `createConsumedTxOut` is only necessary if prune is enabled.
Also it needs to run only once, unless the "IF NOT EXIST" clause can be used. The `wPruneTxOutPreviouslySet` in `runExtraMigrations` can be used to make sure this is run only once.

cardano-db/src/Cardano/Db/Operations/Variant/TxOutDelete.hs
```diff
@@ -0,0 +1,41 @@
...
+deleteCoreTxOutTablesAfterTxId mtxOutId mmaTxOutId = do
+  whenJust mmaTxOutId $ \maTxOutId -> deleteWhere [C.MaTxOutId >=. maTxOutId]
+  whenJust mtxOutId $ \txOutId -> deleteWhere [C.TxOutId >=. txOutId]
+
+-- TODO: cmdv: probably won't need to remove the addressId here but have it just incase
+deleteVariantTxOutTablesAfterTxId :: MonadIO m => Maybe V.TxOutId -> Maybe V.MaTxOutId -> Maybe V.AddressId -> ReaderT SqlBackend m ()
+deleteVariantTxOutTablesAfterTxId mtxOutId mmaTxOutId mAddrId = do
+  whenJust mmaTxOutId $ \maTxOutId -> deleteWhere [V.MaTxOutId >=. maTxOutId]
+  whenJust mtxOutId $ \txOutId -> deleteWhere [V.TxOutId >=. txOutId]
+  whenJust mAddrId $ \addrId -> deleteWhere [V.AddressId >=. addrId]
```
kderme, line 36: Addresses may be shared between multiple different outputs. So I don't believe it should be deleted here, just because one of its outputs is rollback.

cardano-db/src/Cardano/Db/Operations/Core/MinId.hs
```diff
@@ -0,0 +1,165 @@
...
+import qualified Cardano.Db.Schema.Variant.TxOut as V
+import Cardano.Prelude
+import qualified Data.Text as Text
+import Database.Persist.Sql (PersistEntity, PersistField, SqlBackend, fromSqlKey, toSqlKey)
+
+data MinIds (a :: TxOutTableType) = MinIds
+  { minTxInId :: Maybe TxInId
+  , minTxOutId :: Maybe (TxOutIdFor a)
+  , minMaTxOutId :: Maybe (MaTxOutIdFor a)
+  }
```
kderme, line 26: There is a lot of code duplication in this module, could it be reduced by using the `TxOutFields` class more? Alternatively MinIds could be defined using `TxOutIdW` and `TxOutIdW` or even by introducing a `MinIdsW` wraper.

Cmdv: I was skeptical as to what I should do as I don't really understand the original implementation I was just trying to keep original functionality and have `minTxOutId` + `minMaTxOutId` be variants

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Query.hs
```diff
@@ -12,6 +12,8 @@ module Cardano.DbSync.Era.Shelley.Query (
 ) where
 
 import Cardano.Db
+import qualified Cardano.DbSync.Api as Db
```
kderme, line 15: importing this qualfied as Db is a bit confusing

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/Grouped.hs
```diff
@@ -61,12 +63,13 @@ data MissingMaTxOut = MissingMaTxOut
 -- reference outputs that are not inserted to the db yet.
 data ExtendedTxOut = ExtendedTxOut
   { etoTxHash :: !ByteString
-  , etoTxOut :: !DB.TxOut
+  , etoTxOut :: !DB.TxOutW
+  , etoPaymentCred :: !(Maybe ByteString)
```
kderme, line 67: Why is this necessary? The `VTxOutW` already contains the address and the payment credential?

kderme: I see the address in `VTxOutW` can be Nothing. But why is this allowed?

Cmdv: yeah I'd done this way before doing `VTxOutW` at the time it was a way to get to `Address` payment cred. Will change 👍

cardano-db/src/Cardano/Db/Operations/TxOut/TxOutQuery.hs
```diff
@@ -0,0 +1,576 @@
...
+--------------------------------------------------------------------------------
+queryAddressId :: MonadIO m => ByteString -> ReaderT SqlBackend m (Maybe V.AddressId)
+queryAddressId addrRaw = do
+  res <- select $ do
+    addr <- from $ table @V.Address
+    where_ (addr ^. V.AddressRaw ==. val addrRaw)
+    pure (addr ^. V.AddressId)
+  pure $ unValue <$> listToMaybe res
+
+queryAddressById :: MonadIO m => V.AddressId -> ReaderT SqlBackend m (Maybe V.Address)
```
kderme, line 401: Possibly not needed

cardano-db/src/Cardano/Db/Operations/TxOut/TxOutQuery.hs
```diff
@@ -0,0 +1,576 @@
...
+  res <- select $ do
+    addr <- from $ table @V.Address
+    where_ (addr ^. V.AddressId ==. val addrId)
+    pure addr
+  pure $ entityVal <$> listToMaybe res
+
+--------------------------------------------------------------------------------
+-- queryAddressOutputs
+--------------------------------------------------------------------------------
+queryAddressOutputs :: MonadIO m => TxOutTableType -> Text -> ReaderT SqlBackend m DbLovelace
```
kderme, line 412: Move this and similar to tests section

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Genesis.hs
```diff
@@ -150,27 +152,30 @@ insertValidateGenesisDist syncEnv networkName cfg shelleyInitiation = do
...
-              lift $ mapM_ (insertTxOuts syncEnv tracer hasConsumed disInOut bid) $ genesisUtxOs cfg
+              unless disInOut $ do
+                lift $ mapM_ (insertTxOuts syncEnv tracer bid) $ genesisUtxOs cfg
               liftIO . logInfo tracer $
                 "Initial genesis distribution populated. Hash "
                   <> renderByteArray (configGenesisHash cfg)
               when hasStakes $
                 insertStaking tracer useNoCache bid cfg
-              supply <- lift DB.queryTotalSupply
+              supply <- lift $ DB.queryTotalSupply (getTxOutTableType syncEnv)
```
kderme, line 162: can be removed

cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs
```diff
@@ -0,0 +1,541 @@
...
+
+runExtraMigrations :: (MonadBaseControl IO m, MonadIO m) => Trace IO Text -> TxOutTableType -> Word64 -> PruneConsumeMigration -> ReaderT SqlBackend m ()
+runExtraMigrations trce txOutTableType blockNoDiff pcm = do
+  ems <- queryAllExtraMigrations
+  let migrationValues = processMigrationValues ems pcm
+  -- Has the user given txout address config && the migration wasn't previously set
+  when (isTxOutVariantAddress txOutTableType && not (isTxOutAddressPreviouslySet migrationValues)) $ do
+    updateTxOutAndCreateAddress
+    insertExtraMigration TxOutAddressPreviouslySet
+
```
kderme, line 125: Fail if `(not isTxOutVariantAddress) && isTxOutAddressPreviouslySet`

cardano-db/src/Cardano/Db/Types.hs
```diff
@@ -221,6 +241,7 @@ data PruneConsumeMigration = PruneConsumeMigration
   , -- we make the assumption that if the user is using prune flag
     -- they will also want consume automatically set for them.
     pcmConsumeOrPruneTxOut :: Bool
+  , pcmConsumedTxOut :: Bool
```
kderme, line 244: Remove `pcmConsumedTxOut` and rename `pcmConsumeOrPruneTxOut` to `pcmConsumedTxOut`

cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs
```diff
@@ -0,0 +1,541 @@
...
+  ) =>
+  ReaderT SqlBackend m ()
+createConsumedTxOut = do
+  handle exceptHandler $ rawExecute createIndex []
+  handle exceptHandler $ rawExecute addConstraint []
+  where
+    createIndex =
+      "CREATE INDEX IF NOT EXISTS idx_tx_out_consumed_by_tx_id ON tx_out (consumed_by_tx_id)"
+
+    addConstraint =
```
kderme, line 363: We only want this on `prune`, not consumed

cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs
```diff
@@ -0,0 +1,541 @@
...
+  }
+
+--------------------------------------------------------------------------------------------------
+-- Queries
+--------------------------------------------------------------------------------------------------
+queryUpdateListTxOutConsumedByTxId :: MonadIO m => [(TxOutIdW, TxId)] -> ReaderT SqlBackend m ()
+queryUpdateListTxOutConsumedByTxId ls = do
+  mapM_ (uncurry updateTxOutConsumedByTxId) ls
+
+queryTxConsumedColumnExists :: MonadIO m => ReaderT SqlBackend m Bool
```
kderme, line 59: Can be removed

cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs
```diff
@@ -0,0 +1,541 @@
...
+      (MonadIO m, TxOutFields a) =>
+      ReaderT SqlBackend m Word64
+    query = do
+      res <- select $ do
+        txOut <- from $ table @(TxOutTable a)
+        where_ (isJust $ txOut ^. txOutConsumedByTxIdField @a)
+        pure countRows
+      pure $ maybe 0 unValue (listToMaybe res)
+
+_validateMigration :: MonadIO m => Trace IO Text -> TxOutTableType -> ReaderT SqlBackend m Bool
```
kderme, line 509: Can be removed

cardano-db/src/Cardano/Db/Operations/Other/MinId.hs
```diff
@@ -0,0 +1,165 @@
...
+completeMinId ::
+  (MonadIO m) =>
+  Maybe TxId ->
+  MinIdsWrapper ->
+  ReaderT SqlBackend m MinIdsWrapper
+completeMinId mTxId mIdW = case mIdW of
+  CMinIdsWrapper minIds -> CMinIdsWrapper <$> completeMinIdCore mTxId minIds
+  VMinIdsWrapper minIds -> VMinIdsWrapper <$> completeMinIdVariant mTxId minIds
+
+completeMinIdCore :: MonadIO m => Maybe TxId -> MinIds 'TxOutCore -> ReaderT SqlBackend m (MinIds 'TxOutCore)
```
kderme, line 121: Remove duplication

cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs
```diff
@@ -0,0 +1,541 @@
...
+
+updateTxOutAndCreateAddress ::
+  forall m.
+  ( MonadBaseControl IO m
+  , MonadIO m
+  ) =>
+  ReaderT SqlBackend m ()
+updateTxOutAndCreateAddress = do
+  handle exceptHandler $ rawExecute alterTxOutQuery []
+  handle exceptHandler $ rawExecute createAddressTableQuery []
```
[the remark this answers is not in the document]
Cmdv: yeah looks like I'll have to also remove existing indexes/views too:
```
DBExtraMigrationSqlError 
  { sqlState = "2BP01", 
    sqlExecStatus = FatalError, 
    sqlErrorMsg = "cannot drop column address of table tx_out because other objects depend on it", 
    sqlErrorDetail = "view utxo_byron_view depends on column address of table tx_out\nview utxo_view depends on column address of table tx_out", 
    sqlErrorHint = "Use DROP ... CASCADE to drop the dependent objects too."
  }
 ```

**Review by kderme — CHANGES_REQUESTED** [review]
Some indexes are still required in address table

File: cardano-db/src/Cardano/Db/Operations/Other/ConsumedTxOut.hs [hunk 19]
```diff
@@ -0,0 +1,505 @@
...
+    exceptHandler e =
+      liftIO $ throwIO (DBPruneConsumed $ show e)
+
+createConsumedConstraintTxOut ::
+  forall m.
+  ( MonadBaseControl IO m
+  , MonadIO m
+  ) =>
+  ReaderT SqlBackend m ()
+createConsumedConstraintTxOut = do
```
kderme, line 359:
```suggestion
createPruneConstraintTxOut = do
```
since it's only necessary for prune. Also this is not currently used, see https://github.com/IntersectMBO/cardano-db-sync/pull/1820/files#r1757131157

Cmdv: I was working through previous reviews and resolving conversations to the one's that I'd done. Not finished this yet :)

ArturWieczorek:
I had set "prune" level and then change it to "bootstrap:

```json
    "tx_out": {
      "value": "prune",
      "use_address_table": true,
    },
```

At the end of sync process on `sanchonet` I ran into issue: 
```sh
[db-sync-node:Info:81] [2024-09-23 11:15:55.57 UTC] Starting epoch 466
[db-sync-node:Info:81] [2024-09-23 11:15:55.57 UTC] Insert Conway Block: epoch 466, slot 40262417, block 1990094, hash d1e4b009dbff7b18883e09744c1f6eeaf2e7b67c1b9f12f0d45cb75b4b6deef5
[db-sync-node:Info:90] [2024-09-23 11:15:55.73 UTC] Asynchronously wrote a ledger snapshot to ledger-state/sancho/40262388-9a208b3651-465.lstate in 0.164323499s.
[db-sync-node:Info:81] [2024-09-23 11:15:56.04 UTC] Inserted 97 EpochStake for EpochNo 467
[db-sync-node:Info:81] [2024-09-23 11:15:56.75 UTC] Deleted 0 tx_out
[db-sync-node:Info:81] [2024-09-23 11:15:56.92 UTC] Voting Anchor Offchain metadata fetch: 0 results, 3 fetch errors
[db-sync-node:Info:81] [2024-09-23 11:15:56.92 UTC] Starting UTxO bootstrap migration
[db-sync-node:Warning:81] [2024-09-23 11:15:56.97 UTC] Found and deleted 61224 tx_out.
[db-sync-node:Info:81] [2024-09-23 11:15:56.97 UTC] Inserting 87766 tx_out as pages of 100000
[db-sync-node:Info:81] [2024-09-23 11:15:56.97 UTC] Bootstrap in progress 0.0%
[db-sync-node:Info:81] [2024-09-23 11:16:13.68 UTC] UTxO bootstrap migration done
[db-sync-node:Info:81] [2024-09-23 11:16:13.69 UTC] The table epoch_stake was given a new unique constraint called unique_epoch_stake
[db-sync-node:Info:81] [2024-09-23 11:16:13.69 UTC] The table reward was given a new unique constraint called unique_reward
[db-sync-node:Info:81] [2024-09-23 11:16:13.71 UTC] Running database migrations in mode Indexes
[db-sync-node:Info:81] [2024-09-23 11:16:13.71 UTC] Found maintenance_work_mem=1GB, max_parallel_maintenance_workers=6
[db-sync-node:Warning:81] [2024-09-23 11:16:13.71 UTC] Creating Indexes. This may require an extended period of time to perform. Setting a higher maintenance_work_mem from Postgres usually speeds up this process. These indexes are not used by db-sync but are meant for clients. If you want to skip some of these indexes, you can stop db-sync, delete or modify any migration-4-* files in the schema directory and restart it.
ExitFailure 3

Errors in file: /tmp/migrate-2024-09-23T11:16:13.713492769Z.log

[db-sync-node:Error:81] [2024-09-23 11:16:14.78 UTC] runDBThread: ExitFailure 1
[db-sync-node:Error:105] [2024-09-23 11:16:14.78 UTC] ChainSyncWithBlocksPtcl: AsyncCancelled
[db-sync-node.Subscription:Error:101] [2024-09-23 11:16:14.78 UTC] Identity Application Exception: LocalAddress "../sancho/node.socket" SubscriberError {seType = SubscriberWorkerCancelled, seMessage = "SubscriptionWorker exiting", seStack = []}
```

```sh
cat /tmp/migrate-2024-09-23T11:16:13.713492769Z.log
Running : migration-1-0000-20190730.sql
WARNING: password file "config/pgpass-sancho" has group or world access; permissions should be u=rw (0600) or less
 init 
------
 
(1 row)

Running : migration-4-0002-20200810.sql
WARNING: password file "config/pgpass-sancho" has group or world access; permissions should be u=rw (0600) or less
Running : migration-4-0003-20210116.sql
WARNING: password file "config/pgpass-sancho" has group or world access; permissions should be u=rw (0600) or less
psql:schema/migration-4-0003-20210116.sql:12: NOTICE:  relation "idx_block_previous_id" already exists, skipping
psql:schema/migration-4-0003-20210116.sql:20: NOTICE:  relation "idx_pool_update_registered_tx_id" already exists, skipping
psql:schema/migration-4-0003-20210116.sql:37: ERROR:  column "stake_address_id" does not exist
ExitFailure 3
```

kderme:
> At the end of sync process on sanchonet I ran into issue:
 
I think this index also needs to move to file stage 4, version 1, since it's about tx_out.

Cmdv:
Yeah I fixed that yesterday we were talking via dms 👍

kderme (changes requested):

cardano-db/src/Cardano/Db/Migration.hs
```diff
@@ -238,7 +239,10 @@ createMigration source (MigrationDir migdir) = do
     create :: ReaderT SqlBackend (NoLoggingT IO) (Maybe (MigrationVersion, Text))
     create = do
       ver <- getSchemaVersion
-      statements <- getMigration migrateCardanoDb
+      -- here is the place to combine any "core" schemas to the base schema
+      statementsBase <- getMigration migrateBaseCardanoDb
+      statementsTxOut <- getMigration migrateCoreTxOutCardanoDb
```
kderme, line 244: I thought that since tx_out core migration has access to the base migration, only the former needs to be used, but apparently I was wrong?

Also, could we have a flag here, to allow creating the schema for the variant?

Cmdv: I tried with just `statementsBase` then just `statementsTxOut` but they both failed so I put them together. 

I added the option when running the migration but I'll add one for creating migration too 👍

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -83,47 +87,61 @@ basicPrune = do
...
     cmdLineArgs = initCommandLineArgs
     testLabel = "configPrune"
 
 pruneWithSimpleRollback :: IOManager -> [(Text, Text)] -> Assertion
-pruneWithSimpleRollback = do
-  withCustomConfig cmdLineArgs (Just configPruneForceTxIn) babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
+pruneWithSimpleRollback = peformPruneWithSimpleRollback False
+
+pruneWithSimpleRollbackWithAddress :: IOManager -> [(Text, Text)] -> Assertion
+pruneWithSimpleRollbackWithAddress = peformPruneWithSimpleRollback False
```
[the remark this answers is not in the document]
Cmdv: good catch, I'd not paid as much attention to the Babbage tests!

cardano-db/src/Cardano/Db/Schema/Core/TxOut.hs
```diff
@@ -0,0 +1,86 @@
...
+
+schemaDocs :: [EntityDef]
+schemaDocs =
+  document entityDefs $ do
+    TxOut --^ do
+      "A table for transaction outputs."
+      TxOutAddress # "The human readable encoding of the output address. Will be Base58 for Byron era addresses and Bech32 for Shelley era."
+      TxOutAddressHasScript # "Flag which shows if this address is locked by a script."
+      TxOutDataHash # "The hash of the transaction output datum. (NULL for Txs without scripts)."
+      TxOutIndex # "The index of this transaction output with the transaction."
```
kderme, line 73: The docs are missing an entry for `consumedByTxId`, which now exist by default. They should mention that it's not populated by default, unless the necessary flags are on.

kderme (changes requested):
Some comments above.

Also I run gen-schema-docs > doc/schema.md and the result docs were missing the tx_out and ma_tx_out description. Is there any way to add those for the core case?

cardano-db-sync/src/Cardano/DbSync/Era/Byron/Genesis.hs
```diff
@@ -171,52 +175,97 @@ validateGenesisDistribution prunes disInOut tracer networkName cfg bid =
...
+        , V.addressPaymentCred = Nothing -- Byron does not have a payment credential.
+        , V.addressStakeAddressId = Nothing -- Byron does not have a stake address.
+        }
+
+    insertAddress ::
+      (MonadBaseControl IO m, MonadIO m) =>
+      ByteString ->
+      V.Address ->
+      ReaderT SqlBackend m V.AddressId
+    insertAddress addrRaw vAdrs = do
```
kderme, line 263: I think doing a query and then an insert is not the best approach, as it does 2 db operations normally. We could use instead `insertCheckUnique`, which does a single query. This will require a Unique Key in the address.raw field, both in the persistent schema and the raw sql that actually create the schema.

Given that unique key, the index on address.raw then becomes reduandant

kderme: Actually I take that back. The current index causes different problems. Syncing mainnet I got this crash
```
[db-sync-node:Error:85] [2024-09-26 15:16:45.43 UTC] runDBThread: DbInsertException "insertAddress" 
(SqlError {sqlState = "54000", sqlExecStatus = FatalError, sqlErrorMsg = "index row size 7704 exceeds
 btree version 4 maximum 2704 for index \"idx_address_raw\"", sqlErrorDetail = "Index row references
 tuple (32866,36) in relation \"address\".", sqlErrorHint = "Values larger than 1/3 of a buffer page cannot
 be indexed.\nConsider a function index of an MD5 hash of the value, or use full text indexing."})
```
So a unique key will suffer from the same issue. Since [postgres doesn't support unique keys based on hash](https://stackoverflow.com/questions/44274080/postgres-hash-index-with-unique-constraint), the best approach is to replace the current btree index with a hash index

kderme (approved):
Great work!

PR #1827, opened 2024-08-26
Fix consumed with migration

kderme:
# Description

Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1821 with a migration. The fix takes a few hours.
It downloads all Byron blocks and updates the wrong entries.
```
[db-sync-node:Info:95] [2024-08-26 13:36:39.55 UTC] Found 5174749 consumed_by_tx_id wrong entries
[db-sync-node:Info:95] [2024-08-26 13:36:39.55 UTC] Starting chainsync to fix consumed_by_tx_id Byron entries. See issue https://github.com/IntersectMBO/cardano-db-sync/issues/1821. This makes resyncing unnecessary.
[db-sync-node:Info:95] [2024-08-26 13:38:46.85 UTC] Fixed 200023/5174749 entries
[db-sync-node:Info:95] [2024-08-26 13:40:26.10 UTC] Fixed 400429/5174749 entries
[db-sync-node:Info:95] [2024-08-26 13:42:03.60 UTC] Fixed 601414/5174749 entries
```
# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Sync.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Sync.hs
@@ -34,6 +34,7 @@ import Cardano.DbSync.Api.Types (ConsistentLevel (..), FixesRan (..), LedgerEnv 
 import Cardano.DbSync.Config
 import Cardano.DbSync.Database
 import Cardano.DbSync.DbAction
+import Cardano.DbSync.Fix.ConsumedBy
 import Cardano.DbSync.Fix.PlutusDataBytes
 import Cardano.DbSync.Fix.PlutusScripts
 import Cardano.DbSync.LocalStateQuery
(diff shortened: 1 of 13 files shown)


cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -209,6 +210,28 @@ dbSyncProtocols syncEnv metricsSetters tc codecConfig version bversion =
     backend = envBackend syncEnv
 
     initAction channel = do
+      consumedFixed <- getIsConsumedFixed syncEnv
+      case consumedFixed of
+        Nothing -> oldActionFixes channel
+        Just wrongEntriesSize | wrongEntriesSize == 0 -> do
+          logInfo tracer "Found no wrong entries"
```
Cmdv, line 217: might have been nice to have a little more context as to what this is to do with as it might be confusing seeing this message alone in the logs.

PR #1840, opened 2024-09-02
Cherry-pick 13.5.0.2 master

kderme:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
@@ -332,10 +332,12 @@ storeSnapshotAndCleanupMaybe env oldState appResult blkNo isCons syncState =
   case maybeFromStrict (apNewEpoch appResult) of
     Just newEpoch
       | newEpochNo <- unEpochNo (Generic.neEpoch newEpoch)
-      , newEpochNo > 0 -> do
-          -- TODO: Instead of newEpochNo - 1, is there any way to get the epochNo from 'lssOldState'?
-          liftIO $ saveCleanupState env oldState (Just $ EpochNo $ newEpochNo - 1)
-          pure True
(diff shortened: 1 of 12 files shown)


cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
```diff
@@ -332,10 +332,12 @@ storeSnapshotAndCleanupMaybe env oldState appResult blkNo isCons syncState =
   case maybeFromStrict (apNewEpoch appResult) of
     Just newEpoch
       | newEpochNo <- unEpochNo (Generic.neEpoch newEpoch)
-      , newEpochNo > 0 -> do
-          -- TODO: Instead of newEpochNo - 1, is there any way to get the epochNo from 'lssOldState'?
-          liftIO $ saveCleanupState env oldState (Just $ EpochNo $ newEpochNo - 1)
-          pure True
+      , newEpochNo > 0
+      , isCons || (newEpochNo `mod` 10 == 0) || newEpochNo >= 503 ->
```
Cmdv, line 336: is it possible to have a little comment as to why these specific numbers?

rdlrt: This should ideally be calculated as (current - x) epochs, for instance - we have ~25K epochs on guild network (not that cache would matter as much on that one, it's pretty small in tx volume - but it did have ~25K stake accounts to load test systems).

sgillespie (approved):

cardano-db-sync/src/Cardano/DbSync/Ledger/State.hs
kderme: `current` epoch is far from trivial to calculate in general. It could be though specific number of days from the tip.

PR #1850, opened 2024-09-09
Fix: Add OpenSSL C libs

### LINKED ISSUE #1849: Static builds are broken
**OS**
Your OS: All

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`):
PostgreSQL version: All

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: Nix, docker, or binaries

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): Nix, docker, or binaries

**Problem Report**
After removal of HsOpenSSL, all static build jobs are failing. [Example](https://ci.iog.io/build/5266409)

sgillespie:
# Description

Fixes #1849 

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -252,9 +252,25 @@
               (pkgs.lib.mkIf pkgs.hostPlatform.isMusl
                 (let
                   ghcOptions = [
+                    # Postgresql static is pretty broken in nixpkgs. We can't rely on the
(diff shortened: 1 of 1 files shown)


erikd (approved):
I don't know much about Nix, but that **looks** ok, but most of the Hydra test still fail.

PR #1856, opened 2024-09-12
Make it build with ghc-9.10

erikd:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        compiler-nix-name: [ghc810, ghc96, ghc98]
+        compiler-nix-name: [ghc810, ghc96, ghc98, ghc910]
         include:
           # We want a single job, because macOS runners are scarce.
           - os: macos-latest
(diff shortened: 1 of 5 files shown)


erikd:
Tried to add `ghc-9.10` to the build matrix, but that version does not seem to exist in the support GHC versions with the way CI is set up.

angerman:
Questions I have:
- Do we really need the shell? Can we just use the DevX shell? I guess @sgillespie might have an idea about this.
- Must we pin the cabal version? Or can we just go with `cabal = {}`?

angerman:
As @hamishmack pointed out in #1873, @sgillespie bumped cabal. I think we should combine this and #1873? @sgillespie, @erikd what do you think?

angerman:
Let me reiterate my point about the version:
> I'd rather just go with latest, that seems more robust.
> and _if_ we _must_ pin a version have a clear comment that states _why_ we pin the version.

sgillespie:
Since adding GHC-9.10 to the GH action is ongoing, I'd like to go ahead and get this merged into master. We can add back the entry in the build matrix after we fix devx

sgillespie (approved):

PR #1858, opened 2024-09-16
Update test configs for consumed txout

Cmdv:
Modified-by: Cmdv <vincent@cmdv.me>

# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -54,338 +53,333 @@ import Test.Cardano.Db.Mock.UnifiedApi (
...
-    void $
-      withBabbageFindLeaderAndSubmitTx interpreter mockServer $
-        Babbage.mkPaymentTx (UTxOIndex 0) (UTxOIndex 1) 10000 500
-
-    startDBSync dbSyncEnv
-    assertBlockNoBackoff dbSyncEnv 1
-    assertEqQuery dbSyncEnv DB.queryTxConsumedColumnExists True "missing consumed_by_tx_id column when flag --consumed-tx-out active"
+txConsumedColumnCheck ioManager metadata = do
+  syncNodeConfig <- mkCustomSyncNodeConfig babbageConfigDir args migrateTxOutPruneForceTxIn
+  withCustomConfig args (Just syncNodeConfig) babbageConfigDir testLabel action ioManager metadata
```
kderme, line 71: Would it be possible to change the `withCustomConfig` definition, so that it takes the `migrateTxOutPruneForceTxIn` and similar as an argumant? That way the `args` and `babbageConfigDir` don't have to be used 2 times here and the format of tests don't need to change.

Cmdv: I wanted to do that but `withCustomConfig` is used in a lot of places. But happy to do it as we might as most of these files are the same and it's just insert-options that differ

sgillespie: My main issue with these is that the primary testing logic is buried in a `where` clause. What if wee moved `action` to the end, so it would just be a standard continuation?

Cmdv: it looks a lot better when I put the logic inside of `withCustomConfig` then the testing logic isn't in a where clause :)

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -316,15 +293,12 @@ noPruneSameBlock =
...
-      initCommandLineArgs
-        { claConfigFilename = "test-db-sync-config-consumed.json"
-        }
+    cmdLineArgs = initCommandLineArgs
     testLabel = "configNoPruneSameBlock"
 
 migrateAndPruneRestart :: IOManager -> [(Text, Text)] -> Assertion
 migrateAndPruneRestart = do
-  withCustomConfig cmdLineArgs Nothing babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
+  withCustomConfig cmdLineArgs (Just configTxOutPruneForceTxIn) babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
```
kderme, line 301: I think you need a different config fucntion there. It was not prune before. Same for some other cases as well.

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -360,15 +331,12 @@ pruneRestartMissingFlag = do
...
-      initCommandLineArgs
-        { claConfigFilename = "test-db-sync-config-prune.json"
-        }
+    cmdLineArgs = initCommandLineArgs
     testLabel = "configPruneRestartMissingFlag"
 
 bootstrapRestartMissingFlag :: IOManager -> [(Text, Text)] -> Assertion
 bootstrapRestartMissingFlag = do
-  withCustomConfig cmdLineArgs Nothing babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
+  withCustomConfig cmdLineArgs (Just configTxOutPruneForceTxIn) babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
```
kderme, line 339: Same here, you want bootstrap, not prune

kderme (changes requested):

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -54,9 +53,31 @@ import Test.Cardano.Db.Mock.UnifiedApi (
...
+configBootstrap cfg = do
+  cfg {dncInsertOptions = (dncInsertOptions cfg) {sioTxOut = TxOutBootstrap (ForceTxIn False)}}
+
+------------------------------------------------------------------------------
+-- Tests
+------------------------------------------------------------------------------
 txConsumedColumnCheck :: IOManager -> [(Text, Text)] -> Assertion
 txConsumedColumnCheck = do
-  withCustomConfigAndDropDB cmdLineArgs Nothing babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
+  withCustomConfigAndDropDB cmdLineArgs (Just configTxOutPruneForceTxIn) babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
```
kderme, line 80: The file used was `test-db-sync-config-consumed.json` which assumes only consumed is used and not pruned. Also these config functions could be shared between Conway and Babbage. How about moving them to `Test.Cardano.Db.Mock.Config`?

kderme (changes requested):

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
Cmdv: lol couldn't work out what was happening as it's all changed locally. The files weren't saved 🤦 
But yeah I'll move the configs to `Test.Cardano.Db.Mock.Config` 👍

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -291,15 +259,12 @@ pruneSameBlock =
...
-      initCommandLineArgs
-        { claConfigFilename = "test-db-sync-config-prune.json"
-        }
+    cmdLineArgs = initCommandLineArgs
     testLabel = "configPruneSameBlock"
 
 noPruneSameBlock :: IOManager -> [(Text, Text)] -> Assertion
 noPruneSameBlock =
-  withCustomConfig cmdLineArgs Nothing babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
+  withCustomConfig cmdLineArgs (Just configPruneForceTxIn) babbageConfigDir testLabel $ \interpreter mockServer dbSyncEnv -> do
```
kderme, line 267: The file used was `test-db-sync-config-consumed.json`, so this should be consumed instead of prune

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Babbage/Config/MigrateConsumedPruneTxOut.hs
```diff
@@ -385,7 +341,6 @@ bootstrapRestartMissingFlag = do
   where
     cmdLineArgs =
       initCommandLineArgs
-        { claConfigFilename = "test-db-sync-config-bootstrap.json"
-        , claBootstrap = True
+        { claBootstrap = True
```
kderme, line 344: With bootstrap moved to config files, I think this is currectly never used. It doesn't need to be done in this pr, since it's a separate issue.

kderme (changes requested):

kderme (approved):

PR #1877, opened 2024-10-10
1847 - add logging to Rollbacks

Cmdv:
# Description

Adding some login to rollbacks for a clearer picture of what is happening.
Also refactored code and moved functionality to `cardano-db/` library.

SCR-20241010-kbm

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db/src/Cardano/Db/Operations/Delete.hs
```diff
@@ -39,48 +47,69 @@ import Control.Monad.IO.Class (MonadIO, liftIO)
...
   mMinIds <- fmap (textToMinIds txOutTableType =<<) <$> queryReverseIndexBlockId blockId
   (cminIds, completed) <- findMinIdsRec mMinIds mempty
   mTxId <- queryMinRefId TxBlockId blockId
   minIds <- if completed then pure cminIds else completeMinId mTxId cminIds
-  blockCountInt <- deleteTablesAfterBlockId blockId mTxId minIds
-  pure (mTxId, blockCountInt)
+  deleteEpochLogs <- deleteUsingEpochNo epochN
+  (deleteBlockCount, blockDeleteLogs) <- deleteTablesAfterBlockId blockId mTxId minIds
+  whenJust mIsConsumedTxOut $ \_ -> do
+    querySetNullTxOut trce txOutTableType mTxId
```
kderme, line 109: Could we include this in the delete logs summary?
Instead of whenJust here, you could define and use `whenJustDefault`. For inspiration, we have similar utilitis in `Cardano.DbSync.Util.whenDefault,whenMaybe`

cardano-db/src/Cardano/Db/Operations/Delete.hs
```diff
@@ -102,106 +131,230 @@ deleteBlocksBlockId trce txOutTableType blockId = do
       CMinIdsWrapper (MinIds m1 m2 m3) -> isJust m1 && isJust m2 && isJust m3
       VMinIdsWrapper (MinIds m1 m2 m3) -> isJust m1 && isJust m2 && isJust m3
 
--- (MinIds m1 m2 m3) isJust m1 && isJust m2 && isJust m3
+mkRollbackSummary :: [(Text, Int64)] -> Text
+mkRollbackSummary logs = "\n ----------------------- Rollback Summary: ----------------------- \n" <> formattedLog
```
kderme, line 135: Could this become a bit less verbose? For example we could split out logs with 0 deletes and print them in a single row
Count 0: TableA, TableB ...

cardano-db/src/Cardano/Db/Operations/Query.hs
```diff
@@ -218,22 +219,39 @@ queryBlockNoAndEpoch blkNo = do
...
-        Just (Value blkid, Value (Just epoch)) -> Just (blkid, epoch)
-
-queryBlockSlotNo :: MonadIO m => Word64 -> ReaderT SqlBackend m (Maybe BlockId)
+  pure $ convertBlockQuery (listToMaybe res)
+
+queryBlockSlotNo :: MonadIO m => Word64 -> ReaderT SqlBackend m (Maybe (BlockId, Word64))
 queryBlockSlotNo slotNo = do
   res <- select $ do
     blk <- from $ table @Block
     where_ (blk ^. BlockSlotNo ==. just (val slotNo))
```
kderme, line 228: Could we change the behaviour here, so that it deletes all blocks with greater or equal slot_no?
Currently it finds blocks with the provided slot_no and only if there is one it deletes it and everythiong after it. This is a bit counter-intuitive.

cardano-db/src/Cardano/Db/Operations/Delete.hs
Cmdv: updated logs see PR description for updated screenshot

sgillespie (dismissed):
LGTM

sgillespie:
The rollback summary looks great!

sgillespie (approved):

PR #1886, opened 2024-10-29
1843 - limit cache when close to tip

### LINKED ISSUE #1843: Disable or limit caches when db-sync reaches the tip of the chain
Caches are used to speed up syncing. However they use memory. We could disable or reduce the caches when db-sync approaches the tip of the chain ie 30 mins before tip.

Cmdv:
# Description

This fixes #1843 

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/Cache.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/Cache.hs
@@ -81,22 +81,24 @@ rollbackCache (ActiveCache _ cache) blockId = do
     atomically $ modifyTVar (cTxIds cache) FIFO.cleanupCache
     void $ rollbackMapEpochInCache cache blockId
 
--- When syncing and we're close to the tip, we can optimise the caches.
+-- When syncing and we get within 2 minutes of the tip, we can optimise the caches
+-- and set the flag to True on ActiveCache.
 optimiseCaches :: MonadIO m => CacheStatus -> ReaderT SqlBackend m CacheStatus
-optimiseCaches c@(ActiveCache isCacheOptomised cache) = do
-  if isCacheOptomised
-    then liftIO $ do
-      -- empty caches not to be used anymore
-      atomically $ modifyTVar (cTxIds cache) FIFO.cleanupCache
-      atomically $ writeTVar (cStake cache) (StakeCache Map.empty (LRU.empty 0))
-      atomically $ modifyTVar (cDatum cache) (LRU.optimise 0)
-      -- empty then limit the capacity of the cache
-      atomically $ writeTVar (cMultiAssets cache) (LRU.empty 50000)
-      -- leaving the following caches as they are:
-      -- cPools, cPrevBlock, Cstats, cEpoch
-      pure c
-    else pure c
-optimiseCaches c = pure c
+optimiseCaches cache =
+  case cache of
+    NoCache -> pure cache
+    ActiveCache True _ -> pure cache
+    ActiveCache False c -> do
+      liftIO $ do
+        -- empty caches not to be used anymore
+        atomically $ modifyTVar (cTxIds c) FIFO.cleanupCache
+        atomically $ writeTVar (cStake c) (StakeCache Map.empty (LRU.empty 0))
+        atomically $ modifyTVar (cDatum c) (LRU.optimise 0)
+        -- empty then limit the capacity of the cache
+        atomically $ writeTVar (cMultiAssets c) (LRU.empty 50000)
+        -- leaving the following caches as they are:
+        -- cPools, cPrevBlock, Cstats, cEpoch
+        pure $ ActiveCache True c
 
 getCacheStatistics :: CacheStatus -> IO CacheStatistics
 getCacheStatistics cs =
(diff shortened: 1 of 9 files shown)


cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -150,34 +168,36 @@ queryStakeAddrWithCacheRetBs ::
...
-            Right stakeAddrsId -> do
-              let !stakeCache' = case cacheUA of
-                    UpdateCache -> stakeCache {scLruCache = LRU.insert cred stakeAddrsId (scLruCache stakeCache)}
-                    UpdateCacheStrong -> stakeCache {scStableCache = Map.insert cred stakeAddrsId (scStableCache stakeCache)}
-                    _ -> stakeCache
-              liftIO $
-                atomically $
-                  writeTVar (cStake ci) stakeCache'
-              pure $ Right stakeAddrsId
+    NoCache -> mapLeft (,bs) <$> resolveStakeAddress bs
```
kderme, line 171:
```suggestion
    ActiveCache False ci -> do 
      ...
    _ -> mapLeft (,bs) <$> resolveStakeAddress bs
```
Merge the 2 states together

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -365,27 +390,30 @@ queryTxIdWithCache cache txIdLedger = do
...
-        -- Cache miss.
-        Nothing -> do
-          eTxId <- DB.queryTxId txHash
-          liftIO $ missTxIds (cStats cacheInternal)
-          case eTxId of
-            Right txId -> do
-              -- Update cache.
-              liftIO $ atomically $ modifyTVar (cTxIds cacheInternal) $ FIFO.insert txIdLedger txId
-              -- Return ID after updating cache.
+    ActiveCache isCacheOptomised cacheInternal -> do
```
kderme, line 393:
```suggestion
    ActiveCache False cacheInternal -> do
       ...
    _ -> DB.queryTxId txHash
```
Same, merge the 2 states together

kderme (changes requested):
`optimiseCaches` requires fixing.
Also `optomise` is sometimes used instead of `optimise`. Using `reduce` instead of optimise would also make it more specific, since the cache is reduced.
Finally, using patterns as suggested could reduce the cases.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
Cmdv: it is but I'd say showing both `NoCache` & `ActiveCache` cases is more informative and simpler to understand when going back in a couple month's times

kderme: I'm fine this it, if we define `mapLeft (,bs) <$> resolveStakeAddress bs` as a function, to avoid duplicating it. Similarly in other function where this pattern appears

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Block.hs
```diff
@@ -64,16 +65,18 @@ insertBlockUniversal ::
   ApplyResult ->
   ReaderT SqlBackend m (Either SyncNodeError ())
 insertBlockUniversal syncEnv shouldLog withinTwoMins withinHalfHour blk details isMember applyResult = do
+  -- if we're syncing within 2 mins of the tip, we optimise the caches.
+  newCache <- if isSyncedWithintwoMinutes details then optimiseCaches cache else pure cache
```
kderme, line 69: I just realised, the CacheStatus, is not the actual state of the cache, so changing the isOptimised flag here, will not stay in the `syncEnv` for future blocks.

kderme (changes requested):
possibly the new boolean flag need to become a TVar

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -80,6 +81,26 @@ rollbackCache (ActiveCache cache) blockId = do
     atomically $ modifyTVar (cTxIds cache) FIFO.cleanupCache
     void $ rollbackMapEpochInCache cache blockId
 
+-- | When syncing and we get within 2 minutes of the tip, we can optimise the caches
+-- and set the flag to True on ActiveCache.leaving the following caches as they are:
+-- cPools, cPrevBlock, Cstats, cEpoch
+optimiseCaches :: MonadIO m => CacheStatus -> ReaderT SqlBackend m CacheStatus
```
kderme, line 87: I think this function doesn't need to return anything, it should isimply modify the CacheStatus state.

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -150,34 +171,34 @@ queryStakeAddrWithCacheRetBs ::
...
-            Right stakeAddrsId -> do
-              let !stakeCache' = case cacheUA of
-                    UpdateCache -> stakeCache {scLruCache = LRU.insert cred stakeAddrsId (scLruCache stakeCache)}
-                    UpdateCacheStrong -> stakeCache {scStableCache = Map.insert cred stakeAddrsId (scStableCache stakeCache)}
-                    _ -> stakeCache
-              liftIO $
-                atomically $
-                  writeTVar (cStake ci) stakeCache'
-              pure $ Right stakeAddrsId
+      withCacheOptimisationCheck ci (mapLeft (,bs) <$> resolveStakeAddress bs) $ do
```
kderme, line 176: Can this be defined as a function to avoid repeating it in NoCache case above?

kderme (commented):
Just a few stylistic changes and this is good to go

kderme (approved):

Issue #1572, opened 2023-12-06
Test Conway


kderme:
After transfering existing tests to Conway with https://github.com/input-output-hk/cardano-db-sync/issues/1525, we need to add new tests with the new functionality. A WIP list:
- Tests that were not transferable because the underslying mechanism has changed: These are MIR tests mirReward, mirRewardDereg, mirReward and tests that require funds in the treasury  rewardsDeregistration, rewardsReregistration.
- Tests for all COnway feautures: drep registrations, voting, voting delegations, ratified and enacted gov actions (TODO: extend)
- HF tests which use a voted governance action instead of hard coded hard fork based on the epoch number.

sgillespie:
> Tests that were not transferable because the underlying mechanism has changed:

The only tests that weren't transferred to Conway were in the "rewards" suite:

 * rewardsDeregistration
 * rewardsReregistration
 * mirReward
 * mirRewardDereg
 * mirReward

So the Conway equivalents will need to be added (if there are any).

All tests that used Shelley certs were updated to use Conway certs, but AFAICT no other behavior changed (other than governance actions)

The only other change in behavior I saw was that the Conway ledger removed duplicate transactions, so we might want to add some tests that verify this in the database.

PR #1898, opened 2024-11-13
Test: Governance rollback tests

sgillespie:
# Description

Addresses #1572. Add some more Conway tests:

 * rollback new committee member (enacted)
 * rollback new committee proposal (not ratified)
 * chained new committee proposals
 * rollback hard fork

# Checklist

- [x] Self-reviewed the diff

# Migrations


cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/Governance.hs
```diff
@@ -73,72 +74,231 @@ drepDistr =
...
+      dbSync
+      (Query.queryCommitteeMemberCountByTxHash Nothing)
+      4
+      "Unexpected committee member count"
 
-    -- Register a DRep and delegate votes to it
-    void (Api.registerDRepsAndDelegateVotes interpreter server)
+    let
+      -- Propose a new committee member
+      committee1Hash = "e0a714319812c3f773ba04ec5d6b3ffcd5aad85006805b047b082541"
```
Cmdv, line 171: very minor but is it worth abstracting this hash out as it's used in 2 places? Happy to leave as is tbh.

Cmdv (approved):
From my limited understanding of Governance these tests seem correct to me. One very minor comment that can be ignored. 

🚢  :shipit:

Cmdv (approved):

kderme (approved):

PR #1908, opened 2024-11-23
fix: `tx_out` table migration

Vardominator:
# Description

There is an issue with your migration of the `tx_out` table which leads to failure upon migration from version `13.5.0.2` to `13.6.0.2`. Here is the error:

```
Running : migration-2-0044-20240912.sql
psql:/nix/store/s7vh6jy8gibyrrqkkhvcazkcmflvv45k-schema/migration-2-0044-20240912.sql:19: ERROR:  column "consumed_by_tx_id" of relation "tx_out" already exists
CONTEXT:  SQL statement "ALTER TABLE "tx_out" ADD COLUMN "consumed_by_tx_id" INT8 NULL"
PL/pgSQL function migrate() line 9 at EXECUTE
ExitFailure 3
```

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/schema/migration-2-0044-20240912.sql
+++ b/schema/migration-2-0044-20240912.sql
@@ -8,7 +8,7 @@ BEGIN
   IF next_version = 44 THEN
     -- EXECUTE 'ALTER TABLE "reward" ADD COLUMN "id" INT8 NOT NULL' ;
     -- EXECUTE 'ALTER TABLE "reward_rest" ADD COLUMN "id" INT8 NOT NULL' ;
-    EXECUTE 'ALTER TABLE "tx_out" ADD COLUMN "consumed_by_tx_id" INT8 NULL' ;
+    EXECUTE 'ALTER TABLE "tx_out" ADD COLUMN IF NOT EXISTS "consumed_by_tx_id" INT8 NULL' ;
     -- Hand written SQL statements can be added here.
     UPDATE schema_version SET stage_two = next_version ;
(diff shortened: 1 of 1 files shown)


kderme:
I don't see how the error can occur. The schema_version.stage_two should turn 44 the first time this runs, so it won't happen afterwards. Did you have an instance with consumed_tx_out enabled before, which you tried to update?

Vardominator:
@kderme The existing instance has this config:
```
  "insert_options": {
    "tx_out": {
      "value": "consumed"
    },
    "pool_stat": "enable",
    "pool_stats": "enable"
  }
  ```

kderme (approved):
The migration is broken  for the `consumed` tx_out and not for the default db-sync. The pr fixes it.

Vardominator:
@kderme Thanks! Will you cut a release or should I work from a fork?

kderme:
Thank you for the fix. I created a pre-release 13.6.0.3-consumed, since a release may take longer

Vardominator:
@kderme Can you please build and push an image for this. Looks like you don't have images for pre releases or I may be missing something.

kderme:
ghcr upload fails for some reasons. Looking into it.

Vardominator:
@kderme Can you please let me know when there is an image available? This is quite urgent for us. Thanks!

kderme:
Done now https://github.com/intersectmbo/cardano-db-sync/pkgs/container/cardano-db-sync/312147399?tag=13.6.0.3-consumed @Vardominator

Issue #1921, opened 2024-12-12
Remove support for GHC8.10


Cmdv:
We are dropping support for GHC8.10 so we need to remove all references of it from our repo.

rdlrt:
Just a note to ensure ghc9* CI builds work prior to dropping ghc8.10, as [per current latest commit](https://github.com/IntersectMBO/cardano-db-sync/actions/runs/12302545920/job/34335638079) - they dont:
![image](https://github.com/user-attachments/assets/0c9ad583-a041-4998-9030-88f2f1d72e8b)

Cmdv:
There is going to be a slight delay on this as we can only go ahead one `cardano-node` has done this work too. By that time I'd imagine we'll have the CI failures fixed, from my understanding it's nix related and is being worked on.

rdlrt:
> Unfortunately the GitHub actions are in a pretty bad state right now

Afaik - hydra CI is based on nix, the github CI is our only point of reference for working cabal builds

profd2004:
@Cmdv this looks done from the repo side, as there is no GHC 8.10 reference left anywhere in the tree and `.github/workflows/haskell.yml` now builds the ghc96, ghc98 and ghc912 matrix through devx. That also seems to settle rdlrt's caveat about having working non-nix builds before the drop. Can you confirm #1922 covered everything you had in mind here, or say what is still outstanding?

profd2004:
@Cmdv closing this as done. There is no GHC 8.10 reference left anywhere in the tree, and `.github/workflows/haskell.yml` now builds the ghc96, ghc98 and ghc912 matrix through devx, which also covers rdlrt's caveat about having working non-nix builds before the drop. If #1922 missed anything you had in mind, reopen with the specific item.

PR #1922, opened 2024-12-13
1921 - Remove support for GHC8.10

Cmdv:
# Description

This is the ground works for #1921 we will complete it when `cardano-node` has also removed 8.10. Also as mentioned in issue we need to make sure the CI works for all versions before removing 8.10.

# Checklist

- [ ] Self-reviewed the diff

# Migrations


flake.nix
```diff
@@ -67,19 +67,15 @@
...
                 (final: prev: {
-                  # HLint 3.2.x requires GHC >= 8.10 && < 9.0
-                  hlint = final.haskell-nix.tool "ghc8107" "hlint" {
-                    version = "3.2.7";
+                  hlint = final.haskell-nix.tool "ghc96" "hlint" {
+                    version = "3.8";
                   };
 
-                  # Fourmolu 0.10.x requires GHC >= 9.0 && < 9.6
                   fourmolu = final.haskell-nix.tool "ghc928" "fourmolu" {
```
sgillespie, line 74:
```suggestion
                  fourmolu = final.haskell-nix.tool "ghc96" "fourmolu" {
```

flake.nix
```diff
@@ -169,7 +162,7 @@
               cabal = "latest";
               haskell-language-server = {
                 src =
-                  if config.compiler-nix-name == "ghc8107" then
+                  if config.compiler-nix-name == "ghc96" then
```
sgillespie, line 165: This conditional isn't necessary anymore, we can now track HLS latest release

doc/installing.md
```diff
@@ -11,8 +11,8 @@ This guide assumes you have the following tools:
 
 In addition, Cardano DB Sync requires the following software (instructions below):
 
- * [GHC](https://www.haskell.org/ghcup/install/) >= 8.10.7
- * [Cabal](https://www.haskell.org/ghcup/install/) >= 3.10.1.0
+ * [GHC](https://www.haskell.org/ghcup/install/) >= 9.6.5
+ * [Cabal](https://www.haskell.org/ghcup/install/) >= 3.12.1.0
```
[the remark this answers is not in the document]
Cmdv: I just thought why not up everything if we can in one go

flake.nix
Cmdv: what would I make `hls-1.10` if we can track latest `hls-latest`?

doc/installing.md
```diff
@@ -235,7 +235,7 @@ Explicitly set the GHC version that we installed earlier. This avoids defaulting
 system version of GHC that might be different than the one you have installed.
 
 ```bash
-echo "with-compiler: ghc-8.10.7" >> cabal.project.local
+echo "with-compiler: ghc-9.8.5" >> cabal.project.local
```
sgillespie, line 238:
```suggestion
echo "with-compiler: ghc-9.6.5" >> cabal.project.local
```

PR #1930, opened 2025-01-09
1828 - Remove code due to hard fork - Dec 2024

### LINKED ISSUE #1828: Clean code after HF
After the HF we no longer need to support older versions, since all users are forced to upgrade to >= 13.4.0.0.
This means we can clean up a lot of fixing and migration code.

Cmdv:
# Description

This fixes #1828 

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Sync.hs
```diff
@@ -206,78 +197,11 @@ dbSyncProtocols syncEnv metricsSetters tc codecConfig version bversion =
...
-                  chainSyncClientFixConsumed backend syncEnv wrongEntriesSize
-              )
-          logInfo tracer $
-            mconcat ["Fixed ", textShow fixedEntries, " consumed_by_tx_id wrong entries"]
-          pure False
-
-    oldActionFixes channel = do
-      fr <- getIsSyncFixed syncEnv
-      let skipFix = soptSkipFix $ envOptions syncEnv
-      let onlyFix = soptOnlyFix $ envOptions syncEnv
```
kderme, line 237: Probably can remove these 2 command options

sgillespie:
I'm sure you already know this, but CI is now failing with:
```

src/Cardano/DbSync/Sync.hs:203:9: error:
    • Couldn't match type ‘((), Maybe BSL.ByteString)’ with ‘()’
      Expected type: IO () -> IO ((), Maybe BSL.ByteString)
        Actual type: IO ((), Maybe BSL.ByteString)
                     -> IO ((), Maybe BSL.ByteString)
    • The first argument of ($) takes one argument,
      its type is ‘cat0 a0 c0’,
      it is specialized to ‘IO ((), Maybe BSL.ByteString)
                            -> IO ((), Maybe BSL.ByteString)’
      In the expression:
        liftIO . logException tracer "ChainSyncWithBlocksPtcl: "
          $ do logInfo tracer "Starting ChainSync client"
               setConsistentLevel syncEnv Unchecked
               (latestPoints, currentTip) <- waitRestartState tc
               let (inMemory, onDisk) = List.span snd latestPoints
               ....
      In the second argument of ‘($)’, namely
        ‘\ _ctx channel
           -> liftIO . logException tracer "ChainSyncWithBlocksPtcl: "
                $ do logInfo tracer "Starting ChainSync client"
                     setConsistentLevel syncEnv Unchecked
                     ....’
    |
203 |         liftIO . logException tracer "ChainSyncWithBlocksPtcl: " $ do
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

cardano-chain-gen/cardano-chain-gen.cabal
```diff
@@ -130,27 +130,6 @@ test-suite cardano-chain-gen
...
-                        Test.Cardano.Db.Mock.Unit.Babbage.Config.MigrateConsumedPruneTxOut
-                        Test.Cardano.Db.Mock.Unit.Babbage.Config.Parse
-                        Test.Cardano.Db.Mock.Unit.Babbage.InlineAndReference
-                        Test.Cardano.Db.Mock.Unit.Babbage.Other
-                        Test.Cardano.Db.Mock.Unit.Babbage.Plutus
-                        Test.Cardano.Db.Mock.Unit.Babbage.Reward
-                        Test.Cardano.Db.Mock.Unit.Babbage.Rollback
-                        Test.Cardano.Db.Mock.Unit.Babbage.Simple
-                        Test.Cardano.Db.Mock.Unit.Babbage.Stake
-                        Test.Cardano.Db.Mock.Unit.Babbage.Tx
```
sgillespie, line 153: Are we sure we want to remove ALL past era tests?

kderme: We could keep some tests as a reference and some that showcase differences in the behaviour between eras. We could filter them together in a meeting.

kderme (commented):
Except for the issues that Sean raised, this looks good to me

sgillespie:
This all looks pretty good to me. As we discussed, we'll still need to add back a small amount of tests from Alonzo and Babbage:

 * Tx and/or Simple suites
 * Babbage certificates
 * Alonzo rewards re-registration

cardano-chain-gen/test/Main.hs
```diff
@@ -27,7 +29,9 @@ tests iom = do
   pure $
     testGroup
       "cardano-chain-gen"
-      [ Conway.unitTests iom knownMigrationsPlain
+      [ Babbage.unitTests iom knownMigrationsPlain
+      , Alonzo.unitTests iom knownMigrationsPlain
+      , Conway.unitTests iom knownMigrationsPlain
```
kderme, line 34: Conway first please

kderme (approved):
LGTM!

PR #1933, opened 2025-01-21
Fix offchain limit

kderme:
Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/1928

### DIFF
--- a/cardano-db-sync/app/http-get-json-metadata.hs
+++ b/cardano-db-sync/app/http-get-json-metadata.hs
@@ -45,7 +45,7 @@ main = do
         Right bs -> bs
 
     cleanOpt :: [String] -> [String]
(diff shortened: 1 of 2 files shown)


cardano-db-sync/app/http-get-json-metadata.hs
```diff
@@ -45,7 +45,7 @@ main = do
         Right bs -> bs
 
     cleanOpt :: [String] -> [String]
-    cleanOpt ls = List.delete "ga" $ List.delete "vote" $ List.delete "drep" $ List.delete "pool" ls
+    cleanOpt ls = ls List.\\ ["url", "ga", "drep", "other", "vote", "committee_dereg", "const"]
```
Cmdv, line 48: might need a note, letting us know why these were removed?

PR #1935, opened 2025-01-22
Postgresql-setup script updates

rdlrt:
# Description

The postgresql-setup script (as well as docker restore functionality) seems to be only used/tested in a standard configuration and old version of postgresql. With the new options available to dbsync configuration, the setup fails over. This PR hopes to add a few minor updates that are hopefully easy to introduce and would largely assist us sticking to official images and not have to maintain a fork

- [x] docker.nix: Bump postgresql to 17 (as restore snapshot fails if snapshots are made using postgres 17), postgres 14 is ancient anyways and there are various improvements (most prominent one being non-locking index maintenance) docker-compose*.yml: Bump postgresql version to 17, node version to 10.1.4 and dbsync version to 13.6.0.4 postgresql-setup.sh:
- [x] Parse psql output using csv instead of manipulating white chars or table seperators
- [x] Use strict checks for ${PGDATABASE} as there can be another database with a prefix/suffix attached
- [x] When creating/restoring a snapshot, existence of ledger_file should be optional, as there are dbsync config options that don't require lstate files, accordingly - make presence of ledger-state-file optional
- [x] When restoring a snapshot (eg: across postgres versions), there will be cases where an informational error may be expected. The presence of exit-on-error prevents use of these snapshots, especially when run from docker created by nix-ops (where it isnt as straightforward to override scripts)

# Checklist

- [x] Self-reviewed the diff

# Migrations


erikd:
@sgillespie We might need your help on the Nix stuff.

sgillespie:
> @sgillespie We might need your help on the Nix stuff.

I think we just need to update nixpkgs/haskell.nix

sgillespie:
Unfortunately, we will not be able to use postgresql_17 until we update the nixpkgs pin for haskell.nix, currently [in progress](https://github.com/input-output-hk/haskell.nix/pull/2307/commits)

rdlrt:
Will re-base and amend once #1936 is merged

sgillespie:
> Will re-base and amend once #1936 is merged

It's done, you should now be able to rebase from master

rdlrt:
> It's done, you should now be able to rebase from master

Thanks a lot @sgillespie , updated this PR

Cmdv (dismissed):
🎉 :shipit: Thanks for the contribution @rdlrt

scripts/postgresql-setup.sh
```diff
@@ -185,24 +187,25 @@ function restore_snapshot {
...
+	  # statement generated by pg_dump
 	  pg_restore \
 	    --schema=public \
 	    --jobs="${numcores}" \
 	    --format=directory \
 	    --dbname="${PGDATABASE}" \
-            --no-owner \
-	    --exit-on-error \
+	    --no-owner \
+	    "$( [ -z "${SKIP_RESTORE_ERROR}" ] && echo " --exit-on-error" )" \
```
sgillespie, line 208: Won't quoting this make it add `""` to the list of arguments? I suspect this will result in:

```
pg_restore: error: too many command-line arguments (first is "")
```

rdlrt:
Thanks @sgillespie - Done

sgillespie:
Can you also add another shellcheck exception?
```
In ./scripts/postgresql-setup.sh line 208:
            $( [ -z "${SKIP_RESTORE_ERROR:-}" ] && echo --exit-on-error ) \
            ^-- SC2046 (warning): Quote this to prevent word splitting.
```

rdlrt:
> Can you also add another shellcheck exception?
> 
> ```
> In ./scripts/postgresql-setup.sh line 208:
>             $( [ -z "${SKIP_RESTORE_ERROR:-}" ] && echo --exit-on-error ) \
>             ^-- SC2046 (warning): Quote this to prevent word splitting.
> ```

That line [albeit surprisingly] does not seem to pop up a warning anymore on latest shellcheck, but added an exclusion regardless

sgillespie (approved):

sgillespie:
> > Can you also add another shellcheck exception?
> > ```
> > In ./scripts/postgresql-setup.sh line 208:
> >             $( [ -z "${SKIP_RESTORE_ERROR:-}" ] && echo --exit-on-error ) \
> >             ^-- SC2046 (warning): Quote this to prevent word splitting.
> > ```
> 
> That line [albeit surprisingly] does not seem to pop up a warning anymore on latest shellcheck v0.9.0-1, but added an exclusion regardless

Interesting, I'll have to revisit next time we update nixpkgs

Issue #1875, opened 2024-10-08
It


Fell-x27:
**OS**
Your OS: Ubuntu 22

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.5.0.2
PostgreSQL version: 17

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: prebuilt binaries

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): none

**Additional context**
When I'm trying to restore the last snapshot, I see the error:

```
pg_restore: error: could not execute query: ERROR:  type "addr29type" already exists
Command was: CREATE DOMAIN public.addr29type AS bytea;
```

**Problem Report**
It looks like there is missed "IF NOT EXISTS".
As result - public mainnet snapshots-out-of-the-box are useless now...

Fell-x27:
I've removed  --exit-on-error flag and got even more:

```
DROP OWNED
db/
db/4203.dat.gz
db/4258.dat.gz
db/4238.dat.gz
db/4176.dat.gz
db/4246.dat.gz
db/4162.dat.gz
db/4262.dat.gz
db/4152.dat.gz
db/4179.dat.gz
db/4234.dat.gz
db/4232.dat.gz
db/4268.dat.gz
db/4193.dat.gz
db/4178.dat.gz
db/4230.dat.gz
db/4221.dat.gz
db/4266.dat.gz
db/4189.dat.gz
db/4256.dat.gz
db/4280.dat.gz
db/4264.dat.gz
db/4195.dat.gz
db/4215.dat.gz
db/4183.dat.gz
db/4276.dat.gz
db/4142.dat.gz
db/4170.dat.gz
db/4191.dat.gz
db/4254.dat.gz
db/4278.dat.gz
db/4219.dat.gz
db/4211.dat.gz
db/4197.dat.gz
db/4213.dat.gz
db/4181.dat.gz
db/4228.dat.gz
db/4160.dat.gz
db/4252.dat.gz
db/4150.dat.gz
db/4248.dat.gz
db/4168.dat.gz
db/4270.dat.gz
db/4205.dat.gz
db/4134.dat.gz
db/4144.dat.gz
db/4138.dat.gz
db/4244.dat.gz
db/toc.dat
db/4236.dat.gz
db/4158.dat.gz
db/4272.dat.gz
db/4174.dat.gz
db/4172.dat.gz
db/4156.dat.gz
[...]

PR #1936, opened 2025-01-27
Upgrade to PostgreSQL 17

sgillespie:
# Description

Add support for PostgreSQL 17. Note that for static building, we still need pg16 for the time being

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,6 +24,10 @@
       url = "github:IntersectMBO/cardano-haskell-packages?ref=repo";
       flake = false;
     };
+    # Note[PostgreSQL 17]: This is a workaround to get postgresql_17 from nixpkgs. It's
+    # available in nixpkgs unstable, but has not been updated in haskell.nix yet. Remove
+    # this after the next time haskell.nix updates nixpkgs.
+    nixpkgsUpstream.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
   };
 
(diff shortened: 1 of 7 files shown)


flake.nix
```diff
@@ -24,6 +24,10 @@
       url = "github:IntersectMBO/cardano-haskell-packages?ref=repo";
       flake = false;
     };
+    # Note[PostgreSQL 17]: This is a workaround to get postgresql_17 from nixpkgs. It's
```
Cmdv, line 27: we could maybe create an issue and link it here so we can track it being done

sgillespie: Good point, we had the original discussion here: https://github.com/IntersectMBO/cardano-db-sync/issues/1875

Cmdv: sweet if that will jog memories then ✅

sgillespie: > sweet if that will jog memories then ✅

Do you think we should create another issue, just repurpose #1875 above?

Cmdv: totally up to you, new issue always helps with "throughput"

sgillespie: OK, I just renamed it. I think a whole new ticket would make it a blatant duplicate

Cmdv (approved):

kderme (approved):

PR #1938, opened 2025-01-30
Persistent to Hasql

Cmdv:
# Description

Converting Persistent to using Hasql as Persistent can be rather limiting for our user case also having the ability to use raw Sql will open us up to performance improvement and aid with specific scenarios like variant runtime database schemas as well as concurrent queries into the db.


cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -126,38 +129,83 @@ insertOffChainPoolResults trce resultQueue = do
...
-          DB.insertOffChainVoteReference $ offChainVoteReferences accessors ocvdId
-          DB.insertOffChainVoteExternalUpdate $ offChainVoteExternalUpdates accessors ocvdId
-      OffChainVoteResultError fe -> void $ DB.insertOffChainVoteFetchError fe
-
     isFetchError :: OffChainVoteResult -> Bool
     isFetchError = \case
       OffChainVoteResultMetadata {} -> False
       OffChainVoteResultError {} -> True
 
+    processResultsBatched :: MonadIO m => [OffChainVoteResult] -> DB.DbAction m ()
```
kderme, line 151: Good idea to use pipeline here. Personal note to review it again.

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -6,109 +6,135 @@
...
-              , textShow mBlockNo
-              , " while rolling back to ("
-              , renderPoint point
-              , "). Applying blocks until a new block is found. The node is currently at "
-              , textShow serverTip
-              ]
-      pure False
+          mBlockNo <- lift $ DB.queryBlockHashBlockNo (SBS.fromShort . getOneEraHash $ blockPointHash blk)
+          case mBlockNo of
+            Nothing -> throwError $ SNErrRollback "Rollback.prepareRollback: queryBlockHashBlockNo: Block hash not found"
```
kderme, line 97: In general the whole purpose of ExceptT is to have better "flow" and to avoid case patterns for errors. I think we lose this in many places.

kderme (changes requested):
I did a pass of the `cardano-db-sync` diffs

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -44,17 +43,43 @@ module Cardano.DbSync.Api (
...
+
+import Control.Concurrent.Class.MonadSTM.Strict (
+  newTBQueueIO,
+  newTVarIO,
+  readTVar,
+  readTVarIO,
+  writeTVar,
+ )
+import Control.Monad.Trans.Maybe (MaybeT (..))
+import qualified Data.Strict.Maybe as Strict
```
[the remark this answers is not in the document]
Cmdv: Cool we agreed in the meeting to keep it to 2, one for external libraries other for internal modules to this repo. I'll make sure to remove the ones with 3

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -201,26 +198,26 @@ insertBlock syncEnv cblk applyRes firstAfterRollback tookSnapshot = do
...
 
-    commitOrIndexes :: Bool -> Bool -> ExceptT SyncNodeError (ReaderT SqlBackend (LoggingT IO)) ()
+    commitOrIndexes :: MonadIO m => Bool -> Bool -> DB.DbAction m ()
     commitOrIndexes withinTwoMin withinHalfHour = do
       commited <-
         if withinTwoMin || tookSnapshot
-          then do
-            lift DB.transactionCommit
-            pure True
+          then pure True
```
kderme, line 205:
```suggestion
          lift DB.transactionCommit
          then pure True
```
Any reason this was removed?

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Insert/Grouped.hs
```diff
@@ -85,100 +90,148 @@ instance Semigroup BlockGroupedData where
...
+  -- Execute independent operations (TxIn, Metadata, Mint) in parallel
+  txInIds <- executePreparedTxIn preparedTxIn
+
+  -- TxOut-dependent operations (MaTxOut + UTxO consumption)
+  maTxOutIds <- processMaTxOuts syncEnv txOutIds grouped
+
+  -- Execute remaining independent operations in parallel with pools
+  liftIO $ do
+    a1 <- async $ DB.runPoolDbAction (envDbEnv syncEnv) (executePreparedMetadata preparedMetadata)
+    a2 <- async $ DB.runPoolDbAction (envDbEnv syncEnv) (executePreparedMint preparedMint)
```
kderme, line 184: Not that using a separate connection from a pool, means these inserts no longer happen in the same open transaction as the rest of the data of the block. This means that we could have cases where half the block is inserted

cardano-db-sync/src/Cardano/DbSync/Default.hs
```diff
@@ -120,14 +118,25 @@ insertBlock ::
...
   insertNewEpochLedgerEvents syncEnv (sdEpochNo details) (apEvents applyResult)
+
+  -- Check stop condition after successful block insertion
+  let stopAtBlock = sioStopAtBlock $ dncInsertOptions $ envSyncNodeConfig syncEnv
+  case stopAtBlock of
+    Just targetBlock | unBlockNo blkNo >= targetBlock -> do
+      liftIO $
+        logInfo tracer $
+          "Reached stop condition at block " <> textShow targetBlock <> ". Stopping db-sync gracefully."
+      throwError $ SNErrDefault (mkSyncNodeCallStack "insertBlock") "Stop condition reached"
```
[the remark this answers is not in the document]
Cmdv: yeah I was wondering wether to just stick it at the epoch level and do it by epoch, same place as the epoch logs

cardano-db/src/Cardano/Db/Statement/Types.hs
```diff
@@ -0,0 +1,207 @@
...
+  uniqueFields ::
+    Proxy a ->
+    -- | Lists of column names that form unique constraints
+    [Text]
+  default uniqueFields :: Proxy a -> [Text]
+  uniqueFields _ = []
+
+  -- | Manual constraint specification for bulk operations only.
+  -- This doesn't affect singular inserts, only bulk operations with conflict handling.
+  bulkUniqueFields :: Proxy a -> [Text]
```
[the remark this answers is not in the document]
Cmdv: you're right spotted that it's in one type but then never used. Think at the time I was trying to potentially handle uniqueFields differently but in the long run wasn't needed

cardano-db-sync/src/Cardano/DbSync/DbEvent.hs
```diff
@@ -1,75 +1,87 @@
...
+    Right val -> pure val
+
+liftDbError :: ExceptT DB.DbError IO a -> ExceptT SyncNodeError IO a
+liftDbError dbAction = do
+  result <- liftIO $ runExceptT dbAction
+  case result of
+    Left dbErr -> throwError $ SNErrDatabase dbErr
+    Right val -> pure val
+
+acquireDbConnection :: [HsqlSet.Setting] -> IO HsqlC.Connection
```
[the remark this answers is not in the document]
Cmdv: I don't have access to `SNErrDatabase` but I can move it also it doesn't do a transaction. I'll change the name and just throw in the same way as I do in `Cardano.Db.Run`

cardano-db-sync/src/Cardano/DbSync/Era/Universal/Adjust.hs
```diff
@@ -47,53 +40,62 @@ import Database.Esqueleto.Experimental (
...
-      deleteReward trce nw cache epochNo (cred, rwd)
-  crds <- rights <$> forM (Set.toList creds) (queryStakeAddrWithCache trce cache DoNotUpdateCache nw)
-  deleteOrphanedRewards epochNo crds
 
-deleteReward ::
-  (MonadBaseControl IO m, MonadIO m) =>
-  Trace IO Text ->
+  -- Process rewards in batches
+  unless (null rewardsToDelete) $ do
+    forM_ (chunksOf maxBulkSize rewardsToDelete) $ \batch -> do
```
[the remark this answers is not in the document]
Cmdv: I tried 1000, then 5000, then 10,000 after that 20,000 it did seem to improve as bulk size went up. Haven't tested in this week with 40,000 so not sure how that compares

cardano-db/src/Cardano/Db/Operations/Delete.hs
```diff
@@ -1,397 +0,0 @@
...
-
-deleteUsingEpochNo :: MonadIO m => Word64 -> ReaderT SqlBackend m [(Text, Int64)]
-deleteUsingEpochNo epochN = do
-  countLogs <-
-    concat
-      <$> sequence
-        [ onlyDelete "Epoch" [EpochNo ==. epochN]
-        , onlyDelete "DrepDistr" [DrepDistrEpochNo >. epochN]
-        , onlyDelete "RewardRest" [RewardRestSpendableEpoch >. epochN]
-        , onlyDelete "PoolStat" [PoolStatEpochNo >. epochN]
```
kderme, line 130: Note that Rewards are not deleted here

cardano-db/src/Cardano/Db/Statement/Base.hs
```diff
@@ -0,0 +1,1473 @@
...
+    dc <- runSession mkDbCallStack $ HsqlSes.statement epochN (parameterisedCountWhere @SC.DrepDistr "epoch_no" "> $1" epochEncoder)
+
+    liftIO $ updateProgress (Just trce) progressRef 2 "Counting RewardRest records..."
+    rrc <- runSession mkDbCallStack $ HsqlSes.statement epochN (parameterisedCountWhere @SC.RewardRest "spendable_epoch" "> $1" epochEncoder)
+
+    liftIO $ updateProgress (Just trce) progressRef 3 "Counting PoolStat records..."
+    psc <- runSession mkDbCallStack $ HsqlSes.statement epochN (parameterisedCountWhere @SC.PoolStat "epoch_no" "> $1" epochEncoder)
+
+    liftIO $ updateProgress (Just trce) progressRef 4 "Counting Reward records..."
+    rc <- runSession mkDbCallStack $ HsqlSes.statement epochN (parameterisedCountWhere @SC.Reward "spendable_epoch" "> $1" epochEncoder)
```
kderme, line 799: Reward shouldn't be deleted here

kderme (changes requested):

PR #1946, opened 2025-02-21
Implement a graceful container shutdown

TrevorBenson:
# Description

This PR implements a change to the buildImage for the cardano-db-sync container.
1. Uses Dockerfile instruction stopsignal to change the default SIGTERM to SIGINT
2. Uses `exec` so that the primary process is the cardano-db-sync binary process, instead of the bash shell script.
3. Added an entry to the CHANGELOG as checklist suggested.

Should resolve issue [cardano-db-sync container doesn't gracefully shut down](https://github.com/IntersectMBO/cardano-db-sync/issues/1945)

# Checklist

- [X] Commit sequence broadly makes sense
- [X] Commits have useful messages
- [X] Any changes are noted in the [changelog](https://github.com/IntersectMBO/cardano-db-sync/blob/master/cardano-db-sync/CHANGELOG.md)
- [X] Self-reviewed the diff

# Migrations

# Additional Details / Empty Checklist Items
* **New tests added or existing updated if needed**. I didn't find tests specific to building the container image. I left it empty in case I overlooked something and reviewer points me at tests that require fixing.
* **Code is formatted with `fourmolu`**. PR changes nix files. The instructions for using fourmolu appeared to be a specific haskell code, so did not check this (yet).


Cmdv:
@TrevorBenson Thanks for the PR, we're just trying to fix the CI failures which aren't actually related to this PR once those are fixed we'll be able to put this into master.

Cmdv (dismissed):

TrevorBenson:
> @TrevorBenson Thanks for the PR, we're just trying to fix the CI failures which aren't actually related to this PR once those are fixed we'll be able to put this into master.

Thanks. I took a quick look at them the other day and noticed they seemed to exist prior to this PR, so figured I could skip looking for root causes.

Glad to know I wasn't mistaken 👍🏼

sgillespie (approved):

PR #1954, opened 2025-03-14
Prepare 13.6.0.5

kderme:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


sgillespie (dismissed):
I presume you've already tested the migration changes, LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Revision history for cardano-db-sync
 
+## 13.6.0.5
+- Fix offchain data [#1928]
```
rdlrt, line 4: Should this clarify that this is bumping `bytesToRead` and `maxBytes` from `10000` to `300000`?
Looking at the large-ish content floating around, there is no indication on length of the request - meaning this can still occur, but atleast a reference somewhere (ideally in docs instead of changelog) would help us to point to "a source"/decision.

PR #1958, opened 2025-03-18
enh(docker): add some tools to the image

TPXP:
# Description

This adds the following packages to the docker image, useful for debugging
* `grep`: output filtering
* `less`: pager used by the `psql` command (it shows error `sh: line 1: less: command not found` without it if output has more than a few lines - for example when listing tables with `\d+`)

# Checklist

- [x] Self-reviewed the diff

# Migrations
No breaking change


sgillespie (approved):
LGTM. You can also always connect via some other container. You might want to check out: https://github.com/dbcli/pgcli. Thanks for the contribution!

PR #1960, opened 2025-03-21
Parallelize

kderme:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-sync/src/Cardano/DbSync/Cache/Stake.hs
```diff
@@ -0,0 +1,155 @@
...
+  takeTMVar,
+  writeTVar,
+ )
+import qualified Control.Concurrent.STM.TBQueue as TBQ
+import Control.Monad.Trans.Control (MonadBaseControl)
+import Data.Either.Combinators
+import qualified Data.Map.Strict as Map
+import Database.Persist.Postgresql (SqlBackend)
+
+-- | TO be called only by the stake thread
```
Cmdv, line 32: If you fancied it you could have a type level constraint on thread specific functions:

```
-- Define thread-specific phantom types
data StakeThread
data BlockThread
data TxThread

-- Create a type that represents actions constrained to specific threads
newtype ThreadConstrained thread m a = ThreadConstrained { runThreadConstrained :: m a }
  deriving (Functor, Applicative, Monad, MonadIO, MonadBaseControl IO)

-- Modify your function to be constrained to the stake thread
resolveInsertRewardAccount ::
  forall m.
  (MonadBaseControl IO m, MonadIO m) =>
  SyncEnv ->
  CacheAction ->
  RewAccount ->
  ThreadConstrained StakeThread (ReaderT SqlBackend m) DB.StakeAddressId
resolveInsertRewardAccount syncEnv cacheUA ra = ThreadConstrained $ do
  eiStakeId <- queryStakeAddrWithCacheRetBs syncEnv cacheUA False ra -- read only
  case eiStakeId of
    Right stakeId -> pure stakeId
    Left (_, bs) -> insertStakeAddress ra (Just bs)

-- Then in your stake thread code, you would run it like:
runStakeThread :: MonadIO m => ReaderT SqlBackend m ()
runStakeThread = do
  -- Get your parameters from somewhere
  let syncEnv = ...
      cacheUA = ...
      ra = ...
  -- Run the thread-constrained function
  stakeId <- runThreadConstrained (resolveInsertRewardAccount syncEnv cacheUA ra)
  -- Do something with stakeId
  ...

-- If someone tries to call it from a different thread, they'd need to write:
runBlockThread :: MonadIO m => ReaderT SqlBackend m ()
runBlockThread = do
  -- This won't type-check because BlockThread ≠ StakeThread
  stakeId <- runThreadConstrained (resolveInsertRewardAccount syncEnv cacheUA ra)
  ...
  ```

File: cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Genesis.hs [hunk 2]
```diff
@@ -268,15 +267,16 @@ insertTxOuts syncEnv trce blkId (TxIn txInId _, txOut) = do
             }
     DB.TxOutVariantAddress -> do
       addrDetailId <- insertAddressUsingCache cache UpdateCache addrRaw vAddress
-      void . DB.insertTxOut $ DB.VTxOutW (makeVTxOut addrDetailId txId) Nothing
+      void . DB.insertTxOut $ DB.VTxOutW (makeVTxOut addrDetailId) Nothing
   where
+    txId = DB.TxKey $ fromIntegral txIndex + 20000 * DB.unBlockKey blkId -- On mainnet genesis produces ~14000 txs
```
Cmdv, line 272: maybe a little comment as to why we're not using the returned txId from running insertTx. I can't fully remember why :)

cardano-db-sync/src/Cardano/DbSync/Ledger/Async.hs
```diff
@@ -0,0 +1,123 @@
...
+import Data.Set (Set)
+
+--------------------------------------------------------------------------------
+-- EpochStake
+--------------------------------------------------------------------------------
+
+newEpochStakeChannels :: IO EpochStakeChannels
+newEpochStakeChannels =
+  -- This may never be more than 1. But let's keep it a queue for extensibility shake.
+  -- This may allow us to parallelize the events workload even further
```
Cmdv, line 27: "parallelise" 😹

Cmdv (commented):
looking very good so far, just very minor comments

cardano-db-sync/src/Cardano/DbSync.hs
```diff
@@ -237,6 +243,7 @@ extractSyncOptions snp aop snc =
         not isTxOutConsumedBootstrap'
           && ioInOut iopts
           && not (enpEpochDisabled snp || not (enpHasCache snp))
+          && False
```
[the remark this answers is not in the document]
kderme: Yes I have it in the TODO list of the pr

cardano-db-sync/src/Cardano/DbSync/Api.hs
```diff
@@ -390,10 +346,59 @@ mkSyncEnv trce backend connectionString syncOptions protoInfo nw nwMagic systemS
   where
     hasLedger' = hasLedger . sioLedger . dncInsertOptions
     isTxOutConsumedBootstrap' = isTxOutConsumedBootstrap . sioTxOut . dncInsertOptions
+    backend = mainBackend backends
+
+withDBSyncConnections ::
```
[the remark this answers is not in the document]
kderme: Yes it could be. I thought `DbConnections` as something that DBSync maintains and could change based on its needs, but this is debatable

cardano-db-sync/src/Cardano/DbSync/Cache.hs
```diff
@@ -15,44 +14,44 @@ module Cardano.DbSync.Cache (
...
 import qualified Cardano.Db as DB
 import qualified Cardano.Db.Schema.Variant.TxOut as V
+import Cardano.DbSync.Api
+import Cardano.DbSync.Api.Types
 import Cardano.DbSync.Cache.Epoch (rollbackMapEpochInCache)
 import qualified Cardano.DbSync.Cache.FIFO as FIFO
 import qualified Cardano.DbSync.Cache.LRU as LRU
-import Cardano.DbSync.Cache.Types (CacheAction (..), CacheInternal (..), CacheStatistics (..), CacheStatus (..), StakeCache (..), initCacheStatistics, shouldCache)
+import Cardano.DbSync.Cache.Stake as X
+import Cardano.DbSync.Cache.Types
```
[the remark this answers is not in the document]
kderme: Stake is the only part of Cache which is splitted in its own module.
If you're refering to `Cardano.DbSync.Cache.Types`, we don't usually reexpose the Types through its most related module. eg see `Cardano.DbSync.Api.(Types)`

cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Genesis.hs
kderme: Personal note to add
```
We're using app generated unique ids to avoid dependencies between tables, which can damage 
parallelism.
```

PR #1964, opened 2025-03-31
Set redeemer fee to null if it is not null

Saghen:
# Description

This avoids updating every row in the redeemer table when migrating to a no ledger schema, which speeds up startup from 7m 15s to 15s in our case.

# Checklist

- [x] Self-reviewed the diff

# Migrations


Cmdv (approved):
Thanks a lot for the contribution this look like a great improvement :)

kderme (approved):

PR #1987, opened 2025-07-08
1986 - Fix duplicate pool stats

### LINKED ISSUE #1986: Duplicate entries for pools for epoch 568 on mainnet in pool_stat table
**OS**
Your OS:
Ubuntu 24.04

**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.6.0.5, git revision cb61094c82254464fc9de777225e04d154d9c782, while cardano-node is 10.1.4
PostgreSQL version: 17

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: prebuilt binaries

**Run method**

**Additional context**
Seems to be similar to previously opened on sanchonet issue - https://github.com/IntersectMBO/cardano-db-sync/issues/1901. Reported by users who were using koios and started getting duplicate entries for pools returned by koios pool_info endpoint that does a join with pool_stat on pool_hash_id and epoch_no assuming uniqueness.

**Problem Report**
Screenshot of sql query output summaring duplicate entries attached

Image

Cmdv:
# Description

This fixes #1986

Due to complexity of partial entries and rollbacks it was just as performant to add a unique constraints to the table

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway.hs
+++ b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway.hs
@@ -120,6 +120,7 @@ unitTests iom knownMigrations =
         , test "rollback stake address cache" Rollback.stakeAddressRollback
         , test "rollback change order of txs" Rollback.rollbackChangeTxOrder
         , test "rollback full tx" Rollback.rollbackFullTx
+        , test "pool stat rollback no duplicates" Rollback.poolStatRollback
         ]
     , testGroup
         "different configs"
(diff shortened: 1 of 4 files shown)


Cmdv:
ok think I've found it:

`onlyDelete "PoolStat" [PoolStatEpochNo >. epochN]` should be `onlyDelete "PoolStat" [PoolStatEpochNo >=. epochN] `

https://github.com/IntersectMBO/cardano-db-sync/blob/e00405dd4f26212cedcc1b20c92aba38ed8009ee/cardano-db/src/Cardano/Db/Operations/Delete.hs#L122-L138

Will write tests see if that's the case

kderme (changes requested):

PR #1989, opened 2025-07-16
1938 - 1 - Variant Name Change

Cmdv:
# Description

Splitting out #1938 into smaller PRs

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cardano-db-tool/src/Cardano/DbTool/Validate/TxAccounting.hs
```diff
@@ -45,8 +45,8 @@ import qualified System.Random as Random
 
 {- HLINT ignore "Fuse on/on" -}
```
sgillespie, line 46: I know you didn't add this, but since you're already in there, you may as well remove this HLINT hint, since it is ignored globally

Cmdv: I'll do one better and remove them all as it seems it's in 11 places 😗

PR #2058, opened 2026-01-22
Fix and re-instate Babbage `simpleRewards` test

Lucsanszky:
# Description

This is the conclusion of the following investigation: https://github.com/IntersectMBO/cardano-ledger/issues/5477
`allPoolStakeCert` needed some adjustments after the `UMap` purge in `cardano-ledger` and had to make sure that we don't re-register already registered staking credentials. Also, because validations were turned off apparently, `FailedToValidateSlot` was not triggered but rather we ended up in an invalid state (the associated delegations were dropped in the account state), thus resulting in a test failure. With validation, attempting to re-register already registered staking credentials would trigger the `StakeKeyAlreadyRegisteredDELEG` predicate failure as per the `DELEG` rule.

# Checklist

- [x] Code is formatted with [`fourmolu`](https://github.com/fourmolu/fourmolu) on version 0.17.0.0 (which can be run with `scripts/fourmolize.sh`)
- [x] Self-reviewed the diff

# Migrations


kderme:
Thank you for looking into this! I'll take a look

PR #2080, opened 2026-03-10
Update dependencies to node-10.7.1

kderme:
# Description

Tracks branch https://github.com/IntersectMBO/cardano-node/tree/geo2a/10.7-integration

# Checklist

- [ ] Self-reviewed the diff

# Migrations


cabal.project
```diff
@@ -106,19 +106,20 @@ if impl (ghc >= 9.12)
 source-repository-package
   type: git
   location: https://github.com/IntersectMBO/cardano-node
-  tag: 1e6d8228693ab2aa4e1d7305e7bdcc57cdd278e4
-  --sha256: sha256-6b9fX1RNmHdBRpmE4WYFCf13vwZZGg1LHRhHP7GRJHE=
+  tag: 787153e5ba32e1abc9ec44c058a28bd304b805cd
+  --sha256: sha256-TLyI56U9rYF1rA1XHV5jHI1CD9XVYbwCHpkUdqmeGyM=
```
erikd, line 110: `cardano-node@10.7.0` Is in CHaP, but this git tag is not the same as https://chap.intersectmbo.org/package/cardano-node-10.7.0/

Is it worth figuring out if anything is different and if its nothing significant switching to the CHaP version?

cabal.project
```diff
@@ -106,19 +106,20 @@ if impl (ghc >= 9.12)
...
 
 if impl(ghc >= 9.12)
+  -- GHC 9.12 support - master branch
   source-repository-package
     type: git
     location: https://github.com/google/proto-lens
-    tag: 20de5227947b0c37dd6852dcc6f2db1cd5889cee
-    --sha256: sha256-VUYU2swjU7L8Zdu6Zfz6jo2ulW5uPhAamt2GjH5hZRY=
+    tag: 13ce67409ad87e8c29eea58b31aa37f44e82457c
+    --sha256: sha256-8XPGLheOo5ZStL7YD5x29GechWFjubw+in5GsUrKWPc=
```
erikd, line 122: A version of `proto-lens` that compiles with `ghc-9.12` (and `ghc-9.14`) has been released to Hackage. Is it worth updating the `index-states` so we get that Hackage version of `proto-lens`?

cabal.project
sgillespie: After another discussion, I think we're going to remove all of the SRPs here. This PR is a little out of date at this point, I expect @kderme will either update this or push another one next week

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway.hs
```diff
@@ -189,35 +189,37 @@ unitTests iom knownMigrations =
...
-    , testGroup
-        "Pools and smash"
-        [ test "pool registration" Other.poolReg
-        , test "query pool that's not registered" Other.nonexistentPoolQuery
-        , test "pool deregistration" Other.poolDeReg
-        , test "multiple deregistration" Other.poolDeRegMany
-        , test "delist pool" Other.poolDelist
-        ]
-    , testGroup
+    , -- TODO: re-enable after fixing credential format for cardano-node 10.7.1
```
Cmdv, line 192: will these need to be put back?

cardano-db-sync/src/Cardano/DbSync/Api/Ledger.hs
```diff
@@ -140,21 +145,15 @@ storePage syncEnv blkEra percQuantum (n, ls) = do
...
 
-prepareTxOut ::
-  ( Cardano.Ledger.Core.Value era ~ MaryValue
-  , Script era ~ AlonzoScript era
-  , TxOut era ~ BabbageTxOut era
-  , BabbageEraTxOut era
-  , DBPlutusScript era
-  , NativeScript era ~ Timelock era
-  ) =>
+prepareTxOut' ::
```
Cmdv, line 148: have all these got `'` for a reason I can see you took away the contraints but was wondering

Cmdv (dismissed):

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway.hs
kderme: These were broken due to upstream changes in the bulk credentials, so something that only the testing framework depends on and not dbsync. It would be nice to have them back, bot sure how easy it will be.

Cmdv: cool no worries was just incase you missed it 😄

Cmdv (approved):
Just nix related stuff it seems

PR #2084, opened 2026-03-23
Make sure we're using indexes during rollbacks 13.7.0.1

kderme:
Fixes https://github.com/IntersectMBO/cardano-db-sync/issues/2083#issuecomment-4110278201 

The `queryMinRefId` query uses 
```sql
SELECT id FROM <table> WHERE <field> >= $1 ORDER BY id ASC LIMIT 1.
```

The planner sometimes picks a bad plan:

```sql
Index Scan using tx_pkey on tx
  Filter: (block_id >= $1)
```
the filter is not Index Cond, so this ends up in a sequential scan. The index refers to the primary key and is only used for sorting. Instead we want to use
```sql
Sort (top-N heapsort)
  -> Index Scan using idx_tx_block_id
       Index Cond: (block_id >= $1)
```
With this change we use 
```sql
SELECT MIN(id) FROM <table> WHERE <field> >= $1
```
 which forces a good plan:

```sql
Aggregate
  -> Index Scan using idx_tx_block_id
       Index Cond: (block_id >= $1)
```
making sure we're using the Index Cond

### DIFF
--- a/cardano-db/src/Cardano/Db/Statement/MinIds.hs
+++ b/cardano-db/src/Cardano/Db/Statement/MinIds.hs
@@ -37,7 +37,9 @@ import Cardano.Db.Types (DbM)
 -- RAW INT64 QUERIES (for rollback operations)
 ---------------------------------------------------------------------------
 
--- | Find the minimum ID in a table - returns raw Int64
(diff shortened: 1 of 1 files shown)


kderme:
Superseeded by https://github.com/IntersectMBO/cardano-db-sync/pull/2080

PR #2091, opened 2026-04-02
Update offchain metadata fetching

Cmdv:
# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations


Cmdv:
@kderme should we make `off_chain_vote_data` be separate from `governance` config. Like this you have to explicitly turn it on even if you have `governance = enabled` the same way we do with `offchain_pool_data`?

kderme:
> @kderme should we make `off_chain_vote_data` be separate from `governance` config. Like this you have to explicitly turn it on even if you have `governance = enabled` the same way we do with `offchain_pool_data`?

I think in order to have `off_chain_vote_data`, we want both governance and the new flag `offchain_vote_data` enabled, which is what https://github.com/IntersectMBO/cardano-db-sync/pull/2091/changes/e9fa4e8bf4a384ef011e7f97a825a0c95edfc064#diff-3f5de63c10da55fa51303ed6c72a86a7762ec327b3732fc10afd553dd9d365bcR271 implements

Cmdv:
yeah you're right I was thrown off by `when (ioOffChainVoteData iopts && ioGov iopts) $ do ...` it was late my brain wasn't working correctly!

kderme:
Superseeded by https://github.com/IntersectMBO/cardano-db-sync/pull/2080

PR #2116, opened 2026-04-30
Clean up constraints and allow-newers

erikd:
# Description

Clean up constraints and allow-newers

# Checklist

- [x] Self-reviewed the diff

# Migrations

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -78,25 +78,10 @@ constraints:
   , any.proto-lens >= 0.7.1.6
   -- haskell.nix patch does not work for 1.6.8
   , any.crypton-x509-system < 1.6.8
+  , validation < 1.2
 
 allow-newer:
     , cardano-api:plutus-core
-    , resource-registry:nothunks
-
-if impl (ghc >= 9.12)
-  allow-newer:
-    -- https://github.com/kapralVV/Unique/issues/11
-    , Unique:hashable
-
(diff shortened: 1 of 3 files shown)


cabal.project
```diff
@@ -77,26 +77,11 @@ constraints:
   -- code,`Data.Text` is being imported unqualified (bad idea IMO) which
   -- then clashes with the `show` in `Prelude`.
   , text < 2.1.2
-  , hasql ==1.9.3
+  , hasql ^>=1.9.3
```
erikd: Not sure why the original constraint was `hasql ==1.9.3`. Are the bug fixes in `1.9.3.1` and `1.9.3.2` not wanted?

Cmdv: yeah in the newer versions you loose low level implementations and it slows things down for our user case

kderme: > yeah in the newer versions you loose low level implementations and it slows things down for our user case

That's 1.10, which is a major breaking change. For `1.9.3.1` and `1.9.3.2` iirc, they brought some dependencies that broke building.

erikd: Has an issue against `hasql` been raised?

erikd: I have reverted the change to the `hasql` constraint.

Cmdv (approved):

PR #2124, opened 2026-05-05
Prepare release 13.7.0.5

kderme:


rdlrt:
Can we please include #2120 in the release? Understand epoch table is being marked redundant for future, but since you have these fixed on master, would be good to have one less error on release supposedly going for hard fork.

*Unless release will be cut from master - in case of which it's just missing a changelog entry

kderme:
It's just missing a changelog entry which I will fix separately

Cmdv (approved):

PR #2130, opened 2026-05-25
2127-Replace epoch table with maintained finalized + live current views

### LINKED ISSUE #2127: Replace `epoch` table with views (hybrid: live + materialised, exposed via UNION)
### Problem
The `epoch` table is a pure rollup of `block`/`tx` (out_sum, fees, tx_count, blk_count, start_time, end_time per epoch), maintained incrementally by db-sync via an in-memory cache + boundary recompute. This maintenance code has produced multiple production bugs: #2118 (value corruption), restart-induced missing rows (e.g. epoch 631)

### Proposal
Hybrid: drop the `epoch` table; expose the same public name via a `UNION ALL` of two backing objects.

- `epoch_finalized`: `MATERIALIZED VIEW` of all epochs *except* the current one. Refreshed by db-sync at each epoch boundary (`REFRESH MATERIALIZED VIEW CONCURRENTLY epoch_finalized`). Fast (index lookup).
- `epoch_current`: plain `VIEW` of the current epoch (or current + previous, to cover the brief boundary gap). Always live (1–3s aggregation, one epoch's worth of data).
- `epoch`: `VIEW` that `UNION ALL`s the two. Public name preserved; consumers see the same column shape and same rows.

db-sync's entire `epochHandler` + epoch cache layer is deleted. The only remaining responsibility is firing the per-boundary REFRESH.

### Acceptance criteria
- `SELECT * FROM epoch …` returns the same rows and shape as today (modulo `id`).
- Current epoch updates per block (live); historical epochs are index-fast.

Cmdv:
# Description

Drop the epoch table and its in-memory cache layer. The public epoch name is preserved as a `UNION ALL` view of `epoch_finalized` (table, appended per boundary, rolled back via `DELETE WHERE no > $epoch`) and `epoch_current` (live view of the unfinalised epoch).

A one-time backfill from block/tx runs on the first start after upgrade. Replaces the `--disable-epoch` CLI flag with the `disable_epoch` option in `insert_options`. Epoch true by default using `disable_epoch` will disable all epoch table related logic.

This fixes #2127

# Checklist

- [ ] Self-reviewed the diff

# Migrations


kderme (changes requested):
I added a test that shows a bug with rollbacks on epoch boundaries.

schema/migration-2-0048-20260422.sql
```diff
@@ -0,0 +1,61 @@
+CREATE FUNCTION migrate() RETURNS void AS $$
+DECLARE
+  next_version int ;
+BEGIN
+  SELECT stage_two + 1 INTO next_version FROM schema_version ;
+  IF next_version = 48 THEN
```
kderme: change to 49 after rebase

cardano-chain-gen/test/Test/Cardano/Db/Mock/Unit/Conway/CommandLineArg/EpochDisabled.hs
```diff
@@ -3,22 +3,25 @@
 module Test.Cardano.Db.Mock.Unit.Conway.CommandLineArg.EpochDisabled (
```
kderme: Could rename to `Test.Cardano.Db.Mock.Unit.Conway.Epoch`

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
```diff
@@ -60,7 +60,12 @@ rollbackFromBlockNo syncEnv blkNo = do
       -- we always need the constraints.
       addConstraintsIfNotExist syncEnv trce
 
-    rollbackCache cache blockId
+    rollbackCache cache
+    -- The surviving block's epoch is no longer finalised; drop its (now stale) row so epoch_current takes over.
+    when (soptEpochViewEnabled (envOptions syncEnv)) $
```
kderme: put inside `queryBlockNoAndEpoch`

kderme (changes requested):

cardano-db-sync/src/Cardano/DbSync/Rollback.hs
kderme, line 48: Add a test that this correctly rollbacks for blocks that are first or last of an epoch.

kderme:
I saw this result running the second test
```
testing=# select * from epoch;
 id | out_sum | fees | tx_count | blk_count | no |     start_time      |      end_time
----+---------+------+----------+-----------+----+---------------------+---------------------
  1 |       0 |    0 |        0 |        99 |  0 | 2021-11-18 20:22:14 | 2021-11-18 20:30:21
  2 |       0 |    0 |        1 |         1 |  1 | 2021-11-18 20:30:29 | 2021-11-18 20:30:29
(2 rows)

testing=# select * from epoch_current ;
 id | out_sum | fees | tx_count | blk_count | no |     start_time      |      end_time
----+---------+------+----------+-----------+----+---------------------+---------------------
  1 |       0 |    0 |        0 |        99 |  0 | 2021-11-18 20:22:14 | 2021-11-18 20:30:21
  2 |       0 |    0 |        1 |         1 |  1 | 2021-11-18 20:30:29 | 2021-11-18 20:30:29
(2 rows)

testing=# select * from epoch_finalized ;
 id | out_sum | fees | tx_count | blk_count | no | start_time | end_time
----+---------+------+----------+-----------+----+------------+----------
(0 rows)
```
From discussion in meeting:
This is an issue with `generateNewEpochEvents` not generating `LedgerNewEpoch` because `envCurrentEpochNo` has the wrong value. This TVar should be updated on rollback, possibly at `rollbackFromBlockNo`. This I believe also run on startup, so this change means there will be no `LedgerStartAtEpoch`. Whatever this event does (logs etc) could happen in `rollbackFromBlockNo` when the TVar is found empty

kderme (changes requested):

kderme (approved):

PR #2131, opened 2026-05-27
feat(offchain): add --allow-private-offchain-urls flag

ArturWieczorek:
# Description

cardano-db-sync's off-chain pool and vote metadata fetchers reject URLs whose host or resolved IP is in a private, loopback, or link-local range as an SSRF mitigation.

 Local-cluster test setups whose metadata is served from http://localhost:.../poolN.json or similar therefore cannot exercise the success path of the fetcher.

This adds an opt-in CLI flag that replaces both layers of the restriction with a no-op for the duration of the run, intended
for local-cluster testing only. Off by default; the existing restriction continues to apply to every production deployment.

* New SyncNodeParams field enpAllowPrivateOffChainUrls plumbed through SyncOptions (soptAllowPrivateOffChainUrls) into the off-chain fetcher threads.
* parseOffChainUrl now takes a Bool: when True the isLocalhostHost rejection is skipped.
* newRestrictedManager now takes a Bool: when True it installs a permissiveRestriction (addressRestriction returning Nothing for every address) instead of offchainRestriction, so DNS-resolved private IPs are no longer blocked at connect time either.
* New Hedgehog suite Cardano.DbSync.OffChain.HttpTest covering both modes for localhost/127./[::1]/10./192.168. plus a public URL and non-HTTP schemes.

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
+++ b/cardano-chain-gen/test/Test/Cardano/Db/Mock/Config.hs
@@ -325,6 +325,7 @@ mkSyncNodeParams staticDir mutableDir CommandLineArgs {..} = do
       , enpForceIndexes = claForceIndexes
       , enpHasInOut = True
       , enpMaybeRollback = Nothing
+      , enpAllowPrivateOffChainUrls = False
       }
 
 ------------------------------------------------------------------------------
--- a/cardano-db-sync/app/cardano-db-sync.hs
+++ b/cardano-db-sync/app/cardano-db-sync.hs
@@ -104,6 +104,7 @@ pRunDbSyncNode = do
     <*> pForceIndexes
     <*> pHasInOut
     <*> optional pRollbackSlotNo
+    <*> pAllowPrivateOffChainUrls
 
 pConfigFile :: Parser ConfigFile
 pConfigFile =
@@ -170,6 +171,19 @@ pHasCache =
     False
     ( Opt.long "disable-cache"
         <> Opt.help "Disables the db-sync caches. Reduces memory usage but it takes longer to sync."
+    )
+
+pAllowPrivateOffChainUrls :: Parser Bool
+pAllowPrivateOffChainUrls =
+  Opt.flag
+    False
+    True
+    ( Opt.long "allow-private-offchain-urls"
+        <> Opt.help
+          ( "Allow the off-chain pool and vote metadata fetchers to connect to "
+              <> "URLs whose host or resolved IP is in a private, loopback, or "
+              <> "link-local range. Intended for local-cluster testing; off by default."
+          )
     )
 
 pSocketPath :: Parser SocketPath
--- a/cardano-db-sync/app/http-get-json-metadata.hs
+++ b/cardano-db-sync/app/http-get-json-metadata.hs
@@ -105,8 +105,8 @@ runHttpGetPool poolUrl mHash =
   where
     httpGet :: ExceptT OffChainFetchError IO SimplifiedOffChainPoolData
     httpGet = do
-      request <- parseOffChainUrl $ OffChainPoolUrl poolUrl
-      manager <- liftIO newRestrictedManager
+      request <- parseOffChainUrl False $ OffChainPoolUrl poolUrl
+      manager <- liftIO (newRestrictedManager False)
       httpGetOffChainPoolData manager request poolUrl mHash
 
     reportSuccess :: SimplifiedOffChainPoolData -> IO ()
@@ -124,7 +124,7 @@ runHttpGetVote voteUrl mHash vtype =
   reportSuccess =<< runOrThrowIO (runExceptT httpGet)
   where
     httpGet :: ExceptT OffChainFetchError IO SimplifiedOffChainVoteData
-    httpGet = httpGetOffChainVoteData [] voteUrl mHash vtype
+    httpGet = httpGetOffChainVoteData False [] voteUrl mHash vtype
 
     reportSuccess :: SimplifiedOffChainVoteData -> IO ()
     reportSuccess spod = do
--- a/cardano-db-sync/app/test-http-get-json-metadata.hs
+++ b/cardano-db-sync/app/test-http-get-json-metadata.hs
@@ -41,7 +41,7 @@ main = do
       let poolUrl = toUrl testPoolOffChain
           mHash = Just $ toHash testPoolOffChain
       eres <- runExceptT $ do
-        request <- parseOffChainUrl (OffChainPoolUrl poolUrl)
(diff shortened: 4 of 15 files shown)


Cmdv:
looks good so far, but could we have it as a config option rather than a command line flag?

ArturWieczorek:
@Cmdv I intentionally kept this as a CLI flag rather than a config option - my understanding is that the `--allow-private-offchain-urls` flag carries a potential security risk and should never be used against a production instance. 

Making it a config option introduces the risk of it being copied between environments, silently creeping into production configs, and being forgotten about. A CLI flag forces an explicit, conscious decision at invocation time, which is the behaviour I want here. Happy to discuss further if you have a different take on the threat model.

ArturWieczorek:
Full regression tests passed :green_heart: 

## Test run link
https://github.com/IntersectMBO/cardano-node-tests/actions/runs/27124217284/job/80048503636
image

## Artifacts

[testrun-report.zip](https://github.com/user-attachments/files/28707174/testrun-report.zip)

[testrun-files.zip](https://github.com/user-attachments/files/28707175/testrun-files.zip)

ArturWieczorek:
# Local Cluster Test

### Step 1 - start the cluster

```sh
nix develop --accept-flake-config .#testenv --command bash -c '
  cd /home/artur/Projects/cardano-node-tests
  source .source.dev
  make start-cluster'
```

Output:

```
Setting up db dbsync0
generated genesis with: 3 genesis keys, ... 3 stake pools ...
Generating Pool 1/2/3 Secrets + Metadata
Starting db-sync   -> dbsync: started
Starting smash     -> smash: started
Re-registering pools, creating CC members and DReps
Transaction successfully submitted...
Cluster started 🚀
```

Confirmed the flag is live in the running db-sync process:

```sh
ps -u "$USER" -o pid,args | grep '[c]ardano-db-sync --config'
# ... --schema-dir .../schema --allow-private-offchain-urls
# db-sync log: SyncOptions {... soptAllowPrivateOffChainUrls = True}
```

Supervisor status / node tip:

```sh
/var/tmp/cardonnay-of-artur/state-cluster0/supervisorctl_local status all
# dbsync RUNNING / nodes:bft1,pool1,pool2,pool3 RUNNING / smash RUNNING / webserver RUNNING
cardano-cli query tip --testnet-magic 42   # block: 50 (and climbing)
```

### Step 2 - restore the original test

Extracted the pre-`6edf306f` test and re-added it to `cardano_node_tests/tests/test_smash.py`
(temporary, uncommitted):

```sh
git show '6edf306f^:cardano_node_tests/tests/test_smash.py'   # source of the original test
```

```python
@allure.link(helpers.get_vcs_link())
@pytest.mark.smoke
def test_fetch_pool_metadata(
    self, locked_pool: dbsync_types.PoolDataRecord, smash: smash_utils.SmashClient
):
    """Test fetching pool metadata from SMASH."""
    pool_id = locked_pool.hash

    def _query_func():
        pool_metadata = next(
            iter(dbsync_queries.query_off_chain_pool_data(pool_id_bech32=locked_pool.view)), None,
        )
        if pool_metadata is None:
            raise dbsync_utils.DbSyncNoResponseError(f"no off-chain pool data record found for pool {pool_id}")
        return pool_metadata

    metadata_dbsync = dbsync_utils.retry_query(query_func=_query_func, timeout=360)
    expected_metadata = smash_utils.PoolMetadata(
        name=metadata_dbsync.json["name"], description=metadata_dbsync.json["description"],
        ticker=metadata_dbsync.ticker_name, homepage=metadata_dbsync.json["homepage"],
    )
    actual_metadata = smash.get_pool_metadata(pool_id=pool_id, pool_meta_hash=metadata_dbsync.hash.hex())
    assert expected_metadata == actual_metadata
```

### Step 3 - run the test

```sh
pytest -s -v cardano_node_tests/tests/test_smash.py::TestBasicSmash::test_fetch_pool_metadata
```

---

Real test output

```
 WARNING: Using `cardano-cli` from custom path!
 WARNING: Using `cardano-node` from custom path!
============================= test session starts ==============================
platform linux -- Python 3.13.9, pytest-9.0.3, pluggy-1.5.0 -- .../.venv/bin/python
metadata: {... 'cardano-node': '11.0.1', 'TESTNET_VARIANT': 'conway_fast', 'NUM_POOLS': '3',
  'CLUSTERS_COUNT': 1, 'HAS_DBSYNC': 'True', 'db-sync': '13.7.0.5',
  'db-sync rev': 'b86b22a749c5dda93815b1a331bcf2f302f8ccc9',
  'db-sync exe': '.../.bin/cardano-db-sync',
  'HAS_SMASH': 'True', 'smash': '13.7.0.5', 'smash exe': '.../.bin/cardano-smash-server'}
configfile: pyproject.toml
collected 1 item

cardano_node_tests/tests/test_smash.py::TestBasicSmash::test_fetch_pool_metadata@smoke
-------------------------------- live log call ---------------------------------
WARNING  cardano_node_tests.utils.dbsync_utils:dbsync_utils.py:510 Sleeping 3s before repeating query for the 1 time.
WARNING  cardano_node_tests.utils.dbsync_utils:dbsync_utils.py:510 Sleeping 6s before repeating query for the 2 time.
PASSED
------------------------------ live log teardown -------------------------------
INFO     cardano_node_tests.utils.dbsync_conn:dbsync_conn.py:35 Closing connection to db-sync database dbsync0.

============================== 1 passed in 9.68s ===============================
>>> PYTEST EXIT: 0
```

Off-chain data db-sync fetched **from localhost** (the fix at work):

```sql
select pmr.url, ocpd.ticker_name, octet_length(ocpd.bytes) as bytes,
       (select count(*) from off_chain_pool_fetch_error) as fetch_errors
from off_chain_pool_data ocpd join pool_metadata_ref pmr on pmr.id = ocpd.pmr_id
order by pmr.url;
```

```
                url                | ticker_name | bytes | fetch_errors
-----------------------------------+-------------+-------+--------------
 http://localhost:23099/pool1.json | TP1         |   128 |            0
 http://localhost:23099/pool2.json | TP2         |   128 |            0
 http://localhost:23099/pool3.json | TP3         |   128 |            0
(3 rows)
```

---

## Final analysis

**The fix works.** With `--allow-private-offchain-urls`, db-sync's off-chain pool fetcher fetched all three `http://localhost:.../poolN.json` files and populated `off_chain_pool_data` with **zero** `off_chain_pool_fetch_error` rows. SMASH then served identical metadata, so the restored `test_fetch_pool_metadata` - the exact test that broke when the SSRF block landed - passes again. The db-sync logs confirmed the flag was active end-to-end (`soptAllowPrivateOffChainUrls = True`, `enpAllowPrivateOffChainUrls = True`).

Cmdv (approved):

kderme (approved):

Issue #1966, opened 2025-04-03
Off-chain DRep metadata image issue


Scitz0:
**Versions**
The `db-sync` version (eg `cardano-db-sync --version`): 13.6.0.5
PostgreSQL version: 17

**Build/Install Method**
The method you use to build or install `cardano-db-sync`: binary from GitHub

**Run method**
The method you used to run `cardano-db-sync` (eg Nix/Docker/systemd/none): systemd

**Problem Report**

**1.**
The processing of off-chain DRep metadata (CIP-119), when image is in base64 format instead of URL,  data put in the `off_chain_vote_drep_data` table for `image_url` field seems to trim off the data URI prefix part of the contentUrl field from json (`json->'body'->'image'->>'contentUrl'`). i.e. the `data:content/type;base64,` part.

**2.**
For the same table (`off_chain_vote_drep_data`), I have seen duplicate rows for the same `off_chain_vote_data_id`.

For example, this query yields three identical rows for me (mainnet):
select * from off_chain_vote_drep_data where given_name = 'Elite';

ArturWieczorek:
Thank you for reporting! 
Both issues were addressed in subsequent PRs that are now merged into master. 
https://github.com/IntersectMBO/cardano-db-sync/pull/2137
https://github.com/IntersectMBO/cardano-db-sync/pull/2138

They'll be included in the next release of `cardano-db-sync`.

PR #2137, opened 2026-06-08
Fix duplicate off-chain vote metadata child rows (#1966)

ArturWieczorek:
Re-processing the off-chain metadata of an already-stored voting anchor duplicated its child rows (off_chain_vote_drep_data, off_chain_vote_author, off_chain_vote_reference, off_chain_vote_external_update, off_chain_vote_gov_action_data). The parent insert used ON CONFLICT DO UPDATE ... RETURNING id, so a re-fetch returned the existing id, and the children - which have no uniqueness and were plain-inserted - were written again.

Make insertion idempotent: insertBulkOffChainVoteDataReturningNew uses ON CONFLICT DO NOTHING RETURNING (voting_anchor_id, hash, id) so only newly inserted parents come back, and the orchestration inserts child rows only for those, matched by natural key. Adds a WithResultBulkColumns result mode for a custom RETURNING column list. Prevents new duplicates; existing rows on synced DBs are unaffected.

Adds a regression test (cardano-db test-db) asserting a re-fetch creates no duplicate child rows.

# Description

# Checklist

- [ ] Self-reviewed the diff

# Migrations

### DIFF
--- a/cardano-db-sync/src/Cardano/DbSync/OffChain.hs
+++ b/cardano-db-sync/src/Cardano/DbSync/OffChain.hs
@@ -152,11 +152,8 @@ insertOffChainVoteResults trce resultQueue = do
         insertBulkOffChainVoteFetchErrors errors
       -- Process metadata in a pipeline if any
       unless (null metadataWithAccessors) $ do
-        -- Insert the parent rows and get back, for each NEWLY inserted parent, its accessor
-        -- paired with the new row id. Parents that were already stored are not returned, so we
-        -- never re-insert their child rows (this is the fix for the duplicate-rows bug, #1966).
+        -- Only newly-inserted parents are returned, so we never re-insert their children (#1966).
         newMetadata <- insertMetadataWithIds metadataWithAccessors
-        -- Now prepare all the related data for bulk inserts
         let allGovActions = catMaybes [offChainVoteGovAction acc ocvdId | (acc, ocvdId) <- newMetadata]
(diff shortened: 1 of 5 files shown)


ArturWieczorek:
Regression Tests :green_heart: 
https://github.com/IntersectMBO/cardano-node-tests/actions/runs/27565196794

Test Report
[testrun-report.zip](https://github.com/user-attachments/files/28974299/testrun-report.zip)

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
```diff
@@ -182,27 +184,29 @@ insertOffChainVoteResults trce resultQueue = do
                   HsqlP.statement allExternalUpdates DB.insertBulkOffChainVoteExternalUpdatesStmt
               pure ()
 
-    -- Helper function to insert metadata and get back IDs
+    -- Insert the off-chain vote metadata rows and return, for each NEWLY inserted parent, its
```
Cmdv: All looks good what I would say if give it another pass minimising all the code comments as LLM' s are OTT when it comes to writing code comment. They also try to write Haddoc style comments but we don't really use it in this repo.

Cmdv (changes requested):

cardano-db-sync/src/Cardano/DbSync/OffChain.hs
ArturWieczorek: Thanks for review, I updated PR.

Cmdv (dismissed):

Cmdv (approved):

PR #2157, opened 2026-07-06
Optimise indexes and autoanalyze

CodiePP:
# Description

* the unique index "unique_reward" needs proper order of columns with the epoch first so the hash tree has less depth. The size of the index drops from 30 GB to 20 GB on disk. This could also be checked for other compound indexes.
* added a few more indexes
* changed the threshold when "autoanalyze" kicks in on heavy tables. This leads to faster queries as the query planning is improved.

We have been benchmarking this on our ha db-sync setup on [BCA](https://blockchain-applied.com).

# Checklist

- [ ] Self-reviewed the diff

# Migrations


Cmdv:
Thanks for the contribution, sorry I totally missed this. I'll make sure we discuss this in our next meeting. To me it looks perfectly fine. The only hit would be when we run the index migrations as it may increase timings. :)

Issue #2162, opened 2026-07-15
cardano-db-tool `utxo-set` and `validate` crash


ArturWieczorek:
# [BUG] cardano-db-tool `utxo-set` and `validate` crash: `SELECT txout.*` decoded by an id-less TxOut decoder (off-by-one column shift)

**Severity:** `utxo-set` is unusable and `validate` cannot complete on any real database.

**Affected:** cardano-db-tool 13.7.2.1 (and earlier on the Hasql line). Verified against HEAD; the relevant files are unchanged between the shipped binary (`db8cdf66`) and HEAD.

## Summary

`cardano-db-tool utxo-set` crashes immediately, and `cardano-db-tool validate` crashes partway through (at the TxAccounting check), with:

```
DbSessionError ... "ResultError (RowError 0 6 UnexpectedNull) QueryError \"SELECT txout.*, ...\""
```

Root cause: the tool queries `SELECT txout.*` (which includes the `id` primary key at column 0) but decodes the result with `txOutCoreDecoder` / `txOutAddressDecoder`, which deliberately omit `id` and start at `tx_id`. Every decoder field is therefore read one column to the left of where it should be.

## Reproduce

```bash
# any inline-address db-sync DB with real UTxOs
PGPASSFILE=<pgpass> cardano-db-tool utxo-set --slot-no <valid slot_no>
PGPASSFILE=<pgpass> cardano-db-tool validate
```

Reproduced on local cluster, restored preview, and restored preprod (13.7.1.0 and 13.7.2.1 DBs). Example on preprod, slot 125357982:

```
cardano-db-tool: DbSessionError ... ResultError (RowError 0 6 UnexpectedNull)
  QueryError "SELECT txout.*, txout.address, tx2.hash FROM tx_out txout LEFT JOIN tx_in txin ..."
```

`validate` reaches TxAccounting then dies on a different query with the same shift:

```
For 100 transactions out of 5990113 accounting is: cardano-db-tool: DbSessionError ...
  ResultError (RowError 0 6 UnexpectedNull)
  QueryError "SELECT txout.* FROM tx tx INNER JOIN tx_in txin ... INNER JOIN tx_out txout ..."
```

## Root cause (pinned to source)

Physical `tx_out` column order (Core), after `address_raw` was dropped in `schema/migration-2-0031-20240117.sql`:

```
id(0) tx_id(1) index(2) address(3) address_has_script(4) payment_cred(5) stake_address_id(6) value(7) ...
[...]

PR #2168, opened 2026-07-15
Isolate validate checks and render errors cleanly

### LINKED ISSUE #2163: cardano-db-tool: `validate` aborts on first failure, and error messages are misleading + leak Haskell internals
# [BUG] cardano-db-tool: `validate` aborts on first failure, and error messages are misleading + leak Haskell internals

**Severity:**  usability / diagnosability. Independent of, but adjacent to, the decoder crash in https://github.com/IntersectMBO/cardano-db-sync/issues/2162.

**Affected:** cardano-db-tool 13.7.2.1 (verified at HEAD).

Three related error-handling defects in `utxo-set` / `validate`:

## 1. `validate` aborts the whole run on the first failing check

Checks are sequenced in plain IO with no `try` / `catch`, so the first exception kills every later check.

- sequencing: `cardano-db-tool/src/Cardano/DbTool/Validation.hs:34-39`
- TxAccounting even calls `error` on an intended mismatch: `cardano-db-tool/src/Cardano/DbTool/Validate/TxAccounting.hs:39` (broken queries `:101-104`)

Effect: because TxAccounting runs early and currently crashes (see Ticket A), the user never sees the Withdrawals / EpochBlockTxs / TotalSupply checks.

**Fix:** wrap each check in `try` and report pass/fail per check instead of aborting.

## 2. Misleading "Database not initialized or not accessible"

A plain slot-not-found (`DbLookupError "Slot not found for slot_no: N"`, `cardano-db/src/Cardano/Db/Statement/Base.hs:147`) is printed by `utxo-set` as "Database not initialized or not accessible" - the DB is fine.

- `cardano-db-tool/src/Cardano/DbTool/UtxoSet.hs:98`

Reproduce:
```bash
PGPASSFILE=<pgpass> cardano-db-tool utxo-set --slot-no 999999999
# Database not initialized or not accessible: DbLookupError {... "Slot not found for slot_no: 999999999"}
```

**Fix:** distinguish "slot not found" from "DB unavailable".

## 3. Errors dump raw Haskell internals (SrcLoc / callstack / file paths)

`UtxoSet.hs:98` renders errors with `show err` instead of the existing `formatDbCallStack`. `DbLookupError` / `DbCallStack` derive `Show` including `SrcLoc`, and `formatSessionError` also `show`s raw Hasql errors.

- `cardano-db/src/Cardano/Db/Error.hs:30,50` (derive Show), `:94` (`formatDbCallStack`, unused), `:120,122` (`formatSessionError` raw show)
- `cardano-db-tool/src/Cardano/DbTool/UtxoSet.hs:98`

Actual output includes `DbCallStack {...}`, `SrcLoc {srcLocPackage=..., srcLocFile="src/Cardano/DbTool/UtxoSet.hs", ...}`.

**Fix:** render user-facing errors with `formatDbCallStack` / a clean formatter, not `show`.

ArturWieczorek:
Closes #2163.

`validate` sequenced its checks in plain IO, so the first check to fail or throw aborted the whole run and everything after it was skipped. In practice the TxAccounting check runs early, so a single failure there hid the withdrawal, epoch-block-txs and total-supply checks entirely. Each check now runs under `try`: a failure is reported, the run carries on to the remaining checks, and the command exits non-zero at the end if any of them failed.

The error rendering was the other half of the problem. `DbLookupError` and `DbSessionError` relied on their derived `Show` for display, so any surfaced error dumped the `DbCallStack` and `SrcLoc` records with package and file paths. On top of that, `utxo-set` reported a plain "slot has no block" as `Database not initialized or not accessible`, which points at the wrong thing entirely. Both error types now have a `displayException` that shows the message (with a readable call chain for the unexpected session errors), and `utxo-set` uses it, so a missing slot reads simply as `Slot not found for slot_no: N`.

The failing test comes first (`Test.Property.Cardano.Db.ErrorRender`, a pure Hedgehog check that `displayException` shows the message without the `DbCallStack` / `SrcLoc` internals), then the fix that turns it green. The isolation behaviour is verified end to end against a restored preprod database: `validate` now walks every check and prints each result before exiting non-zero, and `utxo-set --slot-no <missing>` prints the clean not-found line with no "not initialized" wording and no call-stack dump.

Note this is independent of the decoder crash in #2162. On a database where that bug is still present, `validate` now surfaces the `RowError` per check and keeps going rather than aborting, which is exactly the behaviour this change is meant to provide.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 - The `epoch` table is redesigned and replaced by a view; the in-memory cache that caused [#2118](https://github.com/IntersectMBO/cardano-db-sync/issues/2118) is removed. Reads are unchanged.
 - Fix `pool_relay.port` overflow [#2135](https://github.com/IntersectMBO/cardano-db-sync/issues/2135); a startup migration repairs existing rows.
 - Add `--allow-private-offchain-urls` CLI flag enabling off-chain metadata fetches against private/loopback addresses. Intended for local-cluster testing only; off by default.
-- `cardano-db-tool validate` now runs every check and reports each result instead of aborting on the first failure, and `utxo-set` / `validate` errors show a readable message instead of dumping Haskell call-stack internals [#2163](https://github.com/IntersectMBO/cardano-db-sync/issues/2163).
 
 ## 13.7.1.0
 - Auto-repair `epoch.out_sum` / `epoch.fees` / `epoch.tx_count` / `epoch.blk_count` corruption introduced by issue [#2118](https://github.com/IntersectMBO/cardano-db-sync/issues/2118) in 13.7.0.0 - 13.7.0.4. A startup migration recomputes every epoch row from the underlying `tx` / `block` tables and rewrites only the rows whose stored values disagree. Adds a one-time 5-15 minute delay on the first startup after upgrade on a mainnet-sized database; subsequent restarts and fresh syncs are unaffected. Operators on older release lines who do not upgrade can apply the same fix manually via [`scripts/fix-epoch-table.sql`](scripts/fix-epoch-table.sql).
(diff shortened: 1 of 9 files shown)


ArturWieczorek:
Testing Logs:
[pr2168-dbtool-validate-error-handling-test-results.md](https://github.com/user-attachments/files/30096571/pr2168-dbtool-validate-error-handling-test-results.md)

kderme (approved):

PR #2171, opened 2026-07-15
Count epoch block txs by id and clamp block-range sampling

### LINKED ISSUE #2170: cardano-db-tool validate: two check bugs (block-tx-count uses block_no as id; block-properties range underflow)
# [BUG] cardano-db-tool validate: two check bugs (block-tx-count uses block_no as id; block-properties range underflow)

**Severity:** Low, and the two differ in reach. Bug 1 affects `validate` on every real database, including all three networks (mainnet, preprod, preview). Bug 2 only affects databases with fewer than 100000 blocks - local/private clusters or a DB early in its first sync - and never triggers on preprod, preview, or mainnet. Both are in the `validate` command's own checks and only produce misleading validation output: the shared queries are correct, the data written by sync is unaffected, and this code never runs during sync.

**Affected:** cardano-db-tool 13.7.2.1 (verified at HEAD).

Both were surfaced by the error-isolation change in #2163: once `validate` stops aborting on the first failure, these later checks run on real databases and their pre-existing bugs become visible.

## 1. EpochBlockTxs passes block_no where a block id is expected

`validateEpochBlockTxs` reports false tx-count mismatches on any real database, including mainnet, preprod, and preview.

- `queryEpochBlockNumbers` selects `COALESCE(block_no, 0)` - it returns the chain height `block_no` (`cardano-db/src/Cardano/Db/Statement/DbTool.hs:883`).
- `queryBlockTxCount` counts `tx WHERE block_id = $1` - it expects the surrogate primary key `block.id` (`cardano-db/src/Cardano/Db/Statement/Base.hs:315`).
- The validator bridges them incorrectly: `queryBlockTxCount $ DB.BlockId (fromIntegral blockNo)` (`cardano-db-tool/src/Cardano/DbTool/Validate/BlockTxs.hs:56`), passing a `block_no` as if it were a `block.id`.

`block_no` and `block.id` diverge on any real chain (Byron EBBs, and the id is just a surrogate PK), so the check counts transactions for the wrong block and prints e.g. `Failed on block no 4786984: expected tx count of 2 but got 0`.

**Not a sync bug.** `queryBlockTxCount` itself is correct and is used correctly by sync - both genesis call sites pass a genuine `BlockId` (`cardano-db-sync/src/Cardano/DbSync/Era/Shelley/Genesis.hs:213`, `.../Byron/Genesis.hs:162`). The defect is only in the db-tool validator.

**Fix:** resolve each `block_no` to its `block.id` before calling `queryBlockTxCount` (or have `queryEpochBlockNumbers` return the id alongside/instead of block_no and use that).

Reproduce: `cardano-db-tool validate` against any restored preprod/preview/mainnet DB; the "All transactions for blocks in epoch N are present" check reports mismatches.

## 2. BlockProperties underflows the random block range on small databases

`validateBlockProperties` picks a nonsensical start block on any database with fewer than 100000 blocks. This means local/private clusters and DBs early in their first sync; it does not affect preprod, preview, or mainnet, which are all far past 100000 blocks.

- `startBlock <- Random.randomRIO (0, blkCount - testBlocks)` with `testBlocks = 100000` (`cardano-db-tool/src/Cardano/DbTool/Validate/BlockProperties.hs:56` and `:82`).
- `blkCount` is `Word64`, so when `blkCount < 100000` the subtraction underflows to a huge value, and the check prints an absurd range such as `Block numbers [12258611005600085411 .. 12258611005600185411] are contiguous`.
- The subsequent query returns no rows, so the contiguous/ordered checks pass trivially - a false "ok" rather than a crash.

**Not a sync bug.** `queryBlockNoList` / `queryBlockTimestamps` take `(start, count)` and are only called by this validator; sync does not use them. The underflow is arithmetic in the validator.

**Fix:** guard the range, e.g. clamp to `max 0 (blkCount - testBlocks)` using signed arithmetic, or shrink `testBlocks` to `min testBlocks blkCount`, and skip/adjust when the DB has fewer than `testBlocks` blocks.

Reproduce: `cardano-db-tool validate` against a small database (e.g. a fresh local cluster) - the "Block numbers [...]" and "Block time stamps [...]" ranges are absurd.

## Scope note

Both bugs live under `Cardano.DbTool.Validate.*`, reachable only via `cardano-db-tool validate`. They affect the accuracy of that diagnostic command, not the database contents or the sync process.

ArturWieczorek:
Closes #2170.

Two `cardano-db-tool validate` checks were quietly wrong. Both were only visible once the error-isolation change in #2163 stops `validate` aborting on the first failure, so they were surfaced during review of that work.

The epoch block-tx check counted transactions with `queryBlockTxCount`, which keys on `block.id`, but fed it the `block_no` returned by `queryEpochBlockNumbers`. `block_no` and the surrogate `block.id` diverge on any real chain - an EBB has a null `block_no` and a positive id, and the id is just a sequence - so the check looked up the wrong block and reported false tx-count mismatches on mainnet, preprod, and preview. `queryEpochBlockNumbers` now returns the block id alongside `block_no`, and the check counts by the id while still printing `block_no` in the message. The shared `queryBlockTxCount` was always correct and is used correctly by sync's genesis validation; only this validator misused it.

The block-property checks (contiguous block numbers, ordered timestamps) picked a random start with `blkCount - 100000` in `Word64`, which underflows to an enormous value on any database with fewer than 100000 blocks - local/private clusters and DBs early in their first sync. That produced absurd ranges like `Block numbers [12258611005600085411 .. ...]` (and a false "ok" on the empty result). A small `sampleWindow` helper now clamps the window to the available block count, so the range is always valid. This never affected preprod, preview, or mainnet, which are all far past 100000 blocks.

The first commit adds a failing regression test in the `cardano-db` IO suite that models the id/block_no divergence (a block with a null `block_no` and a positive id, one transaction) and asserts the returned identifier counts the right block; the second commit is the fix. Bug 2 was verified end to end against a small local db-sync database, where the range now reads `Block numbers [0 .. 16608]` instead of the underflowed value.

Marked WIP/draft. Note this overlaps `Validate/BlockTxs.hs` with #2163 (which changes how the same check reports failures), so expect a small merge conflict there - the two changes are complementary: this PR fixes the check's logic, #2163 fixes its reporting and exit code. Bug 1 is only observable through `validate` end to end once #2162 and #2163 are also present (otherwise the run crashes at, or aborts before, the later checks), so the unit test is its standalone verification here.

### DIFF
--- a/cardano-db-tool/src/Cardano/DbTool/Validate/BlockProperties.hs
+++ b/cardano-db-tool/src/Cardano/DbTool/Validate/BlockProperties.hs
@@ -51,16 +51,24 @@ validateBlockTimesInPast = do
     showFirst (mEpoch, mBlockNo, time) =
       mconcat ["epoch ", show mEpoch, " block ", show mBlockNo, " time ", show time]
 
+-- The random start block and how many blocks to sample, clamped so the range
+-- never underflows on a database with fewer than testBlocks blocks.
+sampleWindow :: Word64 -> Word64 -> (Word64, Word64)
+sampleWindow blkCount testBlocks =
+  let count = min testBlocks blkCount
+   in (blkCount - count, count)
+
 validataBlockNosContiguous :: Word64 -> IO ()
 validataBlockNosContiguous blkCount = do
-  startBlock <- Random.randomRIO (0, blkCount - testBlocks)
+  let (maxStart, count) = sampleWindow blkCount testBlocks
+  startBlock <- Random.randomRIO (0, maxStart)
   putStrF $
     "Block numbers ["
       ++ show startBlock
       ++ " .. "
-      ++ show (startBlock + testBlocks)
+      ++ show (startBlock + count)
       ++ "] are contiguous: "
-  blockNos <- DB.runDbStandaloneSilent $ DB.queryBlockNoList startBlock testBlocks
+  blockNos <- DB.runDbStandaloneSilent $ DB.queryBlockNoList startBlock count
   case checkContinguous blockNos of
     Nothing -> putStrLn $ greenText "ok"
     Just xs -> error $ redText "failed: " ++ show xs
@@ -79,14 +87,15 @@ validataBlockNosContiguous blkCount = do
 
 validateTimestampsOrdered :: Word64 -> IO ()
 validateTimestampsOrdered blkCount = do
-  startBlock <- Random.randomRIO (0, blkCount - testBlocks)
+  let (maxStart, count) = sampleWindow blkCount testBlocks
+  startBlock <- Random.randomRIO (0, maxStart)
   putStrF $
     "Block time stamps for blocks ["
       ++ show startBlock
       ++ " .. "
-      ++ show (startBlock + testBlocks)
+      ++ show (startBlock + count)
       ++ "] are ordered: "
-  ts <- DB.runDbStandaloneSilent $ DB.queryBlockTimestamps startBlock testBlocks
+  ts <- DB.runDbStandaloneSilent $ DB.queryBlockTimestamps startBlock count
   if List.nubOrd ts == ts
     then putStrLn $ greenText "ok"
     else error $ redText "failed: " ++ show ts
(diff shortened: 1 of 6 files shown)


ArturWieczorek:
Tests Logs:
[pr2171-dbtool-validate-check-bugs-test-results.md](https://github.com/user-attachments/files/30098287/pr2171-dbtool-validate-check-bugs-test-results.md)

kderme (approved):

